PR #16, opened 2018-02-28
Add checks for dockerfile existence

### LINKED ISSUE #9: Some checks on the passed argument
Checks on the command line arguments. Currently, the only command line argument is a Docker file.  Check on whether the file actually exists or not.

ankuagar:
Check for existence of file that is used to build the docker image.
This change does the check at two places:

- At the commandline

- The function where the docker commands are actually loaded

Resolves #9

### DIFF
--- a/common.py
+++ b/common.py
@@ -4,6 +4,7 @@ SPDX-License-Identifier: BSD-2-Clause
 '''
 
 import logging
+import os
 import subprocess
 import sys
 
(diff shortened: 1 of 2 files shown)


tern
```diff
@@ -50,7 +57,7 @@ if __name__ == '__main__':
     # subparser for report
     parser_report = subparsers.add_parser('report',
                                           help="Create a report")
-    parser_report.add_argument('-d', '--dockerfile',
+    parser_report.add_argument('-d', '--dockerfile', type=check_file_existence,
                                help="Dockerfile used to build the Docker"
```
nishakm: Perfect! Thank you!

common.py
```diff
@@ -52,6 +53,8 @@
 
 def load_docker_commands(dockerfile_path):
     '''Given a dockerfile get a persistent list of docker commands'''
+    if not os.path.isfile(dockerfile_path):
+        raise IOError("{}: Does not exist".format(dockerfile_path))
     global docker_commands
```
nishakm: Looks good. Thanks!

PR #25, opened 2018-03-26
Add openjdk:8 to base.yml

sbtaylor15:
openjdk:8 is based on Debian

Signed-off-by: Steve Taylor <steve.taylor@openmakesoftware.com>

### DIFF
--- a/command_lib/base.yml
+++ b/command_lib/base.yml
@@ -83,7 +83,44 @@ debian:
             1:
               container:
                 - "pkgs=`dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'`"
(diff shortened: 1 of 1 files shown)


sbtaylor15:
Made changes to the EOF for parsing of the licenses coming back from openjdk.  Ran the following tests:

python verify_invoke.py --container 657b6a1db727 --keys base openjdk tags 8 licenses
python verify_invoke.py --container 657b6a1db727 --keys base openjdk tags 8 src_urls

Both came back with matching package to licenses and packages to src_urls.

nishakm:
Thank you!

Issue #29, opened 2018-04-11
Convert short names of modules to the full name


nishakm:
For example:
from utils import dockerfile as df

Convert that to:
from utils import dockerfile

It would require substituting df in the module file with dockerfile

Note: multiple people can work on this issue - just note which files you are working on in the comments. If anyone would like to take this on, please look at the comments and see which files have been picked up.

nishakm:
Can you hold off on this one for a bit? There is a big merge coming down the pipeline. It should be a couple more weeks.

Vexxorian:
Will [that](https://gist.github.com/Vexxorian/d2c5a33e5f232277fe8beb4a8155e16f) work? Not sure if good or bad idea \\o/

nishakm:
The reason for converting the short names to the full module names is for the sake of readability - you will be able to find the source files faster. So no :)

MattAlp:
@nishakm first timer here, is this being held off as requested or still open? It shouldn't take more than 10-15 minutes

nishakm:
Big merge completed! This issue is up for grabs.

vladn90:
@nishakm Hello! Is issue still open? I'd be glad to work on it :)

nishakm:
@vladn90 it's still open. It's a pretty big undertaking so it's fine if you focus on a couple of files at a time. I'll put this note in the issue.

PR #33, opened 2018-04-30
Convert short names of modules to the full name

vladn90:
Converted short names of modules to the full name to improve readability.
For example, _from utils import dockerfile as df_ >> _from utils import dockerfile_.

Changed global variable _dockerfile_ to _dockerfile_global_, because line
_from utils import dockerfile as df_ was changed to _from utils import dockerfile_.

Resolves issue #29.

Signed-off-by: Vlad Nalimov vlad.nalimov@gmail.com

### DIFF
--- a/common.py
+++ b/common.py
@@ -8,7 +8,7 @@ import logging
 from classes.package import Package
 from classes.notice import Notice
 from classes.command import Command
-from command_lib import command_lib as cmdlib
+from command_lib import command_lib
 from report import formats
 from report import errors
 from report import content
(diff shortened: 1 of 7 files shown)


nishakm:
@vladn90 can you sign-off on your commit message? git commit --amend -s should do the trick

nishakm:
@vladn90 did you run PEP8 checks on your changes?

vladn90:
Thank you for your patience and for the help. I've amended the last commit, signed off on it and ran all the changes through PEP8. Is there anything else?

nishakm:
@vladn90 looks good. Thank you for taking the time!

vladn90:
Thank you and you're welcome. Glad I could help.

Issue #28, opened 2018-04-09
Detect if docker daemon is running


nishakm:
Gracefully exit if the docker daemon is not running

raman-balyan:
I want to help in this issue.

PR #34, opened 2018-05-01
Gracefully exit if the docker daemon is not running

raman-balyan:
This will check whether docker is running or not through the subprocess module. I have used the check_output method which is using "docker ps" command to check whether docker daemon is running or not. If docker is not running then subprocess module raises the ProcessError exception which is being used to raise the customized message on the console and gracefully exit the script.
Resolves: #28

Signed-off-by: Raman Balyan <raman.balyan@gmail.com>

### DIFF
--- a/report/report.py
+++ b/report/report.py
@@ -18,6 +18,7 @@ from classes.package import Package
 from command_lib import command_lib
 import common
 import docker
+import sys
 
 '''
 Create a report
@@ -130,6 +131,11 @@ def generate_report(args, *images):
 def execute_dockerfile(args):
     '''Execution path if given a dockerfile'''
     logger.debug('Setting up...')
+    try:
+        container.docker_command(['docker', 'ps'])
+    except subprocess.CalledProcessError as error:
+        logger.error('Docker daemon is not running: {0}'.format(error.output.decode('utf-8')))
+        sys.exit()
     setup(args.dockerfile)
     dockerfile_parse = False
     # try to get Docker base image metadata
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-### Contributing to the Tern project
+# Contributing to the Tern project
 
 Once again, we hope you have read our [code of conduct](/CODE_OF_CONDUCT.md) before starting.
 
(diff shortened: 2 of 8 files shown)


report/report.py
```diff
@@ -126,10 +127,14 @@ def generate_report(args, *images):
     write_report(report)
 
 
-
 def execute_dockerfile(args):
     '''Execution path if given a dockerfile'''
     logger.debug('Setting up...')
+    try:
+        subprocess.check_output('docker ps', shell=True)
```
nishakm: You could use the docker_command wrapper under utils/container.py. It doesn't use shell=True and takes care of things like logging and checking to see if the user doesn't have root privileges.

report/report.py
```diff
@@ -126,10 +127,14 @@ def generate_report(args, *images):
...
+    except subprocess.CalledProcessError as error:
+        logger.error('Docker daemon is not running: {0}'.format(error.output))
```
nishakm: Is error.output a byte stream or a string?

report/report.py
raman-balyan: thanks for the suggestion. I will update that..

report/report.py
raman-balyan: It is a byte stream and I got my error :)

nishakm: I'm curious then - did you get the 'b' at the beginning of the error string? If yes then you will need to convert the byte stream into a string using error.output.decode('utf-8').

raman-balyan: Yes, I got 'b' and I will update it. Thanks

nishakm: No worries! Normally you would get an error saying you cannot convert a bytestring into a string but format does the !r conversion for you. It's nice to know exactly what types you're working with ^_^

nishakm: Yes, docker_command runs a shell command and raises a CalledProcessError is the command fails so you can put that in a try-except as you have done in report.py
>>> import subprocess
>>> from utils.container import docker_command
>>> try:
...     docker_command(['docker', 'ps'])
... except subprocess.CalledProcessError as error:
...     print(error.output.decode('utf-8'))
...
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

raman-balyan:
@nishakm I have updated the PR according to your comments. Also, I have updated the commit message with the description.

raman-balyan:
@nishakm thanks for bearing with me. I will update that.

PR #38, opened 2018-05-11
Gracefully exit if the docker daemon is not running

raman-balyan:
This will check whether docker is running or not through the docker_command present in container.py file. I have used the "docker ps" command to check whether docker daemon is running or not. If docker is not running then subprocess module raises the ProcessError exception which is being used to raise the customized message on the console and gracefully exit the script.

Resolves: #28

Signed-off-by: Raman Balyan <raman.balyan@gmail.com>

### DIFF
--- a/report/report.py
+++ b/report/report.py
@@ -5,6 +5,7 @@ SPDX-License-Identifier: BSD-2-Clause
 
 import logging
 import subprocess
+import sys
 
 from report import content
 from utils import container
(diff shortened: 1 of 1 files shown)


raman-balyan:
@nishakm I have updated the commit message and moved the 'sys' import at top.

nishakm:
Tested and looks good. Thanks for continuing to work on this PR @raman-balyan! Much appreciated :)

PR #44, opened 2018-05-14
Remove virtual environment creation from documentation

kdestasio:
Remove the lines related to the creation and activation of
a virtual environment. That step is unnecessary if the virtual machine
will be used only for the Tern project.

Signed-off-by: kdestasio <kdestasio@gmail.com>


nishakm:
Looks good! Thanks again!

Issue #45, opened 2018-05-15
Docs: Requesting access for non-vmware users should be more clear


nishakm:
Perhaps some steps for non-vmware users to first request access and then join the slack channel

nishakm:
@kdestasio is working on this

PR #46, opened 2018-05-15
Clarify access requests for non-vmware users

kdestasio:
Resolves issue #45. Put links for non-vmware users first in instructions
to join slack. Create sub-headings for public forum and slack
under the 'Keeping in touch' heading to make options more visible.

Signed-off-by: kdestasio <kdestasio@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,13 +18,14 @@ You can contribute in the following ways:
 * YAML (Tern uses yaml files heavily for data lookup. An overview of YAML can be found [here](http://yaml.org/))
 
 ## Keeping in touch
-
+### Public forum
 If you would like to ask a question or start a discussion, post a topic on the [public forum](https://groups.google.com/forum/#!forum/tern-discussion). You will need to apply to join the group before posting. We will respond to your question or topic within three days unless the post was over the weekend in which case we may take longer to respond. This is our primary communication channel so it is highly likely that you will get a response here.
 
-If you would like to chat live, you can join the [slack channel](https://vmwarecode.slack.com/messages/tern). If you don't have an @vmware.com or @emc.com email, please click [here](https://code.vmware.com/join), enter your email address and click "Request Invite". Although we monitor the channel, we may not respond right away and if the same question was asked on the forum, we will choose to respond there.
+### Slack channel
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,13 +18,14 @@ You can contribute in the following ways:
...
+### Slack channel
+If you would like to chat live, you can join the Slack channel. If you don't have an @vmware.com or @emc.com email, please [follow this link](https://code.vmware.com/join), enter your email address and click "Request Invite". Although we monitor the channel, we may not respond right away and if the same question was asked on the forum, we will choose to respond there. If you are already a member of the slack channel, or if you have an @vmware.com or @emc.com email, you can simply [sign-in to the Slack channel](https://vmwarecode.slack.com/messages/tern).
 
 ## Set-up a virtual machine for the project
-To work on Tern, you may want to set-up a virtual machine that can run Ubuntu. To do so, follow the [instructions here.](/docs/contributing-setup.md)
+To work on Tern, you may want to set-up a virtual machine that can run Ubuntu. To do so, follow [these instructions.](/docs/contributing-setup.md)
```
nishakm: This change is not mentioned in the commit message. Can you add a mention? Even better - can you submit another PR with this change?

nishakm:
@kdestasio there is one change in the file that isn't about changing access to non-vmware users. Can you submit another PR for that change?

CONTRIBUTING.md
kdestasio: @nishakm sure, I can do that. Given that I already submitted one with that change in it, do I just append the last commit message?

nishakm: I think the PR will automatically include two commits if you push a second commit with this change to your fork's master branch. You would have to do a little work on it though:

1. Edit CONTRIBUTING.md to revert this change (basically make it look like the - line again so the change doesn't show up)
2. git add CONTRIBUTING.md, git commit --amend and git push origin master
3. Make the above change in CONTRIBUTING.md
4. git add CONTRIBUTING.md, git commit -s (edit subject and body) and git push origin master
5. Edit the PR subject to say something generic about your 2 commits

There is some git-fu that can separate the different chunks without you having to manually revert the change but the above method is waaay easier.

nishakm:
Cherry-picked and pushed in https://github.com/vmware/tern/commit/686a7a9c645417ff250ca22457d7cd9b0d6ca0ec. Thank you!!

PR #49, opened 2018-05-15
Change to rootfs.unmount_rootfs

snmpboy:
Added a 'num_layers' argument to the unmount_rootfs function in the rootfs module.


nishakm:
Looks good! Merging...

Issue #63, opened 2018-06-21
Proposal: Rather than have a default shell, find the shell binary in the base OS layer


nishakm:
Similar to the package manager discovery, use the shell that comes with the base layer if no shell is listed.

nishakm:
Closed in #67

Issue #61, opened 2018-06-18
Change base.yml main keys to be package manager (or other identifiable binary) names rather than repo-tags


nishakm:
The repo-tag is a container registry construct and could change depending on the registry. So look up the base commands based on the package manager rather than the repo tag.

Eg:
```
vmware/photon:
  latest: '1.0'
  tags:
    '1.0':
      shell: '/usr/bin/bash'
      names:
```
to

```
tdnf:
  repo: 'vmware/photon'
  tags: ['latest', '1.0']
  shell: '/usr/bin/bash'
  names:
```
This is to support #51

nishakm:
@snmpboy Are you free to work on this issue? It's on a deadline so no pressure if you cannot commit.

snmpboy:
@nishakm Kind of slammed at the moment.  Working and studying for https://www.cisco.com/c/en/us/training-events/training-certifications/exams/current-list/300-560-npdev.html

nishakm:
@snmpboy no worries! Good luck with the networking study - that's intense!

nishakm:
repo and tag may not be a good fit for this as they are representative of container distribution and not the contents. Values can be stored in the database instead.

nishakm:
Closed in #67

Issue #60, opened 2018-06-18
Create module to look up a container OS's package manager


nishakm:
Refer to #51 for more information.

Implementation looks something like this:

1. We need a list of package managers and their names in the form of a python dictionary
[{path: /usr/bin/tdnf, name: tdnf},...]
this can live in a .yml file called 'pkg_mgr.yml' in the command_lib folder
2. A module in common.py for base OS discovery. It should return the name of the package manager

nishakm:
@tojo13 Do you think you can take this on? It's under a release deadline so no worries if you cannot commit.

tojo13:
Sure, I can give it a go!

nishakm:
@tojo13 thanks! It's yours. I'll be working in parallel, but that stuff is not going to work until your PR comes in :)

nishakm:
I will be creating a stub module in common.py:
get_base_bin(base_layer)

base_layer is the layer object. The tar file associated with the layer is in base_layer.tar_file

nishakm:
@tojo13 I am at the point where I will need this module. So I'll be assigning it to myself. Sorry! There'll be more stuff to do and hopefully it will align when you have more time. Thanks for offering to help!

tojo13:
No worries. Sorry I couldn't get this done earlier. Good luck!

nishakm:
Closed in #67

Issue #51, opened 2018-05-15
Proposal: Replace package manager lookup with facts discovery


nishakm:
For a container base OS there is no indication on what kind of package manager it supports. Ansible has a module that does the discovery of the package manager: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts/system/pkg_mgr.py
Do something similar for container base filesystem.

nishakm:
Resolved in #67

PR #67, opened 2018-07-14
Changes to work with bare filesystem layers

nishakm:
This PR should resolve the following issues:  #5 #17 #51 #60 #61 #63


nishakm:
I have tested this with the sample Dockerfiles

Issue #78, opened 2018-08-20
Deep comparison of Package objects


nishakm:
Currently, comparisons of Package objects are only done based on name
```
p1 = Package(pkg_name1)
p2 = Package(pkg_name2)
if p1.pkg_name1 == p2.pkg_name2:
    # we assume this is enough to say that the two packages are equal but they are not!
```

Add a method (`def is_equal(self, another_pkg_obj)`) that will do a deep comparison of the package object
```
if self.name == another_pkg_obj.name and self.license == another_pkg_obj.licence ...
    return True
else
   return False
```

nishakm:
@johnmm Do you think you can work on this one? It's quite straightforward.

nishakm:
@johnmm do you still want to work on this? I'll probably need to take it off your hands in a week or so.

johnmm:
@nishakm Sorry, I have been swamped with school work and regular work lately. I'm just getting back around to this.

I have made and unit tested the is_equal function and I'm trying to identify all the areas that need to be stepped up. Please let me know if this list seems complete to you.

Image_Layer.remove_package and Image_Layer.add_package seem to be the only direct impacts, but only remove_package will have outside impacts.

Changing remove_package impacts common.update_master_list which will ripple out to report.analyze_docker_image.

My proposed fix for this is to add a function to Image_Layer, get_packages, which will return the list of packages for the given Image_Layer. This means that the master_list from report.analyze_docker_image will be a list of package objects instead of package names. This should hopefully minimize impacts and allow us to keep most of the current logic in place.  

Tests for all of these areas will also need to be updated for this change.

If this solution sounds acceptable to you, I should be able to submit a pull request this week.

nishakm:
@johnmm, don't worry! The project has a lot of work needed, and if not this one then there will be another one. I am so glad you want to put your time and effort into this and greatly appreciate it.

The is_equal function for the Package class will not be used by anything else in the project, so this change can stand by itself without breaking the code. If you are in doubt, just run the code against a docker image. You will have to install docker and pull and image from dockerhub first. See the README.md for a link to Docker's install page or CONTRIBUTING.md to see how to set up a Vagrant image.

At this point the unit test for Package may be failing (I have not paid attention to them yet). If you have a patch to also add a unit test for is_equal, that would be awesome!

[...]

PR #84, opened 2018-09-25
Added Deep Comparison to Package Object

johnmm:
Used Package.To_Dict to compare two objects so that if
new attributes are added to the Package object, this code
will not be impacted. This commit also added relevant
tests for the function and removed a Add_Notice tests
which no longer exists in the Package object.

Resolves: #78
Signed-off-by: John Miller <jmille40@gmu.edu>

### DIFF
--- a/tests/test_class_package.py
+++ b/tests/test_class_package.py
@@ -6,7 +6,6 @@ SPDX-License-Identifier: BSD-2-Clause
 import unittest
 
 from classes.package import Package
-from classes.notice import Notice
 
 
 class TestClassPackage(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_class_package.py
```diff
@@ -61,6 +50,18 @@ def testToDict(self):
...
+        p.license = 'TestLicense'
+        p.version = '1.0'
+        p.src_url = 'TestUrl'
+        self.package.license = 'TestLicense'
+        self.package.version = '2.0'
+        self.package.src_url = 'TestUrl'
+        self.assertFalse(self.package.is_equal(p))
+        p.version = '2.0'
+        self.assertTrue(self.package.is_equal(p))
+
```
nishakm: Thank you! This looks good.

nishakm:
This is great! Just one small request to add documentation to your new method and I can merge :)

johnmm:
Updated! Also added the missing commentary at the top of the class as well.

nishakm:
Excellent! Could you update your commit stating that you added missing commentary at the top of the class?

johnmm:
Sorry about that! Cleaned up the commit message as well.

Issue #90, opened 2018-09-27
Add to_dict method to docker_image class


nishakm:
The image class has a `to_dict` method which can be used by the derived class in `docker_image.py`.

The dictionary should have properties: repotag (str), repotags (list) and history (dict) along with the superclass's dictionary.

Example of how this would be implemented:
```
super_dict = super().to_dict()
super_dict.update({'repotag': self.repotag})
...
return super_dict
```
This is work towards #70

esantoro:
Can I take this ?

nishakm:
@TileThePlane is working on this

Issue #87, opened 2018-09-27
Fill to_dict method in Image class


nishakm:
See `classes/package.py` and `classes/image_layer.py` as examples.

to_dict should create a python dictionary representation of the Image object. Something like this:
```
{image: {id: <id>,
              name: <name>,
              tag: <tag>,
              manifest: <manifest dict>,
              config: <config dict>,
              layers: <result of each layer's to_dict method appended to a list>
}
```
For the layers, create an empty list and for each layer in self.layers append the result of layer.to_dict to the list

There is already a stub method in `classes/image.py` that returns an empty dictionary. This is to allow for this issue to be worked on independently of the others. Use that method to create the above dictionary.

IpshitaC:
Hey @nishakm 
Can I take this up as a part of hacktoberfest?

nishakm:
@IpshitaC yup! Go for it!

nishakm:
@IpshitaC if you see a bunch of updates messages from github: that's my bad. There's a related PR already in that references this issue but is not for this issue, so it's still yours to work on. Hope to see a PR from you soon!

IpshitaC:
Sure, I'll send in a PR soon!

nishakm:
Fixed in https://github.com/vmware/tern/commit/8ab17e6aff0befc5704c2e60e8c41312ef7613aa

PR #93, opened 2018-10-01
Fix issue #90

esantoro:
These commits fix issue #90

### DIFF
--- a/classes/docker_image.py
+++ b/classes/docker_image.py
@@ -162,3 +162,13 @@ class DockerImage(Image):
             raise
         except IOError:
             raise
+
+    def to_dict(self):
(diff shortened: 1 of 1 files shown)


esantoro:
Also fixes issue #87.

Please validate.

classes/docker_image.py
```diff
@@ -162,3 +162,13 @@ def load_image(self):
             raise
         except IOError:
             raise
+
+    def to_dict(self):
+        '''Returns a dictionary reflecting the properties of the image, including Docker-specific ones'''
+        dictview = super.to_dict(self)
+        dictview.update('repotags', self.repotags())
+        dictview.update('repotag', self.repotag())
+        dictview.update('history', self.history())
```
nishakm: I don't think you need () for a class property...

Issue #103, opened 2018-10-09
Add unit test for NoticeOrigin class


nishakm:
See `classes/notice_origin.py` for documentation, properties and methods of this class. See `tests/test_class_notice.py`, `tests/test_class_package.py` and `tests/test_class_image_layer.py` for examples on how to write a unit test for this class.

- Test instance to see if all the default values for the properties are set when instantiating
- Test addition and removal of a Notice object
- Test print_notices and to_dict methods

This may take a little time to complete.

nishakm:
@ericcheatham Yes! Go for it!

PR #107, opened 2018-10-10
Add unit test for `notice_origin`

ericcheatham:
This PR also fixes a few code issues in the NoticeOrigin class

This PR addresses #103

Signed-off-by: Eric Cheatham ericjcheatham@gmail.com

### DIFF
--- a/classes/notice_origin.py
+++ b/classes/notice_origin.py
@@ -50,10 +50,10 @@ class NoticeOrigin(object):
                 warnings = warnings + notice.message
             if notice.level == 'error':
                 errors = errors + notice.message
-            if notice.level == 'hints':
+            if notice.level == 'hint':
                 hints = hints + notice.message
-        notice_msg = formats.notice_format(
-            origin=self.origin, info=info, warnings=warnings, errors=errors,
+        notice_msg = formats.notice_format.format(
+            origin=self.origin_str, info=info, warnings=warnings, errors=errors,
             hints=hints)
         return notice_msg
 
--- a/tests/test_class_notice_origin.py
+++ b/tests/test_class_notice_origin.py
@@ -1,0 +1,117 @@
+'''
+Copyright (c) 2018 VMware, Inc. All Rights Reserved.
+SPDX-License-Identifier: BSD-2-Clause
+'''
+
+import unittest
+
+from report import formats
+
+from classes.notice import Notice
+from classes.notice import NoticeException
+from classes.notice_origin import NoticeOrigin
+
+
+class TestClassNoticeOrigin(unittest.TestCase):
+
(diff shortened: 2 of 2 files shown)


tests/test_class_notice_origin.py
```diff
@@ -0,0 +1,117 @@
...
+        self.notice_hint = Notice("hint", "hint")
+
+        self.notices = [
+            self.notice_info,
+            self.notice_warn,
+            self.notice_errr,
+            self.notice_hint
+        ]
+
+        self.notice_origin = NoticeOrigin(origin_str='origin_str')
```
nishakm: I don't think you need to specify the key and value here. `NoticeOrigin('origin_str')` is enough. Only use it when there are preset arguments. Eg: `def options(required, optional=None)` so when using this without the extra argument you will do `options('required')` and with the extra argument`options('required', optional='ok')`

classes/notice_origin.py
```diff
@@ -50,10 +50,10 @@ def print_notices(self):
                 warnings = warnings + notice.message
             if notice.level == 'error':
                 errors = errors + notice.message
-            if notice.level == 'hints':
+            if notice.level == 'hint':
```
nishakm: Good catch!

classes/notice_origin.py
```diff
@@ -50,10 +50,10 @@ def print_notices(self):
...
                 hints = hints + notice.message
-        notice_msg = formats.notice_format(
-            origin=self.origin, info=info, warnings=warnings, errors=errors,
+        notice_msg = formats.notice_format.format(
+            origin=self.__origin_str, info=info, warnings=warnings, errors=errors,
```
nishakm: Could you change this to `self.origin_str`? There is a property getter defined for it. Thanks!

nishakm:
@ericcheatham Thanks! This is a lot of work and I appreciate it! Could you elaborate on your work in your commit message?
```
This PR also fixes a few code issues in the NoticeOrigin class <-- could you elaborate on the code issues? Mismatch in key 'hint', using incorrect property 'origin' etc?

This PR addresses #103 <-- replace with 'Resolves #103' 

Signed-off-by: Eric Cheatham <ericjcheatham@gmail.com>
```
There are a couple more changes needed in the file but other than that, the PR looks good.

When you update, please amend this commit rather than adding a new commit. Thanks again for your work!

tests/test_class_notice_origin.py
ericcheatham: I can change that. It's a bit of a force of habit for me. 👍

nishakm:
Looks good to me 👍 Thanks again!

PR #116, opened 2018-11-01
Remove existing_layers, check from save_to_cache

laurieroy:


common.py
```diff
@@ -61,10 +61,9 @@ def load_from_cache(layer):
 
 def save_to_cache(image):
     '''Given an image object, check against the cache to see if a layer id
-    exists. If no then save the layer and package list to the cache'''
```
nishakm: We're not doing any checks whether a layer id exists in the cache at all. So "Given an image object, save all the layers to the cache"

common.py
```diff
@@ -61,10 +61,9 @@ def load_from_cache(layer):
...
-    existing_layers = cache.get_layers()
+    exists. If no then save the package list to the cache'''
     for layer in image.layers:
-        if layer.fs_hash not in existing_layers and layer.packages:
+        if layer.fs_hash not in layer.packages:
```
nishakm: We're only checking if the layer object has a non-empty package list. You can do that in python by just checking to see if layer.packages (`packages` is a member layer and it is a python list) exists. Like this: `if layer.packages:`

laurieroy:
Sorry, thought this submitted days ago

nishakm:
@laurieroy Yes, you had submitted it, but I cannot merge the PR until you update it with the changes mentioned. Once merged, your change will be part of the code base and you will be one of the contributors to the project! It would be awesome if you could edit your change so I can merge it.

Issue #113, opened 2018-10-31
Load from cache only when redo=False


nishakm:
We want to allow a user to repopulate the cache with new data for any existing container layers i.e. redo the cache for just those layers. See #27 

Update the `load_from_cache` def in common.py to take an argument called `redo` that defaults to 'False' (means the default operation should be to not redo the cache)
```
 35 def load_from_cache(layer):  <-- add new argument redo=False                                        
 36     '''Given a layer object, check against cache to see if that layer id exists 
 37     if yes then get the package list and load it in the layer and return true.  
 38     If it doesn't exist return false  <-- add documentation for the extra functionality here                                          
 39     Add notices to the layer's origins matching the origin_str'''               
 40     loaded = False                                                              
 41     origin_layer = 'Layer: ' + layer.fs_hash[:10]                               
 42     if not layer.packages:  <-- add and not redo                                                     
 43         # there are no packages in this layer <-- and we are not repopulating the cache                                 
 44         # try to get it from the cache                                          
```
This is work toward #27

nishakm:
@IpshitaC would you be interested in taking this one on?

nishakm:
@laurieroy is working on this

nishakm:
@laurieroy are you still working on this?

Issue #27, opened 2018-04-06
Tern does not account for modifications to existing methods


nishakm:
Say if tern processes package information and could not get one set of metadata, it will store the package information in the cache with empty data. If next time around, there is an addition to the command library that allows tern to get said metadata, it will not run it because it will read the incomplete information from the cache.

1. When loading form cache, check for missing metadata. If it finds missing metadata, then run through the retrieval steps again
2. When saving to the cache, do a deep comparison of all the package metadata rather than the layer. In python one can override the class's inherited __eq__(self, other) to do the check. If the layers are not equal, save the layer with the packages again

nishakm:
We can just add an option in the command line (--redo, -r) to overwrite the data that is in the cache if it exists.
This would mean writing a function to erase cached data for a particular layer.
This would also mean rewriting the flow to check if the redo flag is not set to then load data from the cache, and if yes then erase data from the cache before loading (load_from_cache function)

PR #130, opened 2018-11-15
Load from cache only when redo=False

nishakm:
Add redo=False to load_from_cache so option to re-populate the cache
when layers are changed; default is to not redo the cache.
Resolves: #113   See: #27

Signed-off-by: Laurie Roy <laurieroy.dev@gmail.com>

### DIFF
--- a/common.py
+++ b/common.py
@@ -32,16 +32,16 @@ def get_shell_commands(shell_command_line):
     return cleaned_list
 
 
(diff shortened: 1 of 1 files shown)


nishakm:
@laurieroy if you don't mind, I created the PR for you. The changes look good. Once tests are done, I can merge

PR #172, opened 2019-02-15
Refactor: Move all source code to src folder

### LINKED ISSUE #115: Move all code to a folder called src
Moving all the code to a folder called src accomplishes these things:
- It consolidates all the code into one place
- It makes the project look less sprawling on github

This would mean moving the folders report, utils, tests and tools into a folder called src
This would also mean modifying the docs to reflect this change.

README.md stays where it is.

This would cause CI/CD to break so that .travis.yml will have to reflect the change as well

nishakm:
As support grows for the project, the higher level repository
is getting cluttered with enabling scripts for various CI/CD and
deployment schemes. Moving all source code to `src` folder so
the project looks clean. Added changes to higher level README
to match this change

- Moved all source code to `src` folder
- Used WORKDIR in Dockerfile to point to `src` folder
- Modified README.md to show commands to run in the new setup
- Updated .travis.yml with new script to run

Resolves #115


nishakm:
Follow up PR will dismiss the Precaution issues (mostly from using blacklisted module `subprocess` which, honestly, this project can't do without).

PR #190, opened 2019-03-22
Make pypi installable

### LINKED ISSUE #136: How about making a release on PyPI?
Being able to `pip install tern` would be so much more user-friendly.

SantiagoTorres:
This should fix #136

There are two small fixes in that may not be completely related to this PR:

- Do not fail with ENOENT when cache.yml doesn't exist
- Do not fail when not running inside a git repository.
- Make the Docker image use the pip installable version

I included these fixes because I think they are necessary to actually use the
pip-installable version, but I can always remove them from this PR if necessary
:)


SantiagoTorres:
I think this PR is easier to review on a commit-by-commit basis. The tests may fail as they need to be aware of the namespace changes though...

SantiagoTorres:
[asciicast](https://asciinema.org/a/OxJhjvVN3C7k7YmzLEtzbwChE)

nishakm:
Wow! This is way cool! Do you mind adding some docs on how to install via pip? Since there isn't any build and release pipeline in place for the project yet, I think for now something like `git clone project`, `python3 -m venv test`, `cd test`, `pip install path/to/git/clone`, `tern -l report -i archlinux/base > report.txt` should be OK for now.

nishakm:
Why did you remove cache.yml again?

SantiagoTorres:
> Why did you remove cache.yml again?

Oh, I was intending of later using the pkg_resource cache module to be able to put it on a cross-OS location that's guaranteed to be read-write-able. There's more to come on this PR :smile:

tern/utils/cache.py
```diff
@@ -27,6 +27,11 @@
 def load():
     '''Load the cache'''
     global cache
+
+    # small guard to stop ENOENT if the cache is not missing
```
nishakm: :D Not everyone knows what ENOENT means (now I'm curious on what your traceback looks like and if you're using any tools to view it).

tern/__main__.py
```diff
@@ -36,7 +36,7 @@
 logger.addHandler(log_handler)
 
 
-def main(args):
+def do_main(args):
```
nishakm: I'm actually not familiar with setuptools so I had to do a bit of reading up. Is this change necessary for this PR? Can it exist in a follow-up one?

setup.py
```diff
@@ -0,0 +1,40 @@
...
+    'Operating System :: POSIX',
+    'Operating System :: POSIX :: Linux',
+    'Operating System :: MacOS :: MacOS X',
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: Implementation :: CPython',
+    'Topic :: Software Development'
+  ],
+  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests",
+      "debian"]),
+  install_requires=["pyyaml"],
```
nishakm: Also needs "docker". You may have added that in a later commit...

setup.py
```diff
@@ -0,0 +1,40 @@
...
+  author="VMWare Inc",
+  author_email="FIXME@FIXWHAT.THEEMAIL",
+  url="https://github.com/vmware/tern/",
+  description=("An inspection tool to find the metadata of the packages" 
+    " installed in a container image"),
+  long_description=_read_long_desc()
+  license="BSD-2.0",
+  keywords="Distribution, Container, Cloud-Native",
+  classifiers = [
+    'Development Status :: 4 - Beta',
```
nishakm: Should be `3 - Alpha` and hopefully not Beta before this merges :)

setup.py
```diff
@@ -0,0 +1,40 @@
...
+
+def _read_long_desc():
+    with open("README.md") as fp:
+        return fp.read()
+
+setup(
+  name="tern",
+  version=version,
+  author="VMWare Inc",
+  author_email="FIXME@FIXWHAT.THEEMAIL",
```
nishakm: I can cherry-pick and replace these values if needed

setup.py
```diff
@@ -30,11 +35,10 @@ def _read_long_desc():
     'Programming Language :: Python :: Implementation :: CPython',
     'Topic :: Software Development'
   ],
-  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests",
-      "debian"]),
-  install_requires=["pyyaml"],
+  packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"]),
+  install_requires=_get_requirements(),
```
nishakm: OK, this is where you added the additional requirements. Looks good.

setup.py
```diff
@@ -30,11 +35,10 @@ def _read_long_desc():
...
   test_suite="tests.runtests",
   entry_points={
-    "console_scripts": ["tern = tern.__init__.py:main"]
+    "console_scripts": ["tern = tern.__main__:main"]
```
nishakm: I'm still trying to get my head around how `entry_points` work. Will come back to this later.

tern/utils/general.py
```diff
@@ -77,7 +77,11 @@ def get_git_rev():
     '''Assuming we are operating within a git repository, get the SHA
     of the current commit'''
     command = ['git', 'show', '--format=%H', 'HEAD']
-    output = subprocess.check_output(command)  # nosec
+    try:
+        output = subprocess.check_output(command)  # nosec
+    except subprocess.CalledProcessError as e:
+        # FIXME
+        output = "HEAD"
```
nishakm: Hmm...we might want to do away with the git sha1 reporting of the version. Would it be possible to read the version from setup.py for example?

tern/utils/cache.py
SantiagoTorres: Ah! sorry I'll rephrase and make it clearer :+1:

tern/__main__.py
SantiagoTorres: It's necessary for setuptools. entrypoints require a function to call to create the executable stub, so I moved the if __main__ into a function called main. I wonder if another name would've been more convenient/descriptive...

setup.py
SantiagoTorres: Ah! I'll update this :+1:

setup.py
SantiagoTorres: Yeah, I wasn't sure what to put in here...

tern/utils/general.py
SantiagoTorres: Yep, another option would be to have a version file within the module and load it within setup.py and at runtime.

tern/__main__.py
nishakm: `entry_point`?

tern/utils/general.py
nishakm: I like that idea better!

nishakm:
@SantiagoTorres after you're done with this PR, would you mind if I rearrange the commits a bit to make the history more clear? I think I understand how this is coming together. I'll also save you some commit documentation time.

tern/__main__.py
SantiagoTorres: [yep](https://github.com/vmware/tern/pull/190/files/edaff38ab10761ae2b2c3524ea37fdbfe684f477#diff-2eeaed663bd0d25b7e608891384b7298R37), it is of the form [name-of-executable]:[path-to-module.function]

(p.s. sorry I'm in a 60% keyboard and I can't add any syntactic sugar to my comments rn :[ )

SantiagoTorres:
Not at all! I'll ping you when I'm done and you can take over afterwards. Thanks!

nishakm:
> > Why did you remove cache.yml again?
> 
> Oh, I was intending of later using the pkg_resource cache module to be able to put it on a cross-OS location that's guaranteed to be read-write-able. There's more to come on this PR 😄
OK, I finally got to reading about what cache in the `pkg_resources` context. Unless I misunderstood that cache is a place in the filesystem to store python eggs, this `cache.yml` is not that kind of cache. It's a stand-in for a key-value store. In the future, it may be replaced by a real key-value store or some relational database like postgres. So I'd like it to stay there as a placeholder.

SantiagoTorres:
@nishakm just to be sure, should I ammend the existing commits with the fixes or should I add them on top?

nishakm:
> @nishakm just to be sure, should I ammend the existing commits with the fixes or should I add them on top?

If you can amend the commits, that'll be great! I thought you were adding things on top 😄

SantiagoTorres:
>  OK, I finally got to reading about what cache in the pkg_resources context. Unless I misunderstood that cache is a place in the filesystem to store python eggs, this cache.yml is not that kind of cache. It's a stand-in for a key-value store. In the future, it may be replaced by a real key-value store or some relational database like postgres. So I'd like it to stay there as a placeholder.

Aha, ok, then I won't add much on this. The issue with leaving it as a resource is that it's possible the cache file is written to a non-write-able location (e.g., only writeable by the superuser). Should we leave as it is? (i.e., if there is no cache file in the current directory load an empty dictionary from memory?

nishakm:
Yup! Leave the check for no file existing but keep the file in the repo. Thank you!

SantiagoTorres:
@nishakm I pushed everything. I'm getting the DCO error again. I'm starting to wonder if it's related to me having multiple emails registered under my key or so (or that this computer has an old gitconfig?)

nishakm:
Looks like it's multiple emails but it's supposed to compare the `Author` field with your `Signed-off-by` in the commit msg. So even with different gitconfigs, it should still work. Let me take a look.

nishakm:
OK, I think I found the offending commit:
```
commit 8fb95c4a6aaec281271f222c13285badc2c98753
Author: Santiago Torres <torresariass@gmail.com>
Date:   Fri Mar 22 19:01:26 2019 -0400

    utils/cache: catch non-existing cache
    
    The cache.yml may not exist on disk. If that's the case, do not load it.
    
    Signed-off-by: Santiago Torres <santiago@archlinux.org>
```

SantiagoTorres:
Ah ok, sorry about that let me correct this. I'm amending that commit anyway :+1:

SantiagoTorres:
Aha, I think this should do it?

Dockerfile
```diff
@@ -6,15 +6,15 @@ FROM photon:3.0
...
 
 # copy tern repo into root
-COPY . .
+COPY . tern
 
 # install app dependencies
-RUN pip3 install --upgrade pip && pip3 install -r requirements.txt
+RUN pip3 install --upgrade pip && pip3 install ./tern
 
 WORKDIR /src
```
nishakm: Remove this line to make the script run in the container.

nishakm:
@SantiagoTorres One last fix and I think we're ready to merge this. Thanks again for all your work!

nishakm:
@SantiagoTorres I merged all your changes via command line: https://github.com/vmware/tern/commit/540f1e81cc12f305aa07f317fb5e53bd38c255a0
Wish github could actually detect that so it looks like your PR is merged here.
Thanks again! This was a really big item <3

SantiagoTorres:
Awesome! Thanks!

PR #194, opened 2019-03-27
Resolve code check isues for src/utils/rootfs.py

mkrohan:
The issues raised by prospector for "src/utils/rootfs.py" are:

    src/utils/rootfs.py
      Line: 14
        pylint: pointless-string-statement / String statement has no effect
      Line: 52
        pylint: logging-not-lazy / Specify string format arguments as logging function parameters (col 4)
      Line: 183
        pylint: try-except-raise / The except handler raises immediately (col 4)

This change resolves these issues.  The final warning is suppressed
as this is a coding pattern common for Tern (could, alternatively,
add a log message).


src/utils/rootfs.py
```diff
@@ -1,17 +1,18 @@
...
+Operations to mount container filesystems and run commands against them
+"""
+
 import hashlib
 import logging
 import os
 import subprocess  # nosec
 import tarfile
 
 from . import constants
```
nishakm: When I ran prospector on this, I got this message: `pylint: relative-beyond-top-level / Attempted relative import beyond top-level package`

The file has since been moved to the `tern` folder and is now under a module name. So this can be solved using:
Using `from tern.utils import constants`

nishakm:
@mkrohan Thanks again for the PR! Can you rebase your commit on top of master and resubmit with the change I mentioned? I can merge the PR then.

nishakm:
#195 addresses this without all the duplicate commits.

PR #196, opened 2019-03-29
Resolve code check isues for src/utils/metadata.py

mkrohan:
The issues raised by prospector for "src/utils/metadata.py" are:

    tern/utils/metadata.py
      Line: 8
        pylint: import-error / Unable to import 'utils.constants'
      Line: 12
        pylint: pointless-string-statement / String statement has no effect

This change resolves these issues.  Also added a encoding comment
to rootfs.py (update to previous commit).

### DIFF
--- a/tern/utils/metadata.py
+++ b/tern/utils/metadata.py
@@ -1,15 +1,14 @@
-'''
-Copyright (c) 2017 VMware, Inc. All Rights Reserved.
-SPDX-License-Identifier: BSD-2-Clause
-'''
+#
+# Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
+"""
+Container metadata operations
+"""
 
 import os
 import shutil
-from utils.constants import temp_folder
-
-'''
-Container metadata operations
-'''
+from tern.utils.constants import temp_folder
 
 
 def clean_temp():


tern/utils/rootfs.py
```diff
@@ -1,4 +1,4 @@
-#
+# -*- coding: utf-8 -*-
```
nishakm: Any possibility of adding encoding enforcement en mas in another PR?

tern/utils/metadata.py
```diff
@@ -1,15 +1,14 @@
-'''
-Copyright (c) 2017 VMware, Inc. All Rights Reserved.
-SPDX-License-Identifier: BSD-2-Clause
-'''
+# -*- coding: utf-8 -*-
```
nishakm: This is new. Did the linter want to enforce utf-8 encoding?

nishakm:
@mkrohan Thanks for working on this! Small nit on the spelling of `issues` in the comment.

I like the idea of enforcing utf-8 encoding (didn't know it was possible in Python). Would you mind rolling those changes in another PR related to that issue?

tern/utils/metadata.py
mkrohan: No, just part of my standard header!  Will remove and add a new issue.

mkrohan: BTW, it's more a hint for editors, not sure if Python respects it.

nishakm: Apparently the interpreter does per [PEP263](https://www.python.org/dev/peps/pep-0263/). It's something I didn't know of and I think it would be a good addition.

mkrohan: Yep, just found that and added https://github.com/vmware/tern/issues/197

PR #198, opened 2019-04-01
Resolve code check issues for tern/utils/general.py

mkrohan:
The issues raised by prospector for "tern/utils/general.py" are:

    tern/utils/general.py
      Line: 66
        pylint: len-as-condition / Do not use `len(SEQUENCE)` to determine if a sequence is empty (col 15)
      Line: 86
        pylint: unidiomatic-typecheck / Using type() instead of isinstance() for a typecheck. (col 11)
      Line: 91
        pylint: unused-variable / Unused variable 'e' (col 4)

This change resolves these issues.

Tests: Tern on debian:buster produced an output of 22161 lines

### DIFF
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -1,7 +1,7 @@
-'''
-Copyright (c) 2017-2018 VMware, Inc. All Rights Reserved.
-SPDX-License-Identifier: BSD-2-Clause
-'''
+#
+# Copyright (c) 2017-2018 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
 import os
 import random
 import re
@@ -13,7 +13,7 @@ from tern import Version
 
 from contextlib import contextmanager
 
-from . import constants
+from tern.utils import constants
 
 
 # from https://stackoverflow.com/questions/6194499/pushd-through-os-system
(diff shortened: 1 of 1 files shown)


tern/utils/general.py
```diff
@@ -1,7 +1,7 @@
-'''
-Copyright (c) 2017-2018 VMware, Inc. All Rights Reserved.
-SPDX-License-Identifier: BSD-2-Clause
-'''
+#
+# Copyright (c) 2017-2018 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
```
nishakm: Sorry for not noticing this before, but prospector doesn't complain about this string for me. Why change the `''' '''` commit into `#` commits?

nishakm:
@mkrohan Have you thought about submitting a PR with a list of commits touching one file at a time instead of waiting to have each commit merged?

tern/utils/general.py
mkrohan: The `"""` or `'''` leading string is the module document string printed if you were to do a `help(tern.utils.general)` and likely will eventually be pulled into code documentation, e.g., via autodoc.  The copyright and license is not really documentation.  It's a statement about the source file.  This file should have a doc string.  For this cycle of changes, I'm just addressing surface issues.  Makes sense?

nishakm: Yup, makes sense. Didn't realize this file didn't have a doc string. I'll merge it.

PR #200, opened 2019-04-02
Prospector utils

mkrohan:


tern/utils/container.py
```diff
@@ -35,7 +35,7 @@
 except IOError:
     logger.critical("Docker daemon not running")
     raise Exception("Critical Error using Docker API. See logs for details")
-except OSError:
+except OSError:  # pylint: disable=duplicate-except
```
nishakm: I wonder why pylint caught this as a duplicate-except. Is it because they raise the same general exception? If that's the case we can fix it by raising IOError and OSError with that sentence.

tern/utils/container.py
```diff
@@ -63,25 +63,25 @@ def check_container():
 
 def check_image(image_tag_string):
     '''Check if image exists'''
-    logger.debug("Checking if image {} is available on disk...".format(
-        image_tag_string))
+    logger.debug("Checking if image \"%s\" is available on disk...",
+        image_tag_string)
```
nishakm:
```
>>> logger.debug("can I do %s ?", "this")
can I do this ?
```
Why are the escapes needed?

tern/utils/container.py
```diff
@@ -63,25 +63,25 @@ def check_container():
...
     try:
         client.images.get(image_tag_string)
-        logger.debug("Image {} found".format(image_tag_string))
+        logger.debug("Image \"%s\" found", image_tag_string)
```
nishakm: Refer to above comment

tern/utils/container.py
```diff
@@ -63,25 +63,25 @@ def check_container():
...
         return True
     except docker.errors.ImageNotFound:
         return False
 
 
 def pull_image(image_tag_string):
     '''Try to pull an image from Dockerhub'''
-    logger.debug("Attempting to pull image {}".format(image_tag_string))
+    logger.debug("Attempting to pull image \"%s\"", image_tag_string)
```
nishakm: Refer to above comment

tern/utils/container.py
```diff
@@ -63,25 +63,25 @@ def check_container():
...
     try:
         client.images.pull(image_tag_string)
-        logger.debug("Image {} downloaded".format(image_tag_string))
+        logger.debug("Image \"%s\" downloaded", image_tag_string)
         return True
     except docker.errors.ImageNotFound:
-        logger.warning("No such image: {}".format(image_tag_string))
+        logger.warning("No such image: \"%s\"", image_tag_string)
```
nishakm: Refer to above comment for these two instances.

tern/utils/container.py
mkrohan: Haven't dug into it, but it could be that IOError and OSError are not dups on Windows?

tern/utils/container.py
```diff
@@ -92,7 +92,7 @@ def build_container(dockerfile, image_tag_string):
     if not check_image(image_tag_string):
         try:
             client.images.build(path=path, tag=image_tag_string, nocache=True)
-        except(TypeError, docker.errors.APIError, docker.errors.BuildError):
+        except (TypeError, docker.errors.APIError, docker.errors.BuildError):  # pylint: disable=try-except-raise
```
nishakm: Could you suggest another way of raising an error that includes the error message? Maybe:
```
try:
    client.images.build...
except (TypeError, docker.errors.APIError, ...) as e:
    raise type(e)
```
?

tern/utils/container.py
```diff
@@ -150,5 +150,5 @@ def extract_image_metadata(image_tag_string):
         os.remove(temp_tarfile)
         if not os.path.exists(temp_path):
             raise IOError('Unable to untar Docker image')
-    except docker.errors.APIError:
+    except docker.errors.APIError:  # pylint: disable=try-except-raise
```
nishakm: See above comment.

tern/utils/container.py
mkrohan: It's generally good practice to include non-numeric values in quotes when logging, etc.  It makes it a little easier when scanning logs to see what value is key.

tern/utils/container.py
mkrohan: I'm not sure what your intent is here?  Do you want to raise a general TernError exception but pull the message from the exception raised?

tern/utils/container.py
nishakm: Oh I see. Here the log message looks like this: `Checking if image "blah" is available on disk...`

mkrohan: Yep.

tern/utils/container.py
nishakm: No, I want to bubble-up the caught exception whether it is `TypeError` or one of the `docker.errors` with the associated error message.

tern/utils/container.py
mkrohan: Since you have
```python
    try:
        ...
    except ...:
        raise
```
The try/catch doesn't really serve any purpose, all exceptions raised will bubble up.  I was interpreting it as a place-holder allowing you add a log message in special cases?

nishakm: The try-except here is to make sure tern exits gracefully if for some reason an API call to Docker doesn't work. This is something that is littered throughout the code as tern uses syscalls via `subprocess` quite liberally. Rather than squash the pylint error for these try-except-raises, I would like to find the "right" way to write it.

nishakm: Perhaps fixing the try-except-raises linting errors falls under technical debt. I'll create an issue for it so it's not overhead for just clearing pylint errors.

mkrohan: OK, sounds good.  Will also try thinking about a clean pattern.

PR #203, opened 2019-04-04
Prospector classes

mkrohan:


tern/classes/docker_image.py
```diff
@@ -25,7 +25,7 @@ class DockerImage(Image):
         history: a list of commands used to create the filesystem layers
         to_dict: return a dict representation of the object
     '''
-    def __init__(self, repotag=None, id=None):
+    def __init__(self, repotag=None, id=None):  # pylint: disable=redefined-builtin
```
nishakm: How is this a redefined-builtin?

tern/classes/image.py
```diff
@@ -19,7 +19,7 @@ class Image(object):
         get_layer_diff_ids: returns a list of layer diff ids only
         to_dict: return a python dictionary representation of the image
     '''
-    def __init__(self, id=None):
+    def __init__(self, id=None):  # pylint: disable=redefined-builtin
```
nishakm: I am really interested in knowing why pylint picks this up as a redefined-builtin while it doesn't pick it up for the Command class that also has the same pattern: `def __init__(self, shell_command)`

tern/classes/package.py
```diff
@@ -43,7 +43,7 @@ def license(self):
         return self.__license
 
     @license.setter
-    def license(self, license):
+    def license(self, license):  # pylint: disable=redefined-builtin
```
nishakm: Again, why did pylint pick up this particular pattern while the other patterns above (like `def version(self, version)` didn't bring up anything?

tern/classes/image.py
mkrohan: "id" is the built-in here:  ```id(object) -> integer``` Return the identity of an object.

tern/classes/package.py
mkrohan: "license" is the built-in here: ```license()``` print the Python license.

PR #214, opened 2019-04-05
classes: Added abstract class Template

### LINKED ISSUE #184: Create abstract class Template
Something like this:
```
from abc import abstractmethod
from abc import ABCMeta

class Template(object):
    __metaclass__ == ABCMeta

    @abstractmethod
    def package():
        pass

   @abstractmethod
    def layer():
        pass
    
    @abstractmethod
    def image():
        pass

   # these don't have to be implemented but would be nice to
    def notice():
         pass

    def notice_origin():
         pass

    def origins():
         pass
  }
```
See #180 for layout. This is a sub-issue of #163

nishakm:
This is the base class for creating custom templates for report.
Any subclass must implement package, layer and image properties
and may implement notice, notice_origin and origins.

Resolves #184

### DIFF
--- a/tern/classes/template.py
+++ b/tern/classes/template.py
@@ -1,0 +1,48 @@
+#
+# Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+from abc import ABCMeta
+from abc import abstractmethod
+
+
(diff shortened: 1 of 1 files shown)


nishakm:
# WIP - Don't merge!

I'm still not sure if this way of creating templates is going to work

nishakm:
More commits need to follow. Closing this one.

Issue #220, opened 2019-04-09
Tern report command report "Cannot retrieve full image metadata"


rparikh:
**Describe the bug**
Following the tern guide on its homepage, 'tern report' command fails with the warning of Cannot retrieve full image metadata. This happens right after `fs_hash.sh` script is called. 

**To Reproduce**
1. Use the vagrant file provided by tern to create a VM in vagrant.
2. Use steps to setup all tern requirements on the VM and setup tern repository
3. Run command `tern -l report -f output.txt -i hello-world`
4. 2019-04-09 02:54:55,847 - WARNING - report - Cannot retrieve full image metadata

**Error in terminal**
(ternenv) vagrant@ubuntu-bionic:~/ternenv/tern$ tern -l report -f output.txt -i hello-world
2019-04-09 02:54:55,818 - DEBUG - __main__ - Starting...
2019-04-09 02:54:55,818 - DEBUG - report - Setting up...
2019-04-09 02:54:55,819 - DEBUG - container - Checking if image "hello-world" is available on disk...
2019-04-09 02:54:55,828 - DEBUG - container - Image "hello-world" found
2019-04-09 02:54:55,838 - DEBUG - rootfs - Running command: sudo /home/vagrant/ternenv/lib/python3.6/site-packages/tern/tools/fs_hash.sh /home/vagrant/ternenv/tern/temp/cdccdf50922d90e847e097347de49119be0f17c18b4a2d98da9919fa5884479d/contents
2019-04-09 02:54:55,847 - WARNING - report - Cannot retrieve full image metadata
2019-04-09 02:54:55,848 - DEBUG - report - Teardown...
2019-04-09 02:54:55,848 - DEBUG - rootfs - Running command: sudo rm -rf temp/mergedir
2019-04-09 02:54:55,855 - DEBUG - rootfs - Running command: sudo rm -rf temp/workdir
2019-04-09 02:54:55,864 - DEBUG - __main__ - Report completed.
2019-04-09 02:54:55,865 - DEBUG - __main__ - Finished

**Expected behavior**
Report for the container must be generated.

**Environment you are running Tern on**
Enter all that apply
- Operating System (Linux Distro and version or Mac or Windows)
- Vagrant file - https://github.com/vmware/tern/blob/master/vagrant/Vagrantfile
- Container OS - hello-world
- Python version (3.6 or higher)
- Cloud environment (AWS, Azure, GCP)

**Please attach files if they exist**
- tern.log
- report.*

PR #221, opened 2019-04-09
Adding 'attr' install in vagrant bootstrap.

rparikh:
fs_hash.sh uses getfattr to collect detailed file attributes. This
functionality is not available on native ubuntu so bootstrap has to
take care of installing it on the VM.

As 'attr' library was not available on the VM, the dependency to run
'tern report' was missing. That resulted in failure to generate report
with 'Cannot retrieve full image metadata'. fs_hash.sh collects detailed
file attributes using 'attr' library. As this dependency was not met,
the command used to return error.

Resolves: #220

Signed-off-by: Ravi Parikh <parikhr@vmware.com>

### DIFF
--- a/vagrant/bootstrap.sh
+++ b/vagrant/bootstrap.sh
@@ -7,8 +7,8 @@ sudo apt-get update
 # Upgrade all currently installed packages
 sudo apt-get -y upgrade
 
-# Python3 versions
-sudo apt-get install -y python3 python3-pip python3-venv
+# Python3 versions and system dependencies
+sudo apt-get install -y python3 python3-pip python3-venv attr
 
 # Install Docker
 sudo apt-get install -y docker.io


vagrant/bootstrap.sh
```diff
@@ -7,6 +7,9 @@ sudo apt-get update
 # Upgrade all currently installed packages
 sudo apt-get -y upgrade
 
+# Install attr package on the machine
+sudo apt-get install -y attr
+
 # Python3 versions
 sudo apt-get install -y python3 python3-pip python3-venv
```
nishakm: Wouldn't it be better to just add attr to this list? You can edit the comment to say `Install system dependencies`

rparikh: Sure, it makes sense. I will take care of this and update it.

nishakm:
Thank you! Could you squash your two commits into one?

nishakm:
Thank you! This looks good.

PR #223, opened 2019-04-10
Resolve code check issues for tern/__main__.py

mkrohan:
The issues raised by prospector for "tern/__main__.py" are:

    tern/utils/dockerfile.py
      Line: 8
        pylint: pointless-string-statement / String statement has no
                effect
      Line: 30
        pep8: W605 / invalid escape sequence '\$' (col 3)
        pep8: W605 / invalid escape sequence '\{' (col 5)
        pep8: W605 / invalid escape sequence '\}' (col 7)
      Line: 54
        pylint: simplifiable-if-statement / The if statement can be
                replaced with 'var = bool(test)' (col 8)
      Line: 152
        pylint: consider-iterating-dictionary / Consider iterating
                the dictionary directly instead of calling .keys()
                (col 15)

This change resolves these issues.

NOTE: The PEP8 issues wrt escape sequences were suppressed rather
than resolved.  It was clear what testing would be needed if this
string were replaced with a raw string: r'...'

### DIFF
--- a/tern/utils/dockerfile.py
+++ b/tern/utils/dockerfile.py
@@ -1,11 +1,12 @@
-'''
-Copyright (c) 2017-2018 VMware, Inc. All Rights Reserved.
-SPDX-License-Identifier: BSD-2-Clause
-'''
-import re
-'''
+#
+# Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
+"""
 Dockerfile parser and information retrieval
-'''
+"""
+
+import re
 
 directives = ['FROM',
               'ARG',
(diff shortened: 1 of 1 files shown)


tern/utils/dockerfile.py
```diff
@@ -27,7 +28,7 @@
 
 # regex strings
 cleaning = '[\t\\\\]'
-bash_var = '[\$\{\}]'
+bash_var = '[\$\{\}]'  # noqa
```
nishakm: Probably replacing both strings with a raw string will not cause an issue. Will file a technical-debt issue for this.

nishakm: Filed issue #224

nishakm:
Commit says `__main__.py` but file changed is `tern/utils/dockerfile.py`. Do you mind updating the commit?

mkrohan:
Commit message has been updated.

Issue #218, opened 2019-04-08
Remove unused argument in get_base_bin function in common.py


nishakm:
PR #216 dismisses an unused binary which should be removed.

PR #228, opened 2019-04-13
Removing an unused argument from get_base_bin.

rparikh:
Removing an unused argument from function get_base_bin
in common.py. Also removing relevant pylint comment.

Resolves: #218

Signed-off-by: Ravi Parikh <parikhr@vmware.com>

### DIFF
--- a/tern/common.py
+++ b/tern/common.py
@@ -65,7 +65,7 @@ def save_to_cache(image):
             cache.add_layer(layer)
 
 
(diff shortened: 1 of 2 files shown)


nishakm:
@rparikh Looks good! One small favor: can you choose a more specific commit heading? `Removing an unused argument from a function.` is too generic. How about: `Remove unused argument from get_base_bin`?

Issue #213, opened 2019-04-05
Rename `id` and `license` class properties


nishakm:
**Problem Statement**
`id` and `license` are built-in Python objects. PR #203 dismisses them but for clarity, they need to be renamed to something else. `image_id and `pkg_license` sounds like the most straightforward thing. 

**Describe the Proposal**
Rename properties `id` and `license` in all classes. Then rename the references throughout the code.

ManishaTripathy:
Hi , is it still an open issue? i would like to work on it

nishakm:
@ManishaTripathy sure, go for it! Here's what you'll need to do:
1. Find within `tern/classes` all classes that have a property called `id` (I think Image is the only one that has it). Rename it to something that makes sense
2. Replace all places in the test for that class in `tests`. Run the tests and make sure they pass.
3. Replace all places in the code where that property occurs. Make sure you test out your changes with some docker images
4. Repeat for classes with `license` as the property (I think package is the only one that has that for now)

Feel free to submit 2 PRs - one for the `id` replacement and one for the `license` replacement.

Don't hesitate to ask any questions. Thanks for working on this!

ManishaTripathy:
I see id in docker_image.py and image.py. Do we need to replace both or just image_py?
After replacing, we need to run only the changed tests or all tests in general?

nishakm:
@ManishaTripathy good catch! Please replace `id` in both `image.py` and `docker_image.py`. Hold off on the tests. Let me see if they pass now.

nishakm:
@ManishaTripathy currently only `test_class_image.py` passes. `test_class_docker_image.py` needs some work still so no need to run that one. The Travis build should cover that case.

nishakm:
Resolved in #231 and #233

PR #233, opened 2019-04-16
Renaming license to pkg_license in all classes and tests

ManishaTripathy:
This PR adds renaming of license to pkg_license in package class properties and corresponding tests. After renaming, I have verified that all tests in tests.test_class_package.py passed.

Resolves issue #213

Signed-off-by: Manisha Tripathy <tripathym@vmware.com>

### DIFF
--- a/tern/classes/package.py
+++ b/tern/classes/package.py
@@ -12,7 +12,7 @@ class Package:
     attributes:
         name: package name
         version: package version
-        license: package license
+        pkg_license: package license
         src_url: package source url
         origins: a list of NoticeOrigin objects
 
@@ -23,7 +23,7 @@ class Package:
     def __init__(self, name):
         self.__name = name
         self.__version = ''
-        self.__license = ''
+        self.__pkg_license = ''
         self.__src_url = ''
         self.__origins = Origins()
 
@@ -40,12 +40,12 @@ class Package:
         self.__version = version
 
     @property
-    def license(self):
-        return self.__license
+    def pkg_license(self):
+        return self.__pkg_license
 
-    @license.setter
-    def license(self, license):  # pylint: disable=redefined-builtin
-        self.__license = license
+    @pkg_license.setter
+    def pkg_license(self, pkg_license):  
+        self.__pkg_license =pkg_license
 
     @property
     def src_url(self):
@@ -86,7 +86,7 @@ class Package:
         '''The package dict looks like this:
             name: <name>
             version: <version>
-            license: <license string>
+            pkg_license: <packagelicense string>
             src_url: <source url>
         the way to use this method is to instantiate the class with the
         name and then give it a package dictionary to fill in the rest
@@ -95,7 +95,7 @@ class Package:
         success = True
         if self.name == package_dict['name']:
             self.version = package_dict['version']
-            self.license = package_dict['license']
+            self.pkg_license = package_dict['pkg_license']
             self.src_url = package_dict['src_url']
         else:
             success = False
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -28,7 +28,7 @@ class TestTemplate1(Template):
     def package(self):
         return {'name': 'package.name',
                 'version': 'package.version',
-                'license': 'package.license'}
+                'pkg_license': 'package.license'}
 
     def image_layer(self):
         return {'diff_id': 'layer.diff',
@@ -36,7 +36,7 @@ class TestTemplate1(Template):
                 'packages': 'layer.packages'}
 
     def image(self):
-        return {'id': 'image.id',
+        return {'image_id': 'image.id',
                 'layers': 'image.layers'}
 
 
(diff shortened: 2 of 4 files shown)


tern/classes/package.py
```diff
@@ -12,7 +12,7 @@ class Package:
     attributes:
         name: package name
         version: package version
-        license: package license
+        license : package license
```
nishakm: This change is not needed

nishakm: Or rather, if you want to follow the same pattern throughout the file, it should be `pkg_license: package license`

tests/test_fixtures.py
```diff
@@ -1,6 +1,6 @@
 #
 # Copyright (c) 2019 VMware, Inc. All Rights Reserved.
-# SPDX-License-Identifier: BSD-2-Clause
+# SPDX-pkg_license-Identifier: BSD-2-Clause
```
nishakm: Looks like a bulk replace :) . This change is not needed.

tests/test_fixtures.py
```diff
@@ -28,15 +28,15 @@ class TestTemplate1(Template):
     def package(self):
         return {'name': 'package.name',
                 'version': 'package.version',
-                'license': 'package.license'}
+                'pkg_license': 'package.pkg_license'}
```
nishakm: Actually, if you could, can you rewrite this as `'pkg_license': 'package.license'`?

nishakm: This is a recently added feature (I'm continuing to work on it), and it would be neat to demonstrate that the value can be whatever people want to call `pkg_license`.

tests/test_fixtures.py
```diff
@@ -28,15 +28,15 @@ class TestTemplate1(Template):
...
 
     def image_layer(self):
         return {'diff_id': 'layer.diff',
                 'tar_file': 'layer.tarfile',
                 'packages': 'layer.packages'}
 
     def image(self):
-        return {'id': 'image.id',
+        return {'image_id': 'image.image_id',
```
nishakm: Good catch! Currently, the tests are *not* catching this. I'll see what's going on.

tests/test_fixtures.py
```diff
@@ -45,7 +45,7 @@ class TestTemplate2(Template):
     def package(self):
         mapping = {'name': 'package.name',
                    'version': 'package.version',
-                   'license': 'package.license',
+                   'pkg_license': 'package.pkg_license',
```
nishakm: Same here - would be great if `package.pkg_license` was replaced with `package.license`

tests/test_fixtures.py
```diff
@@ -60,7 +60,7 @@ def image_layer(self):
         return mapping
 
     def image(self):
-        mapping = {'id': 'image.id',
+        mapping = {'image_id': 'image.image_id',
```
nishakm: Tests should be failing for this and they're not. I'll take a look.

tern/classes/package.py
```diff
@@ -40,12 +40,12 @@ def version(self, version):
...
-        return self.__license
+    def pkg_license(self):
+        return self.__pkg_license
 
-    @license.setter
-    def license(self, license):  # pylint: disable=redefined-builtin
-        self.__license = license
+    @pkg_license.setter
+    def pkg_license(self, pkg_license):  # pylint: disable=redefined-builtin
+        self.__pkg_license =pkg_license
```
nishakm: You'll need to remove the `# pylint: disable=redefined-builtin` string as well

nishakm:
Also, in your commit message body, if you can add something from the actual issue to indicate what the fix does, that would be great.

nishakm:
Haha! I didn't have a test for that particular feature in the Image class yet, that's why the tests were passing. So many thanks for catching it!

tern/classes/package.py
ManishaTripathy: Sure will do the change!

tests/test_fixtures.py
ManishaTripathy: yes my bad! will revert it back to License

tests/test_fixtures.py
ManishaTripathy: Sure, if we want to keep 'pkg_license': 'package.license' here, we would need to have the same in tests/test_class_package.py", line 64, in testToDictTemplate
    self.assertEqual(dict1['package.pkg_license'] -changed to  self.assertEqual(dict1['package.license'] 
I will make the changes

nishakm:
@ManishaTripathy can you combine your 3 commits into one? Also, can you add all of my suggested edits to your commit message?

A note: if you are having a tough time updating or refactoring your commits, I would highly suggest following the git workflow here: https://github.com/nishakm/puns. Feel free to practice on it before you update your PR. You can always ask me questions on specific git operations. Thanks!

tern/classes/package.py
ManishaTripathy: Done!

ManishaTripathy:
Thanks @nishakm for the git workflow link, i have updated the PR with the suggested changes, please review!

nishakm:
@ManishaTripathy Thank you!! One small nit: Your commit message has two signed-off-by's can you remove the first one?

ManishaTripathy:
@nishakm please review the updated PR

nishakm:
PR looks good all around! Much appreciate you sticking with this. I hope you can submit another!

Issue #207, opened 2019-04-05
Traceback running tern according to Getting Started documentation


ericwb:
**Describe the bug**
While following the steps provided in the Getting Started on Linux, tern throws a traceback. Most likely due to missing dependencies (docker?).

**To Reproduce**
1. Go to https://github.com/vmware/tern#getting-started-on-linux
2. Follow instructions exactly
3. Run `tern -l report -f output.txt -i debian:buster`

**Error in terminal**
```
(ternenv) browne@ubuntu18:~/ternenv/tern$ tern -l report -f output.txt -i debian:buster
2019-04-05 00:27:08,017 - DEBUG - __main__ - Starting...
2019-04-05 00:27:08,018 - DEBUG - report - Setting up...
2019-04-05 00:27:08,018 - DEBUG - container - Checking if image "debian:buster" is available on disk...
Traceback (most recent call last):
  File "/home/browne/ternenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/home/browne/ternenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 354, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.6/http/client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1234, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1026, in _send_output
    self.send(msg)
  File "/usr/lib/python3.6/http/client.py", line 964, in send
    self.connect()
  File "/home/browne/ternenv/lib/python3.6/site-packages/docker/transport/unixconn.py", line 42, in connect
    sock.connect(self.unix_socket)
FileNotFoundError: [Errno 2] No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/browne/ternenv/lib/python3.6/site-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/home/browne/ternenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 638, in urlopen
    _stacktrace=sys.exc_info()[2])
[...]

PR #237, opened 2019-04-25
RFC: improve docker socket error handling

SantiagoTorres:
From issue #207. We can probably better handle a misconfigured docker daemon by
blowing up *and* telling the user that it's not possible to contact the docker
daemon. This is more friendly than the actual behavior, which is 3 nested
exceptions taking place...

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -11,8 +11,14 @@ Tern executable
 import argparse
 import logging
 import os
+import sys
 
-from tern.report import report
+try:
+    from tern.report import report
+except Exception as e:
+    print("Couldn't set up the report module:\n\t{}".format(e))
+    sys.exit(1)
+
 from tern.utils import cache
 from tern.utils import constants
 
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -13,12 +13,6 @@ import logging
 import os
 import sys
 
-try:
-    from tern.report import report
-except Exception as e:
-    print("Couldn't set up the report module:\n\t{}".format(e))
-    sys.exit(1)
-
 from tern.utils import cache
 from tern.utils import constants
 
@@ -40,6 +34,13 @@ log_handler.setFormatter(formatter)
 console.setFormatter(formatter)
 
 logger.addHandler(log_handler)
+
+try:
+    from tern.report import report
+except Exception as e:
+    logger.addHandler(console)
+    logger.critical("Couldn't set up the report module:\n\t%s", e)
+    sys.exit(1)
 
 
 def do_main(args):
--- a/tern/report/report.py
+++ b/tern/report/report.py
@@ -13,11 +13,7 @@ import shutil
 import subprocess  # nosec
 import sys
 
-try:
-    from tern.utils import container
-except Exception:
-    raise
-
+from tern.utils import container
 from tern.report import content
 from tern.report import errors
 from tern.report import formats
--- a/tern/report/report.py
+++ b/tern/report/report.py
@@ -13,10 +13,14 @@ import shutil
 import subprocess  # nosec
 import sys
 
+try:
+    from tern.utils import container
+except Exception:
+    raise
+
 from tern.report import content
 from tern.report import errors
 from tern.report import formats
-from tern.utils import container
 from tern.utils import constants
 from tern.utils import cache
 from tern.utils import general
--- a/tern/utils/container.py
+++ b/tern/utils/container.py
@@ -32,12 +32,17 @@ logger = logging.getLogger(logger_name)
 client = None
 try:
     client = docker.from_env()
+    client.ping()
 except IOError:
     logger.critical("Docker daemon not running")
     raise Exception("Critical Error using Docker API. See logs for details")
 except OSError:  # pylint: disable=duplicate-except
     logger.critical("User has no access to docker unix socket")
     raise Exception("Critical Error using Docker API. See logs for details")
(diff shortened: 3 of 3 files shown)


nishakm:
Thanks so much! I'd like to check some code style on this before merging.

tern/utils/container.py
```diff
@@ -59,6 +64,8 @@ def check_container():
         return True
     except docker.errors.NotFound:
         return False
+    except (ConnectionRefusedError, ConnectionError):
```
nishakm: Maybe add a logger message here?

tern/utils/container.py
```diff
@@ -71,6 +78,9 @@ def check_image(image_tag_string):
         return True
     except docker.errors.ImageNotFound:
         return False
+    except (ConnectionRefusedError, ConnectionError):
```
nishakm: Add a logger message here?

tern/utils/container.py
```diff
@@ -83,6 +93,9 @@ def pull_image(image_tag_string):
     except docker.errors.ImageNotFound:
         logger.warning("No such image: \"%s\"", image_tag_string)
         return False
+    except (ConnectionRefusedError, ConnectionError):
```
nishakm: Same as above

tern/report/report.py
```diff
@@ -13,10 +13,14 @@
 import subprocess  # nosec
 import sys
 
+try:
+    from tern.utils import container
+except Exception:
+    raise
```
nishakm: Prospector doesn't like immediate raises. Would this also work?
```
except Exception as e:
    raise Exception(e)
```

tern/__main__.py
```diff
@@ -11,8 +11,14 @@
 import argparse
 import logging
 import os
+import sys
+
+try:
+    from tern.report import report
+except Exception as e:
+    print("Couldn't set up the report module:\n\t{}".format(e))
+    sys.exit(1)
```
nishakm: This is reasonable if running in a shell but not if you're running in a detached docker container you won't see this error in a situation where you're trying to connect to a registry or the host's unix socket. Maybe replace the `print` with the `logger`? That would mean moving the logger setup below before this import.

tern/utils/container.py
```diff
@@ -32,12 +32,17 @@
...
 except IOError:
     logger.critical("Docker daemon not running")
     raise Exception("Critical Error using Docker API. See logs for details")
 except OSError:  # pylint: disable=duplicate-except
     logger.critical("User has no access to docker unix socket")
     raise Exception("Critical Error using Docker API. See logs for details")
+except ConnectionError:
+    logger.critical("Can't connect to the Docker daemon!")
+    raise Exception("Critical Error using Docker API. See logs for details")
+
```
nishakm: Remove one blank line here

tern/utils/container.py
```diff
@@ -13,7 +13,7 @@
 import pwd
 import tarfile
 import time
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, ConnectionError
```
nishakm: ConnectionError is a built-in exception. No need to import it.

tern/utils/container.py
```diff
@@ -32,12 +32,17 @@
...
 except IOError:
     logger.critical("Docker daemon not running")
     raise Exception("Critical Error using Docker API. See logs for details")
 except OSError:  # pylint: disable=duplicate-except
     logger.critical("User has no access to docker unix socket")
     raise Exception("Critical Error using Docker API. See logs for details")
+except ConnectionError:
+    logger.critical("Can't connect to the Docker daemon!")
+    raise Exception("Critical Error using Docker API. See logs for details")
+
```
nishakm: ConnectionError should occur before an IOError. Can you move this except up to be the first one?

SantiagoTorres:
thanks @nishakm ! I added some changes to address your review. Hope this works :)

tern/__main__.py
```diff
@@ -35,6 +35,13 @@
 
 logger.addHandler(log_handler)
 
+try:
+    from tern.report import report
+except Exception as e:
+    logger.addHandler(console)
+    logger.critical("Couldn't set up the report module:\n\t%s", e)
+    sys.exit(1)
+
```
nishakm: On further review, I don't think this is a good way to solve this issue long term. Docker isn't supposed to be a hard dependency. It's only needed when you are specifically using the Docker API to manage images. I know the code looks like it's servicing only containers created by Docker but long-term, it needs to account for other image types. So I think rather than do the check here, we should do the check during execution.

nishakm: This requires a lot more messing around with `tern/utils/container.py` because the client is global. Rather than have that piece of code global, we can turn that into a function that returns the client object. The have the other functions accept an extra client object if there is any Docker interaction. That way, `execute_dockerfile` and `execute_docker_image` can do a check for the client and exit gracefully on error.

SantiagoTorres: Completely agree. I went down this path as I didn't want to refactor a lot of things without knowing what's the idea long term. I was thinking that, for now, this would achieve what we want on a less than 30-line diff.

We can close it and I can take a stab at refactoring this docker module to handle things more appropriately, or we could merge it and have a ticket about this that I can also gladly work on :+1: 
Thanks!

nishakm: I'd greatly appreciate it if you can work on a long term solution! I will close this PR.

Issue #251, opened 2019-05-03
Add a test for get_package_id method in Package class


nishakm:
**Description**
Add a test in `tests/test_class_package.py` called testGetPackageId

**Implementation**
Check to see if the method returns a string containing the package name and version.
For example, the test currently has an object `self.p1`. Test if `self.p1.get_package_id` returns "p1.1.0"

Make sure the tests in `test_class_package.py` pass.
```
pip install .
python tests/test_class_package.py
```
**References**
See other tests in `test_class_package.py` for reference

**Super Issues**
#239

nishakm:
@rnjudge Would you like to try this one?

rnjudge:
@nishakm sign me up!

nishakm:
@rnjudge awesome! It's yours.

Issue #241, opened 2019-04-25
Add get_package_id method to the Package class


nishakm:
**Description**
SPDX documents require a unique package identifier. Add a get_package_id method to the Package class that returns a string like this:
`name.version`

**Implementation**
Something like this, but with appropriate docstring
```
def get_package_id(self):
  return "{0}.{1}".format(self.name, self.version)
```
**Work Towards**
#239

PR #257, opened 2019-05-08
Add test for get_package_id in Package class

rnjudge:
Add a test to verify that the get_package_id
method in the Package class returns a proper
string containing the package name.version.

Resolves: #251
See also: #241

Signed-off-by: Rose Judge <rose.harber@gmail.com>


nishakm:
Thank you!

Issue #254, opened 2019-05-06
Rename 'src_url' to 'proj_url'


nishakm:
**Problem Statement**
The term `src_url` is misleading as the scripts do not retrieve the source url. Usually the method of recovering sources is varied and more complicated. There is already a listing tag called `src` which covers this scenario.

**Describe the Proposal**
Replace all instances of `src_url` with `proj_url` to reflect what gets retrieved.

**Steps to Implement Proposal**
Replace all instances of `src_url` with `proj_url`

PR #258, opened 2019-05-09
Rename 'src_url' to 'proj_url'

rnjudge:
The term src_url is misleading as the scripts do not
retrieve the source url. Replace all instances of
src_url with proj_url to reflect what gets retrieved.

Resolves: #254

Signed-off-by: Rose Judge <rose.harber@gmail.com>


nishakm:
@rnjudge This issue has been around since the beginning of the project! Many thanks for resolving it!

PR #270, opened 2019-05-15
WIP: spdx example doc for review

nishakm:


nishakm:
Sending this file through the validator produced this error:

```
Unable to parse the file: File /var/www/spdxtools/spdx-online-tools/src/app/media/AnonymousUser/1557942618/spdx_output_golang_1.12_alpine.txt is not a recognized RDF/XML or tag/value format. While verifying for Tag/Value format: Error converting tag/value to RDF/XML format: Invalid checksum: 820b5a21e229f23001048d25d04ecc8af9210e6983ff529513aad75ead5416ce at line number 26. While verifying for RDF/XML format: [line: 1, col: 1 ] Content is not allowed in prolog.
```
@swinslow any idea why it is rejecting the checksum?

swinslow:
Ah, yes -- for Checksum fields, the tag-value format requires a sub-tag that specifies which checksum hashing algorithm is being used. Take a look at the SPDX spec in section 3.10 (direct link: https://spdx.org/spdx-specification-21-web-version#h.wz2ycokhf58)

So this should probably be e.g.:
`PackageChecksum: SHA256: 820b5a21e229f23001048d25d04ecc8af9210e6983ff529513aad75ead5416ce`

(or SHA1 or MD5, which are the other checksum types currently defined in SPDX)

nishakm:
OK, solving that and the validator gives a even more vague error:
```
Unable to parse the file: File /var/www/spdxtools/spdx-online-tools/src/app/media/AnonymousUser/1557944533/tern_spdx_output.txt is not a recognized RDF/XML or tag/value format.
While verifying for Tag/Value format: Error converting tag/value to RDF/XML format: Invalid license expression. Expecting more operands.. 
While verifying for RDF/XML format: [line: 1, col: 1 ] Content is not allowed in prolog.
```
I suppose for the first error, I have hit on the license string formatting issue. For the second part, I'm not sure what this error means.

swinslow:
Hi @nishakm, yes, the first error looks to me like it's complaining that the license strings in some of the `PackageLicenseDeclared` fields don't match the expected format (e.g., SPDX License List IDs and/or license expressions with AND/OR, etc.)

For the second error (`Content is not allowed in prolog`), I have no idea. :)  

Adding @goneall -- do you happen to know what the last error message in the validator output above is referring to? Or is that just a statement that some error (e.g., the invalid license strings) was encountered so the validator is failing?

nishakm:
Note - updated with the corrected document (SHA256: subtag added). The validator still produces the above errors.

goneall:
The second error relates to interpreting the text as an RDF/XML document.  Since this is a tag/value format file you can ignore any reported errors where it starts with "While verifying for RDF/XML format:"

docs/spdx_output_golang_1.12_alpine.txt
```diff
@@ -0,0 +1,337 @@
...
+SPDXID: SPDXRef-libssl1.1.libssl1.1-1.1.1b-r1
+FilesAnalyzed: false
+PackageComment: <text>libssl1.1:
+	warning: No metadata for key: copyright
+	warning: No metadata for key: download_url
+</text>
+
+PackageName: ca-certificates-cacert
+PackageVersion: ca-certificates-cacert-20190108-r0
+PackageLicenseDeclared: MPL-2.0 GPL-2.0-or-later
```
goneall: There is a missing operand between these two licenses - an AND is required if this is a conjunctive license where the terms of both licenses apply or an OR is required if this is a disjunctive license where there is a choice of license terms

docs/spdx_output_golang_1.12_alpine.txt
```diff
@@ -0,0 +1,337 @@
...
+SPDXID: SPDXRef-scanelf.scanelf-1.2.3-r0
+FilesAnalyzed: false
+PackageComment: <text>scanelf:
+	warning: No metadata for key: copyright
+	warning: No metadata for key: download_url
+</text>
+
+PackageName: musl-utils
+PackageVersion: musl-utils-1.1.20-r4
+PackageLicenseDeclared: MIT BSD GPL2+
```
goneall: Similar comment as above - need an AND or OR

docs/spdx_output_golang_1.12_alpine.txt
```diff
@@ -0,0 +1,337 @@
...
+SPDXID: SPDXRef-libc-utils.libc-utils-0.7.1-r0
+FilesAnalyzed: false
+PackageComment: <text>libc-utils:
+	warning: No metadata for key: copyright
+	warning: No metadata for key: download_url
+</text>
+
+PackageName: ca-certificates
+PackageVersion: ca-certificates-20190108-r0
+PackageLicenseDeclared: MPL-2.0 GPL-2.0-or-later
```
goneall: Similar comment - need to add an AND or OR between the license IDs

nishakm:
@goneall Unfortunately, there isn't any way I can automagically figure if the license is conjunctive or not from the package manager. All I can do is report out what the `license` part of the metadata is (which may be incorrect). @swinslow says he has some ideas on how to make this work but at this point, I don't know enough to automate that part.

tsteenbe:
@goneall We can fix the license findings to be valid SPDX in OSS Review Toolkit (ORT) if needed - we currently already parse things like BSD-2 to BSD-2-Clause as we have built-in SPDX-parser in ORT. 

@nishakm  Could you fetch source code location for the detected package by Tern?
This will allow us in ORT to scan the sources and those scan results can then be used to create the applicable SPDX expression for a package.

tsteenbe:
Quick fix for tern to be SPDX compliant is to simply add "LicenseRef-" in front of each license and "AND" it e.g. "MIT BSD GPL2+" => "LicenseRef-MIT AND LicenseRef-BSD AND LicenseRef-GPL2+"

nishakm:
> @nishakm Could you fetch source code location for the detected package by Tern?
> This will allow us in ORT to scan the sources and those scan results can then be used to create the applicable SPDX expression for a package.

Via package managers? Not unless the package manager itself has that functionality. We can download the sources and feed it to ORT. Would that work?

docs/spdx_output_golang_1.12_alpine.txt
```diff
@@ -0,0 +1,272 @@
...
+PackageDownloadLocation: golang:1.12-alpine
+PackageLicenseDeclared: NOASSERTION
+PackageLicenseConcluded: NOASSERTION
+PackageCopyrightText: NOASSERTION
+FilesAnalyzed: false
+PackageComment: <text>Layer: 820b5a21e2:
+	info: Loading packages from cache for layer 820b5a21e2
+</text>
+
+Relationship: SPDXRef-f1b5933fe4 CONTAINS SPDXRef-musl.musl-1.1.20-r4
```
nishakm: I get errors like this:
`Relationship at line 40 invalid: Missing required license text for LicenseRef-MIT in musl`

That means there should be some reference to some text in the document itself?

goneall:
> That means there should be some reference to some text in the document itself?

Yes - there should be a licenses section at the end of the SPDX document.  Something like the following:

```
## License Information
LicenseID: LicenseRef-MIT
ExtractedText: <text>Uses the MIT License</text>
```

nishakm:
> ## License Information
> LicenseID: LicenseRef-MIT
> ExtractedText: <text>Uses the MIT License</text>

I get an error:
```
Error converting tag/value to RDF/XML format: Expecting a definition of a file, package, license information, or document property at LicenseId: LicenseRef-MIT line number 371
```
That part of the document looks like this:
```
LicenseId: LicenseRef-MIT
ExtractedText: <text>Original license: MIT</text>
```

goneall:
The LicenseIds with the text definitions need to all be grouped together and placed at the end of the file.

nishakm:
> The LicenseIds with the text definitions need to all be grouped together and placed at the end of the file.

I actually tried that first - placed all the LicenseIds at the end of the file. I got the same error.

nishakm:
OK! The problem was that the LicenseIDs were not unique. So the updated document here passes validation. I will merge the new code and close this issue. Thanks @tsteenbe and @goneall!

nishakm:
Merged in https://github.com/vmware/tern/pull/280 with the quickfix

nishakm:
Filed issue https://github.com/vmware/tern/issues/281 referring to creating a module for the string mapping. Closing this one now.

goneall:
@nishakm Thanks for the update.  Not a very friendly error message from the SPDX validator IMHO.  

I added an issue to improve the error message: https://github.com/spdx/tools/issues/190

If you get a chance, can you review the issue and make sure I captured it correctly.  Also, if you still have the SPDX document that exhibits the error, could you attach it to the issue?

PR #283, opened 2019-05-23
Modify sentence in README

jrgarcia:
This particular sentence confused me upon reading it. This is an
attempt to fix verb tense and improve the clarity of the warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ To get just a list of packages, you can use the `-s` option to get a summary rep
 ```
 $ tern -l report -s -i golang:1.12 -f output.txt
 ```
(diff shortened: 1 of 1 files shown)


nishakm:
@jrgarcia Thanks! I agree the sentence is too long and doesn't read well (although grammatically correct). Could you modify your commit so it is more clear?
`Modify sentence in README -> README: Break up unnecessarily long sentence`

Also, your commit message body refers to just you as finding the sentence confusing, but I would wager a lot of folks probably do. So if you could refer to other readers here, that would be awesome!
`This particular sentence confused me upon reading it. -> This sentence might be confusing to readers`.

I appreciate the PR and the updates.

jrgarcia:
@nishakm Done! Thanks for the suggestions.

nishakm:
Thanks! I will merge this.

Issue #210, opened 2019-04-05
Add missing SPDX copyright header to files


ericwb:
**Describe the bug**
A fair number of files in the tern code base are missing copyright headers.  Some such examples include:
- requirements.txt
- setup.py
- Dockerfile
- Vagrantfile
- bootstrap.sh

Might be others, didn't look at every file.

**To Reproduce**
1. Browse each file and check for copyright header

**Error in terminal**
n/a

**Expected behavior**
Every source code file to have a copyright header.

**Environment you are running Tern on**
n/a

PR #288, opened 2019-06-01
Cleanup and add missing SPDX copyright headers

rnjudge:
This commit made 3 primary changes:
  1) Added missing SPDX copyright headers to any files
     where they were missing.
  2) Cleaned up a few copyright dates that were incorrect.
  3) Cleaned up empty comments at the end of copyright
     headers.

Resolves: #210

### DIFF
--- a/tern/classes/__init__.py
+++ b/tern/classes/__init__.py
@@ -1,1 +1,4 @@
 # -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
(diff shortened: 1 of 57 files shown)


tern/classes/__init__.py
```diff
@@ -1 +1,4 @@
-# -*- coding: utf-8 -*-
```
nishakm: Trailing whitespace

nishakm: Ugh..nvm, this line got deleted

PR #327, opened 2019-06-17
ci: Make test_suite a dictionary of lists

### LINKED ISSUE #322: Make test_suite a dictionary of lists instead of lists and strings
**Description**
The test_suite dictionary in 'ci/test_files_touched.py' currently contains values of mixed types (both strings and lists of strings). As a result, we have to check for the instance type when we are adding the tests to the alltests string (line 117). An simpler way to do this would be to have a dictionary of only lists and remove the instance type check.

**To Do**
1. Change all single-string values in test_suite with single-element lists.
2. Remove the check for 'if isinstance' in line 117 and simply extend all elements to alltests instead of append.

**Background**
We got the script to run and then realized there was a cleaner way to do it. This issue is mostly a reminder for me to clean this up on Monday.

**Super Issues**
Enter the issue numbers this issue is work towards resolving

rnjudge:
The test_suite dictionary in 'ci/test_files_touched' used to contain
values of mixed types (both strings and lists of strings). As a result,
we were checking for the instance type before adding the value to the
alltests string.

This commit makes the following change:
  1) Changes all values in tests_suite to lists of strings, even if the
     list is only one element.
  2) Removes the isinstance check before extending the value to alltests

Resolves #322

### DIFF
--- a/ci/test_files_touched.py
+++ b/ci/test_files_touched.py
@@ -31,33 +31,35 @@ if not changes:
 
 test_suite = {
     # dev-requirements.txt
-    re.compile('dev-requirements.txt'): 'tern -l report -i photon:3.0',
+    re.compile('dev-requirements.txt'): ['tern -l report -i photon:3.0'],
     # requirements.txt
-    re.compile('requirements.txt'): 'tern -l report -i photon:3.0',
+    re.compile('requirements.txt'): ['tern -l report -i photon:3.0'],
     # Dockerfile
     re.compile('Dockerfile'): [
         'docker build -t ternd .',
         './docker_run.sh workdir ternd "report -i golang:alpine"'],
     # Files under tern directory
     re.compile('tern/__init__.py|tern/__main__.py'):
-    'tern -l report -i golang:alpine',
+    ['tern -l report -i golang:alpine'],
     # tern/classes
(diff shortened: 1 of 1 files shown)


ci/test_files_touched.py
```diff
@@ -78,41 +78,37 @@
...
+        ['python tests/test_class_template.py']
         }
 
 alltests = []
 for change in changes:
     for check in test_suite.keys():
         if check.match(change):
-            # add to list differently based on type
-            if isinstance(test_suite[check], list):
                 alltests.extend(test_suite[check])
```
nishakm: I think this line needs to be indented in. Run `pycodestyle ci/test_files_touched.py` to find PEP8 errors.

rnjudge: I found a few lines that were too long with pycodestyle and amended my commit to fix that but nothing came up for the indentation:

(ternenv) rjudge@ubuntu:~/ternenv/tern$ pycodestyle ci/test_files_touched.py 
(ternenv) rjudge@ubuntu:~/ternenv/tern$ 

But just looking at it, I noticed the alltests.extend line was indented too far, so fixed that too. Thanks, Nisha!

nishakm:
Looks good! Merging...

PR #354, opened 2019-06-28
ci/cd: Create build and release pipeline for Tern

### LINKED ISSUE #211: Use pbr for managing setuptools packaging
**Describe the Feature**
Currently I see that setup.py has code to parse requirements.txt and set various metadata through code. The pbr module does all of this for you and can make your setup.cfg/setup.py look much simpler and clean.

https://docs.openstack.org/pbr/latest/

**Use Cases**
Easier and cleaner setuptools package management.

**Implementation Changes**
Describe what changes you think would need to happen to the code to implement this feature
```
Add pbr as a dependency
Strip out a lot of code in setup.py

See https://docs.openstack.org/pbr/latest/user/using.html
```

### LINKED ISSUE #287: Create a build and release pipeline for tern
**Problem Statement**
Tern should be able to create packaged releases to PyPI

**Describe the Proposal**
Create a build and release pipeline using Circle CI

This issue will be a super issue for the steps to implement the build and release pipeline

rnjudge:
This commit makes change to package Tern and upload it to PyPI
after each new release version. Specifically, the following
changes are relevant:

1) Add a pypi_deploy job+workflow to config.yml that will only trigger
   on GitHub version tags. The pypi_deploy job installs package
   requirements and creates a ~/.pypirc file with PyPI credentials
   (via circleci private environment variables). The value of these env.
   variables is not accessible by forked repositories but care should be
   taken when merging PRs to check if any forked repositories are
   making changes to .circleci/config.yml in a bad-faith attempt to
   expose their values. pypi_deploy goes on to package tern using
   setup.py and then pushes the package to PyPI using twine.

2) Add a VerifyVersion custom command to setup.py. The VerifyVersion
   command 'verify' is called in .circleci/config.yml to make sure that
   the git tag ($CIRCLE_TAG) matches the version of the project
   according to tern/__init__.py before packaging Tern and pushing to
   PyPI. If not, return a helpful error message that highlights the
   discrepancy between the git tag and Version declared in
   tern/__init__.py and do not push to PyPI.

3) Add a setup.cfg file to utilize the pbr module. The pbr module will
   allow for a cleaner setup.py file and does a lot of the work that
   setup.py was previously doing for us. This commit moves most of the
   code previously in setup.py to an INI-like setup.cfg file and removes
   the custom parsing functions (_read_long_desc() and
   _get_requirements()). Using pbr is preferred for easier and cleaner
   setuptools package management.

Resolves #287
Resolves #211

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,14 @@ commands:
       - checkout
       - run: pyenv global 3.6.5
       - run: pip install --upgrade pip
+  create_pypirc:
+    steps:
+      - run:
+          name: Create pypirc file
+          command: |
+           echo -e "[pypi]" >> ~/.pypirc
+           echo -e "username = $R_PYPI_UN" >> ~/.pypirc
+           echo -e "password = $PYPI_PW" >> ~/.pypiruc
 
 jobs:
   # linting using Prospector
@@ -58,8 +66,20 @@ jobs:
       - setup
       - run: pip install .
       - run: tern -l report -i photon:3.0
+  # Deploy to PyPi
+  pypi_deploy:
+   executor: ubuntu1604
+   steps:
+      - setup
+      - run: pip install .
+      - run: pip install twine
+      # make sure git tag matches release version
+      - run: python setup.py verify
+      - create_pypirc 
+      - run: python setup.py sdist
+      - run: twine upload dist/* 
+
 workflows:
-  # run a full functional test for photonOS
   version: 2
   PRs:
     jobs:
@@ -67,3 +87,13 @@ workflows:
       - commit_check
       - security
       - test_changes
+  Release:
+    jobs:
+      - pypi_deploy:
+          filters:
+            # ignore all commits on any branch by default
+            branches:
+              ignore: /.*/
+            # only run on version tags
+            tags:
+              only: /^v[0-9]+(\.[0-9]+)*$/
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,46 @@
+#!/usr/bin/env python                                                           
+# -*- coding: utf-8 -*-                                                         
+#                                                                               
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.                          
+# SPDX-License-Identifier: BSD-2-Clause
+
+[metadata]
+name = tern
+author = VMware Inc
+author-email = nishak@vmware.com
+summary = An inspection tool to find the OSS compliance metadata of the packages installed in a container image.
+description-file = README.md
+description-content-type = text/markdown; charset=UTF-8
+home-page = https://github.com/vmware/tern/
+project_urls =
+    Documentation = https://github.com/vmware/tern/tree/master/docs
+    Source Code = https://github.com/vmware/tern 
+    Issues = https://github.com/vmware/tern/issues
+license = BSD-2.0
+keywords = 
+    Distribution
+    Container
+    Cloud-Native
+classifier = 
+    Development Status :: 3 - Alpha
+    Environment :: Console
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
+    Natural Language :: English
+    Operating System :: POSIX
+    Operating System :: POSIX :: Linux
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: Implementation :: CPython
+    Topic :: Software Development
+
+[files]
+packages = 
+    tern
+
+[options]
+include_package_data = True
+
+[entry_points]
+console_scripts = 
+    tern = tern.__main__:main
--- a/setup.py
+++ b/setup.py
@@ -4,52 +4,31 @@
 # Copyright (c) 2019 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-from setuptools import setup, find_packages
+import os
+import sys
 
 from tern import Version
+from setuptools.command.install import install
+from setuptools import setup
 
 
-def _read_long_desc():
-    with open("README.md") as fp:
-        return fp.read()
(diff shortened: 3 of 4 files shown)


.circleci/config.yml
```diff
@@ -58,12 +66,32 @@ jobs:
...
 workflows:
-  # run a full functional test for photonOS
   version: 2
   PRs:
     jobs:
       - linting
       - commit_check
       - security
       - test_changes
+      - pypi_deploy:
```
nishakm: I feel like this should come under a different workflow called 'release'. I think there's something in CircleCI called 'trigger' that would trigger this workflow only if a tag appears.

setup.cfg
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python                                                           
+# -*- coding: utf-8 -*-                                                         
+#                                                                               
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.                          
+# SPDX-License-Identifier: BSD-2-Clause
+
+[metadata]
+name = tern
+author = VMWare Inc
+author-email = FIXME@FIXWHAT.THEEMAIL
```
nishakm: We'll need to change this to a legit email

setup.py
```diff
@@ -4,52 +4,31 @@
...
+from setuptools.command.install import install
+from setuptools import setup
 
 
-def _read_long_desc():
-    with open("README.md") as fp:
-        return fp.read()
+class VerifyVersion(install):
+    """Run a custom verify command"""
+    description = "Verify that the git tag matches current release version."
```
nishakm: This is nice!

setup.cfg
rnjudge: Done! Changed to Nisha's email.

.circleci/config.yml
rnjudge: I added a new workflow called `Release` and put the `pypi_deploy` under it. I think we *could* use trigger but don't necessarily need to. The 'filters' I have around `pypi_deploy` will only have it run on version tags.

.circleci/config.yml
```diff
@@ -58,12 +66,34 @@ jobs:
...
+  Release:
+    jobs:
+      - pypi_deploy:
+          filters:
+            # ignore all commits on any branch by default
+            branches:
+              ignore: /.*/
+            # only run on version tags
+            tags:
+              only: /^v[0-9]+(\.[0-9]+)*$/
```
nishakm: Did we decide to stick in a post-deploy job for testing if the pip package worked and have it be dependent on the pypi_deploy job?

setup.cfg
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python                                                           
+# -*- coding: utf-8 -*-                                                         
+#                                                                               
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.                          
+# SPDX-License-Identifier: BSD-2-Clause
+
+[metadata]
+name = tern
+author = VMWare Inc
```
nishakm: VMware is very particular about the small "w" after the big "M". This should be "VMware, Inc."
Sorry I didn't catch that before.

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+author-email = nishak@vmware.com
+summary = A compliance tool for containers
```
nishakm: "A container inspection tool for OSS license compliance"
Is there a character limit for this?

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+    Cloud-Native
+classifier = 
+    Development Status :: 3 - Alpha
+    Environment :: Console
+    Intended Audience :: Developers
+    License :: OSI Approved :: BSD License
+    Natural Language :: English
+    Operating System :: POSIX
+    Operating System :: POSIX :: Linux
+    Operating System :: MacOS :: MacOS X
```
nishakm: We might want to get rid of this one since tern can't run natively on Mac OSX

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+    Programming Language :: Python :: 3.6
```
nishakm: I'd like to add `3.7` as well.

setup.py
```diff
@@ -4,52 +4,31 @@
...
 
-
-def _get_requirements():
-
-    with open("requirements.txt") as fp:
-        return [requirement for requirement in fp]
+    def run(self):
+        tag = os.getenv('CIRCLE_TAG')
+        if tag != Version:
```
nishakm: Actually, the tag is of the form `v<major>.<minor>.<patch>` while `Version` is `<major>.<minor>.<patch>` so we would have to strip out the `v` first before doing a comparison.

setup.cfg
nishakm: It was originally:
```
description=("An inspection tool to find the OSS compliance metadata of"
                      " the packages installed in a container image."),
```

.circleci/config.yml
rnjudge: Yes we did. I thought this functionality would be added after the 0.5.0 release, however, since currently there is nothing to verify on PyPI. What about opening a separate issue for this test script and merging it as a separate PR?

nishakm: That works for me.

setup.cfg
rnjudge: I couldn't figure out how to wrap it but also didn't give it much time. Let me look more in to this.

rnjudge: Since there is no linting on this file, we can put the entire description on one line.

setup.cfg
rnjudge: Removed.

setup.py
rnjudge: stripped the v from tag so it will properly compare to 'Version.' Nice catch!

setup.cfg
rnjudge: Added.

PR #355, opened 2019-07-02
Add a version argument to tern CLI

### LINKED ISSUE #209: Add a version argument to tern CLI
**Describe the Feature**
Currently there is no way to identify the version of tern installed.  Typically a CLI will have a `--version` argument to provide such information.

As a added bonus, it's good to put environmental information in the version output that way you can tell people to simply run a command and copy the output into a bug being reported.

**Use Cases**
This would be very helpful for those opening bugs and need to provide version and environment information.

**Implementation Changes**
Describe what changes you think would need to happen to the code to implement this feature
```
pseudocode or python code is fine here
```

rnjudge:
This commit adds a --version option to Tern's CLI. This option will
display the version of Tern that is installed and also output the
python version. This commit also updates the GitHub issue template.
This will be useful for individuals opening bugs to provide the
version of Tern that they are opening the issue against.

Resolves #209

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -13,10 +13,13 @@ import argparse
 import importlib
 import logging
 import os
+import sys
 
 from tern.report import report
 from tern.utils import cache
 from tern.utils import constants
+from tern import Version
+
 
 # global logger
 logger = logging.getLogger(constants.logger_name)
@@ -77,6 +80,7 @@ def do_main(args):
 
 def main():
     parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
         prog='Tern',
         description='''
            Tern is a container image component curation tool. Tern retrieves
@@ -95,6 +99,11 @@ def main():
                         "Needed when running from within a container")
     parser.add_argument('-r', '--redo', action='store_true',
                         help="Repopulate the cache for found layers")
+    # sys.version gives more information than we care to print
+    py_ver = sys.version.replace('\n', '').split('[')[0]
+    parser.add_argument('-V', '--version', action='version',
+                        version="%(prog)s {0}\n   python version = {1}".format(
+                            Version, py_ver))
     subparsers = parser.add_subparsers(help='Subcommands')
     # subparser for report
     parser_report = subparsers.add_parser('report',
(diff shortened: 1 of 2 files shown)


ericwb:
It's also often useful to put environment information in the version output.  That way when a bug is report you can simply ask the originator to copy the output from the --version command.

For example, see some of the comments on a similar PR for bandit:
https://github.com/PyCQA/bandit/pull/284

rnjudge:
> It's also often useful to put environment information in the version output. That way when a bug is reported you can simply ask the originator to copy the output from the --version command.
> 
> For example, see some of the comments on a similar PR for bandit:
> [PyCQA/bandit#284](https://github.com/PyCQA/bandit/pull/284)

Thanks for the pointer, Eric! I updated my PR to print the python version for now. I'll also work on adding OS information and amend this PR tomorrow or Monday.

tern/__main__.py
```diff
@@ -95,6 +99,11 @@ def main():
                         "Needed when running from within a container")
     parser.add_argument('-r', '--redo', action='store_true',
                         help="Repopulate the cache for found layers")
+    # sys.version gives more information than we care to print
+    py_ver = sys.version.replace('\n', '').split('[')[0]
+    parser.add_argument('-V', '--version', action='version',
+                        version="%(prog)s {0}\n   python version = {1}".format(
```
nishakm: I'm not familiar with "%(prog)s" string formatter. What does it do?

tern/__main__.py
```diff
@@ -77,6 +80,7 @@ def do_main(args):
 
 def main():
     parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawTextHelpFormatter,
```
nishakm: So far I've seen that the help text wraps at a reasonable length. I'm not sure what adding this does.

tern/__main__.py
ericwb: https://docs.python.org/3/library/argparse.html#prog

tern/__main__.py
rnjudge: That was not my experience when I was testing these changes but perhaps it was just my environment? Do you want to give it a try with and without this and see if you have the same experience?

nishakm: Yup I'll test it out. It's probably the right call if the version text doesn't wrap like the rest of them.

nishakm: Sorry, I misinterpreted. You want the new line in the formatting. Looks good!

PR #365, opened 2019-08-06
Pin requirements only upto the minor version

### LINKED ISSUE #236: Should we only pin up to a minor version?
**Problem Statement**
The way we are defining requirements in `requirements.txt` and `setup.py` pin up to the patch version on dependencies. This seems a little bit too strict for downstream packagers. 

**Describe the Proposal**
We should pin the dependencies up to the minor version only using the syntax defined on [PEP-0440](https://www.python.org/dev/peps/pep-0440/#compatible-release)

I'm assuming we can basically `sed -i 's/==/~=/g' requirements.txt'`, but we may want to test things first :)

nishakm:
For development, in order to be less restrictive on the direct
dependency versions, we only require major and minor versions.

PyYAML - we will allow the latest version if available.
docker - we allow compatible versions.
requests - this is a docker dependency and only used in the code
for catching exceptions. We will allow compatible versions so
it doesn't break the docker API.
prospector - we will allow the latest version if available.
bandit - we will allow compatible versions.
GitPython - this is a bandit dependency. We will allow compatible
versions so it doesn't conflict with bandit.

Resolves #236

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,9 +5,9 @@
 -r requirements.txt
 
 # For development work, pull in dev support utilities
-prospector==1.1.6.2
+prospector>=1.1
 # For security linting, use Bandit
-bandit==1.6.0
+bandit~=1.6
 
 # For commit message linting in circleci we need GitPython
-GitPython==2.1.11
+GitPython~=2.1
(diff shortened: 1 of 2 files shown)


nishakm:
Also tested with calico:bird image from Dockerhub. Works as expected.

PR #368, opened 2019-08-06
Report base OS for image with /etc/os-release

### LINKED ISSUE #366: "warning: Unable to determine the base OS" for an image with an os-release file
**Describe the bug**
I used tern on a distroless image. I got a warning:
```
warning: Unable to determine the base OS of the image gcr.io/distroless/base:latest
```
This image has an os-release file: /etc/os-release

**To Reproduce**
1. Run: `tern -kl report -i gcr.io/distroless/base:latest`
2. See warning:
```
This report was generated by the Tern Project
https://github.com/vmware/tern/commit/e141c00d8dc314b58bf37cd135f576fa7b75a227

Docker image: gcr.io/distroless/base:latest:
        Layer: 18310facd7:
                warning: Unable to determine the base OS of the image gcr.io/distroless/base:latest
```
3. Go to `temp/mergedir/etc` and see os-release file.

**Expected behavior**
Tern should report what is in the os-release file and not give this warning.

- Output of 'tern --version'
```
Tern 0.4.0
   python version = 3.7.3 (default, Apr  3 2019, 19:16:38)
```
- Operating System
Ubuntu 18.04 LTS VM

rnjudge:
Currently if an image layer does not have a binary package manager Tern
does not report on the underlying base OS because it assumes that no
package manager means no distro. Generally, this is true but we recently
came across a use case for 'distroless' where there is no package
manager but an /etc/os-release file does exist. This commit will make
two primary changes:

1) Even when there is no binary, check for the existence of an
   /etc/os-release file. If the file is present, report on the findings
   inside.

2) Instead of reporting "Unable to determine the base OS of the image"
   when there is no binary package manager, change the warning message
   to be more accurate -- "Unable to find a known package manager.
   Cannot list packages."

Resolves #366

### DIFF
--- a/tern/helpers/common.py
+++ b/tern/helpers/common.py
@@ -424,6 +424,10 @@ def get_os_style(image_layer, binary):
         # We know with high degree of certainty what the OS is
         image_layer.origins.add_notice_to_origins(origin_layer, Notice(
             formats.os_release.format(os_style=get_os_release()), 'info'))
+    elif binary is None:
+        # No binary and no os-release means we have no idea about base OS
+        image_layer.origins.add_notice_to_origins(origin_layer, Notice(
+            errors.no_etc_release, 'warning'))
     else:
         # We make a guess about the OS based on pkg_format + binary
         # First check that binary exists in base.yml
@@ -433,7 +437,7 @@ def get_os_style(image_layer, binary):
                     errors.no_listing_for_base_key.format(listing_key=binary),
                     'warning'))
         else:
-            # Assign image layer attributes
+            # Assign image layer attributes and guess OS
             image_layer.pkg_format = pkg_format
             image_layer.os_guess = os_guess
             image_layer.origins.add_notice_to_origins(origin_layer, Notice(
--- a/tern/report/errors.py
+++ b/tern/report/errors.py
@@ -12,15 +12,18 @@ unrecognized_base = '''Unable to determine the base OS of the image ''' \
 no_packages = '''Unable to recover packages for layer {layer_id}. ''' \
     '''Consider either entering them manually or create a bash script to ''' \
     '''retrieve the package in the command library.\n'''
+no_package_manager = '''Unable to find a known package manager. Cannot ''' \
+    '''list packages.\n'''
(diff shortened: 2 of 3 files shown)


rnjudge:
Should mention, this is the new output when I run Tern with a distroless container with these changes:
```
Docker image: gcr.io/distroless/static:latest:
        Layer: 2d0ec87df3:
                info: Found 'Distroless' in /etc/os-release.
                warning: Unable to find a known package manager. Cannot list packages.
```


This is the output with a container with no /etc/os-release and no binary:
```
Docker image: hello-world:
        Layer: 684acdfb82:
                warning: Unable to find a known package manager. Cannot list packages.
```

And here is the output for an image like debian where /etc/os-release is available:
```
Docker image: debian:buster:
        Layer: 5e8a7aa54d:
                info: Found 'Debian GNU/Linux 10 (buster)' in /etc/os-release
```

nishakm:
> This is the output with a container with no /etc/os-release and no binary:
> 
> ```
> Docker image: hello-world:
>         Layer: 684acdfb82:
>                 warning: Unable to find a known package manager. Cannot list packages.
> ```

I think in this situation, we ought to inform the user that this is an unknown base OS.
I think it's OK to have a warning message saying: "Unable to find an os-release file"

I am yet to find a situation where there is no os-release file but there is a package manager. I think that's fine because you're checking it in the `guess_os_style` function.

rnjudge:
> > This is the output with a container with no /etc/os-release and no binary:
> > ```
> > Docker image: hello-world:
> >         Layer: 684acdfb82:
> >                 warning: Unable to find a known package manager. Cannot list packages.
> > ```
> 
> I think in this situation, we ought to inform the user that this is an unknown base OS.
> I think it's OK to have a warning message saying: "Unable to find an os-release file"
> 
> I am yet to find a situation where there is no os-release file but there is a package manager. I think that's fine because you're checking it in the `guess_os_style` function.

I think we should report both in the case of no /etc/release and no binary.
1) Inform that it is an unknown base OS
2) Inform that there is no package manager and we cannot list packages.

nishakm:
Looks good to me. Merging.

PR #369, opened 2019-08-06
Add git commit sha1 from git repo to -V option

### LINKED ISSUE #367: tern --version should report git commit if working from a git repo
**Describe the Feature**
Currently `tern --version` reports whatever the variable `Version` is in the code. If working from a git repo, have it report the git commit sha1.

**Use Cases**
If a developer is working on top of master or a development branch and they come across an issue, it would be helpful that they report the git commit rather than the version as the development branch would have proceeded beyond the release version. In this case the git commit would be more relevant information.

**Implementation Changes**
Use function `get_git_rev_or_version()` in `tern/utils/general` to get either the version or the commit. I don't think `ver_type` is needed unless you want to specify what is being used.

rnjudge:
Currently, `tern --version` reports whatever the variable `Version` is in
the code. If a user is working from a git repo, it should report the git
commit sha1 rather than the version as the development branch would have
proceeded beyond the release version and the git commit would be more
relevant.

This commit utilizes an exisitng function in `tern/util/general`,
`get_git_rev_or_version`, and outputs either the static version of Tern or
the latest git commit sha1 if available. Since multiple lines of code
are needed to determine the string that should be outputted to the user
when the --version command line option is called, a new function was
created in `tern/__main__.py` called `get_version()`. This will keep the
parser.add_argument code cleaner and more organized.

Resolves #367

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -18,7 +18,7 @@ import sys
 from tern.report import report
 from tern.utils import cache
 from tern.utils import constants
-from tern import Version
+from tern.utils import general
 
 
 # global logger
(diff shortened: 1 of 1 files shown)


nishakm:
Looks good to me. Merging.

PR #371, opened 2019-08-08
Build a binary wheel of the Tern module for PyPI

rnjudge:
Instead of packaging Tern for PyPI with only sdist, package with
bdist_wheel. This is preferrable as it will also package the
dependencies. In order to do this, add a line to pip install wheel as
part of the build and release pipeline in .circleci/config.yml.

This commit also adds wheel as a requirement in dev-requirements.txt.
PyYAML automatically builds from source and needs wheel. Adding this
requirement will eliminate developers needing to install wheel
separately from Tern.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,11 +73,11 @@ jobs:
       - setup
       - run: pip install .
       - run: pip install twine
+      - run: pip install wheel
       # make sure git tag matches release version
       - run: python setup.py verify
       - create_pypirc 
-      - run: python setup.py sdist
-      - run: twine upload dist/* 
+      - run: python setup.py sdist bdist_wheel upload
 
 workflows:
   version: 2


dev-requirements.txt
```diff
@@ -11,3 +11,6 @@ bandit~=1.6
 
 # For commit message linting in circleci we need GitPython
 GitPython~=2.1
+
+# PyYAML builds from source and needs bdist_wheel
```
nishakm: I tested this out and it doesn't work. Unfortunately, pip will only install packages listed in alphabetical order. So this change will not work. I will submit a PR with a doc change in the development setup.

nishakm:
I'm not sure if we should add `wheel` to `setup_requires` in the `setup.py` because it seems as if `pbr` requires `wheel` anyway. It's probably something we will find out when we test the release :)

nishakm:
Looks good to me. Merging.

PR #372, opened 2019-08-09
docs: Require wheel to be installed

nishakm:
Some dependencies for tern do not build python wheels, resulting in
a bunch of warnings about bdist_wheel not being present and pip
resorting to installing from the binary packages.
It's better to have these installed from source. In order to do that
a developer needs to install wheel first.

We cannot add these in the requirements because pip checks and installs
packages in alphabetical order. So we will require the developer to
install wheel first before they install from the dev-requirements file.


rnjudge:
This looks good to me. Merging now.

PR #376, opened 2019-08-13
Add list of direct and transitive dependencies

### LINKED ISSUE #374: Add a yaml file dependency document that lists transitive dependencies
**Problem Statement**
When we cut a release, it would be helpful to see a list of all the dependencies, ordered by which package pulls them in. 

**Describe the Proposal**
List these dependencies in yaml format so we can easily parse the list later down the road.

pip freeze will give you the list of deps but not which are transitive and which are the primary requirement. Use a package manager like npm to help sort which are transitive and which package pulls them in.

nishakm:
When using pip to list packages that were installed during
development, it is difficult to figure out which dependencies
are direct and which dependencies are transitive. This is
especially difficult for transitive dependencies which we are using
(to catch specific exceptions for example).

We add a yaml file which lists the direct and transitive dependencies
for Tern. The file is in yaml so we can automate updating dependencies
later on.

Resolves #374

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -1,0 +1,18 @@
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# This file describes the direct and transitive dependencies for tern
+# It is mostly meant for informational purposes but can be used for
+# automating housekeeping tasks.
+
+tern:
+    - PyYAML
+    - docker:
+      - certifi
+      - chardet
+      - idna
+      - requests
+      - six
+      - urllib3
+      - websocket-client
+


rnjudge:
Do we want versions on this list?

nishakm:
> Do we want versions on this list?

I don't think so? Versions for all direct dependencies are listed in requirements.txt. Is there any use case where we would need versions here?

rnjudge:
> > Do we want versions on this list?
> 
> I don't think so? Versions for all direct dependencies are listed in requirements.txt. Is there any use case where we would need versions here?

When we cut a release and perform the step to "Do pip freeze and check the major and minor versions for all the dependencies in the requirements.txt file." I was thinking we might automate the comparison later down the line. But maybe we are not going to use this list for any comparison of package versions?

nishakm:
I'd like to close this PR in favor of using pip-tools: https://github.com/jazzband/pip-tools
It will allow us to create a release freeze file and records hashes and the transitive dependencies. It also allows us to keep the squishy pins in our current requirements.txt file and a workflow to check for updates.

nishakm:
Closing in favor of #387

PR #381, opened 2019-08-15
WIP: Feature: Add stevedore to Tern

### LINKED ISSUE #370: Incorporate plugin architecture to Tern
**Problem Statement**
Currently, Tern and all of its reporting are under the same core application code. It would be nice to separate some of the reporting functionality in to separate plugins. Enabling Tern for plugins will also help achieve better extensibility for the project as there is potential for external tools to plug in to Tern's ability to gather a container image's internals.

**Describe the Proposal**
The end-goal of this issue is to more easily add plugin report styles to Tern. This will require re-factoring to the code to separate the core of Tern and the report generation. 

The OpenStack Stevedore python modules seems to be a good place to start. Stevedore allows for dynamic loading of plugins through setuptools entry points.

Start by separating spdxtagvalue report into it's own plugin and then work to separate other styles of reporting from there.

rnjudge:
Commit 6907918 changes the following:

1) Modify setup.cfg to add an entrypoint for the spdxtagvalue plugin.

2) Introduce the Stevedore python module to report.py. Create a
   DriverManager that dynamically loads the plugin called by the user
   when they invoke Tern from the command line. The DriverManager will
   look for a 'generate' function in the plugin derived class to create
   the custom report. If no plugin with the name invoked by the user is
   found, no report is generated. This is a change from the way Tern
   currently runs.

3) Create a new directory tern/formats. This is where all plugins will
   live. Within that folder, create a new file, generator.py, that will
   serve as the base class for all plugins to inherit from.
   tern/formats/generator.py contains one abstract method, generate(),
   that must be implemented by all future plugins to generate the
   output document for the user.

4) Removes the following files, as they will be added in a new location
   in the following commit:
   - tern/classes/templates/__init__.py
   - tern/classes/templates/spdx.py
   - tern/report/spdxtagvalue/__init__.py
   - tern/report/spdxtagvalue/formats.py
   - tern/report/spdxtagvalue/generator.py

-------------------------------------------------------------------------------------------------

Commit 7e1f6c0 makes the following changes:

1) Move tern/report/spdxtagvalue/formats.py ->
   tern/formats/spdx/formats.py. tern/formats is where all plugins
   will live. Change any references to formats.py to import from
   the correct location.

2) Move tern/report/spdxtagvalue/generator.py ->
   tern/formats/spdx/spdxtagvalue/generator.py. Also create a
   derived class called Spdxtagvalue to inherit from
   tern/formats/generator.py's GenerateReport class. Put the
   generate() function that was available previously in this file
   inside the Spdxtagvalue derived class. This was merely a matter
   of copying/pasting the code and adding the correct indentation.

3) Copy tern/classes/templates/spdx.py ->
   tern/formats/spdx/spdx.py.

4) Remove check_format_type() from tern/__main__.py. Now that we are
   using Stevedore to manage the plugins, this check is no longer
   needed as it is all handled by Stevedore.

Resolves #370


tern/formats/generator.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import abc
+
+
+class GenerateReport(object):
```
nishakm: Either name this Generator because the file is called `generator.py` or call the file generate_report.py. It's for better readability.

tern/formats/generator.py
```diff
@@ -0,0 +1,16 @@
...
+    '''Base class for report plugins'''
+    def __init__(self):
+        '''Do nothing'''
```
nishakm: Line: 16
    pylint: pointless-string-statement / String statement has no effect (col 8)

Do you need an `__init__`?

tern/formats/generator.py
```diff
@@ -0,0 +1,16 @@
...
+    @abc.abstractmethod
+    def generate(self, image_obj_list):
+        '''Format the report according to the plugin style'''
+        '''Each subclass is responsible for their own formatting.'''
```
nishakm: Prospector didn't complian about the "pointless string statement" but you can run these statements together and then put `pass` at the end. See `tern/classes/template.py` as an example.

tern/formats/spdx/spdxtagvalue/generator.py
```diff
@@ -159,133 +160,135 @@ def get_license_block(license_list):
...
-        PackageCopyrightText='NOASSERTION',
-        FilesAnalyzed='false') + '\n'
-    # Add image relationships
-    report = report + get_image_relationships(image_obj) + '\n'
-
-    # Add the layer part for each layer
-    for index, layer_obj in enumerate(image_obj.layers):
-        # this is the main block for the layer
+
+class Spdxtagvalue(generator.GenerateReport):
```
nishakm: Could you rename this class to `SpdxTagValue` following CamelCase?

tern/report/report.py
```diff
@@ -341,14 +343,21 @@ def generate_format(images, format_string):
...
+        )
+        return mgr.driver.generate(images)
+    except NoMatches:
+        pass
 
 def report_out(args, *images):
     report = generate_report(args, *images)
-    if args.file:
+    if not report:
+        logger.error("{} not a recognized plugin.".format(args.report_format))
```
nishakm: tern/report/report.py
  Line: 359
    pylint: logging-format-interpolation / Use % formatting in logging functions and pass the % parameters as arguments (col 21)

nishakm:
Regarding the change in the CLI:
1. When running `tern report -i <image>` you should get the default human-readable report
2. When running `tern report -m spdxtagvalue -i <image>` you should get the spdx tag-value format
3. When running `tern report -m random -i <image>` you should get an error that says "unknown plugin" and not default to the human-readable report

Is this the outcome you expect or want? It sounds reasonable to me but just checking :)

rnjudge:
> Regarding the change in the CLI:
> 
>     1. When running `tern report -i <image>` you should get the default human-readable report
> 
>     2. When running `tern report -m spdxtagvalue -i <image>` you should get the spdx tag-value format
> 
>     3. When running `tern report -m random -i <image>` you should get an error that says "unknown plugin" and not default to the human-readable report
> 
> 
> Is this the outcome you expect or want? It sounds reasonable to me but just checking :)

Yes, this is my expected outcome :)

tern/formats/generator.py
rnjudge: Done! Renamed `GenerateReport` to just `Generate` and inherited from ABCMetaclass.

tern/formats/generator.py
rnjudge: Removed __init__.

tern/formats/generator.py
rnjudge: Fixed in the updated commit.

tern/report/report.py
rnjudge: Fixed in the updated commit.

tern/formats/spdx/spdxtagvalue/generator.py
rnjudge: Done.

nishakm:
All looks good. Merging

PR #393, opened 2019-08-20
classes: Removed unnecessary pass

nishakm:
Prospector complains of unnecessary pass in empty abstract
methods.

The test for the package mapping asserts an invalid string that
doesn't match what is in the test fixture. Changing the string
to the right one.


nishakm:
TravisCI test passed. Merging.

PR #397, opened 2019-08-21
ci/cd: Add a pre-deploy to use release reqs file

### LINKED ISSUE #389: Update PyPI deployment script to use the release requirements file
**Description**
This tasks should modify the CircleCI config.yml file to run steps when tagging a release listed in #386 

**Super Issues**
#386

rnjudge:
When we cut a release for Tern, we gather the direct and transient
dependencies via pip-compile. We copy the output of this command
to `docs/releases/vx_x_x-requirements.txt` in order to maintain
reproducibility for the future. However, when we build and release the
Tern PyPI package there is a chance that the requirements may change
between when we cut the release and when we build and release the
package to PyPI since the official requirements.txt does not pin to
exact versions and only lists the direct dependencies.

This commit adds a step to the deploy process that copies the
`docs/releases/vx_x_x-requirements.txt` file to `requirements.txt` so that
pbr and PyPI will package Tern using the same requirements recorded
when the release was cut.

Resolves #389

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,11 @@ commands:
            echo -e "[pypi]" >> ~/.pypirc
            echo -e "username = $R_PYPI_UN" >> ~/.pypirc
            echo -e "password = $PYPI_PW" >> ~/.pypiruc
+  pre_deploy:
+    steps:
+      - run:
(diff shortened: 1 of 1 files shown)


nishakm:
Looks good to me. Merging.

PR #404, opened 2019-08-24
Allow tern to install in dev mode

PrajwalM2212:
Currently developers have to unistall and install tern in order to test their changes. This can be solved by using pip install -e. tern also have dev requirements. This can also also be accomodated as pip install -e.[dev] by adding options.extras_require in setup.cfg

Resolves: https://github.com/vmware/tern/issues/403

Signed-off-by: Prajwal M prajwalmmath@gmail.com


PrajwalM2212:
`Do you think we will still need the dev-requirements.txt file with this workflow?`
We still use `dev-requirements.txt` in `circleci`.  We can replace it 
```python
pip install -r dev-requirements.txt
pip install .
```
with
```python
pip install .[dev]
```
I think then we can remove `dev-requirements.txt`

PR #420, opened 2019-08-28
Wrap python2.7 error message

PrajwalM2212:
Wraps python2.7 error message with user-friendly error message

Resolves https://github.com/vmware/tern/issues/362

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -145,7 +145,11 @@ def main():
     args = parser.parse_args()
 
     # execute
-    do_main(args)
+    if sys.version_info < (3, 0):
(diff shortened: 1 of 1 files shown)


PrajwalM2212:
@nishakm @rnjudge Can you review this ?

PrajwalM2212:
```Implementation Changes
The error is being hit in tern/classes/docker_image.py. Might consider using the traceback library: https://docs.python.org/3/library/traceback.html

Another option would be for Tern to check for dependencies somehow before running.
```

@rnjudge I chose to check for dependencies before running.  If I wrap the error in `docker_image.py`, it may turn up in some other class, so I thought that it would good to check for dependencies before running.

PR #428, opened 2019-08-29
Allow Tern to pull by digest

PrajwalM2212:
Tern actually has the functionality to pull by digest, but it is not evident as the docker python api does not explicitly specify it.

Resolves https://github.com/vmware/tern/issues/418

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -113,7 +113,10 @@ def main():
                                " image")
     parser_report.add_argument('-i', '--docker-image',
                                help="Docker image that exists locally -"
-                               " image:tag")
+                               " image:tag"
+                               " The option can be used to pull docker"
+                               " images by digest as well -"
+                               " <repo>@<digest-type>:<digest>")
     parser_report.add_argument('-s', '--summary', action='store_true',
                                help="Summarize the report as a list of"
                                " packages with associated information")


PrajwalM2212:
I started looking into the docker python api, when I found that RegistryData has pull by digest feature. https://github.com/docker/docker-py/blob/1fc62e6a32c2d3257d5d033fcf924fabb16190ac/docker/models/images.py#L148

Further inspecting the function we can see that it call pull https://github.com/docker/docker-py/blob/1fc62e6a32c2d3257d5d033fcf924fabb16190ac/docker/models/images.py#L160 which is same as pull we use in tern https://github.com/docker/docker-py/blob/1fc62e6a32c2d3257d5d033fcf924fabb16190ac/docker/models/images.py#L398

This can be further confirmed by looking into https://github.com/docker/docker-py/blob/8fea5738d32cfb712fcb05f541bc63eaf1d7ed71/docker/utils/utils.py#L207. We see that it has the capability to work on digests as well. 

So the pull function we currently use can pull based on digests as well.

nishakm:
@PrajwalM2212 This is great that the API supports this! The trouble is that tern doesn't. The argument `args.image` is expected to be a string of the form `image:tag` and not of the form `<repo>@<digest-type>:<digest>`.
You can see this in `setup` and `load_full_image` functions in `tern/report/report.py`. So the whole solution has a number of steps:
1. The `DockerImage` class ` __init__` function should take either an image id, a string of the form `repo:tag` or of the form `<repo>@<digest-type>:<digest>`. The `Image` superclass should also have a `digest_type` property or similar. There is an issue filed that is related to this: https://github.com/vmware/tern/issues/271. It's OK if you want to work on that one first. I don't think the superclass cares what the string is. It just assigns it to id and leaves all the other properties blank.
2. Under `tern/utils/container.py` there is a function called `check_image` which tries to look for an image of the form `image:tag` on disk before it attempts to pull. So you will have to first parse the input string to figure out what the user wants to do. Perhaps it is easier to check if the string is of the form `<repo>@<digest-type>:<digest>` and pull regardless. This check is in the `setup` function.
3. Because we had a beginning assumption that images given to tern are only of the form `image:tag` the parsing has mostly been to `image_tag_string.split(':')` to get the repo name and the tag name throughout the project. You will need to make one image string parsing function and use it throughout the project.

PrajwalM2212:
@nishakm 

I had already pulled debian:buster. I got the digest of that as `debian@sha256:2f04d3d33b6027bb74ecc81397abe780649ec89f1a2af18d7022737d0482cefe`
.
Then I deleted the image form my system.

Now,
Screen Shot 2019-08-30 at 8 59 58 PM
Screen Shot 2019-08-30 at 9 00 09 PM

Now `vi output.txt`
Screen Shot 2019-08-30 at 9 00 52 PM


So I can pull images successfully using digests without making any changes. 

The `check_image` in `tern/utils/container.py` calls `client.images.get(image_tag_string)`. The dc oker api recognises image of the form `<repo>@<digest-type>:<digest>`. 

Correct me if I am wrong.

nishakm:
@PrajwalM2212 you are correct - there isn't any change to the functionality. The change that is needed is just what a 'repo' and a 'tag' is. Currently if I use the whole digest string to pull an image, the image `tag` is the digest.
I suppose it doesn't hurt now but this is a good precursor to https://github.com/vmware/tern/issues/271 and I'll probably need to file more issues on this. Meanwhile, I will merge this.

Issue #226, opened 2019-04-12
Usage of tarfile.extractall() function without validating tarfile members


yilmi-vmw:
**Describe the bug**

The usage of the tarfile.extractall() function without validating members could be used to perform archive attacks.

This happens in two places in tern:
- tern/utils/container.py under the extract_image_metadata() function
- tern/utils/rootfs.py under the extract_layer_tar() function

**Expected behavior**

Two things can be done to protect against this attack:

1. Check the tarfile entries for directory traversal
(see warning under https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall)
2. Disable symlink creation

We should validate that the tarfile members names listed in its header don't contain invalid characters (../) and are not symlinks.

You can find a good description of how to protect against this here https://stackoverflow.com/a/10077309 and also an example of how to protect against these type of attacks here https://github.com/vmware/pyvcloud/pull/268

nishakm:
So this is interesting as the tarballs that Tern works on come from container images built by tools like Docker. It would be really cool to detect such issues with container images. The flip side to this is if most container images folks work with are subject to this attack, Tern can only sternly warn and continue until the problem is fixed in the container images.

A note: It seems that this issue *might* be fixed in Python 3.8 with the addition of a class `SafeTarFile` that satisfies most, if not all, attacks from a directory walk: https://bugs.python.org/issue21109. I doubt it though, as there is only 1 month left till feature freeze.

nishakm:
On further review of the conversations in the Python issue, I think they're not going to merge their fix any time soon, mostly because the security issues with the Python library inherit from the underlying security issues with `tar`. The suggested method of checking for absolute path, relative path and symlinks are not enough. Furthermore, I think there is a good subset of container images for which these checks will break untarring. The general suggestion is to be careful of what you untar.

However, tricking someone into downloading a naughty image from Dockerhub is definitely a thing, so I don't think we should ignore it completely.

There's a patch which takes care of most of the issues: https://bugs.python.org/file47826/safetarfile-4.diff

rparikh:
@nishakm I would like to work on this if it is not assigned yet. Please let me know.

Thank you, Ravi.

yilmi-vmw:
@nishakm, thanks for looking at the best approach to solve it.

[...]

Issue #430, opened 2019-08-30
When Tern executes fs_hash.sh, it cannot cd to the contents directory.


Katsunori04:
Hello, Tern Team.
When I tried to analyze the "yarara / base: v1" image, Tern output the following error.
Tern can't use command's "cd".
Is it my environmental problem or the Tern specification problem?

**Problem Statement**
When Tern executes fs_hash.sh, it cannot cd to the contens directory.

**To Reproduce**
```
vagrant@ubuntu-bionic:~$ tern -l  -k report -i yarara/base:v1
fatal: not a git repository (or any of the parent directories): .git
2019-08-30 06:46:33,865 - DEBUG - __main__ - Starting...
2019-08-30 06:46:33,866 - DEBUG - report - Setting up...
2019-08-30 06:46:33,866 - DEBUG - container - Checking if image "yarara/base:v1" is available on disk...
2019-08-30 06:46:33,871 - DEBUG - container - Attempting to pull image "yarara/base:v1"
2019-08-30 06:46:56,728 - DEBUG - container - Image "yarara/base:v1" downloaded
2019-08-30 06:47:05,844 - DEBUG - rootfs - Running command: sudo /usr/local/lib/python3.6/dist-packages/tern/tools/fs_hash.sh /home/vagrant/temp/3bf9dc6ca1042e8f1e9ce7b6a62bbe1ea0c224323b6fbc34e5e5e35905cd4206/contents
2019-08-30 06:47:08,952 - ERROR - rootfs - Command failed. /usr/local/lib/python3.6/dist-packages/tern/tools/fs_hash.sh: 22: cd: can't cd to /home/vagrant/temp/3bf9dc6ca1042e8f1e9ce7b6a62bbe1ea0c224323b6fbc34e5e5e35905cd4206/contents

2019-08-30 06:47:08,953 - WARNING - report - Cannot retrieve full image metadata
2019-08-30 06:47:08,953 - DEBUG - report - Teardown...
2019-08-30 06:47:08,954 - DEBUG - rootfs - Running command: sudo rm -rf temp/mergedir
2019-08-30 06:47:08,965 - DEBUG - rootfs - Running command: sudo rm -rf temp/workdir
2019-08-30 06:47:08,975 - DEBUG - __main__ - Report completed.
2019-08-30 06:47:08,976 - DEBUG - __main__ - Finished
```

nishakm:
Hello!

[...]

PR #435, opened 2019-08-30
Extract tarballs in Ubuntu:14.04

PrajwalM2212:
Some tarballs cannot be extracted without running as root. Hence tar will be extracted by running the root command

Resolves https://github.com/vmware/tern/issues/433

### DIFF
--- a/tern/utils/rootfs.py
+++ b/tern/utils/rootfs.py
@@ -10,7 +10,6 @@ import hashlib
 import logging
 import os
 import subprocess  # nosec
-import tarfile
 import pkg_resources
 
 from tern.utils import constants
@@ -24,6 +23,7 @@ mount_proc = ['mount', '-t', 'proc', '/proc']
 mount_sys = ['mount', '-o', 'bind', '/sys']
 mount_dev = ['mount', '-o', 'bind', '/dev']
 unmount = ['umount']
+extract_tar = ['tar', '-xvf']
 
 # enable host DNS settings
 host_dns = ['cp', constants.resolv_path]
@@ -58,8 +58,7 @@ def root_command(command, *extra):
     if error:
         logger.error("Command failed. %s", error.decode())
         raise subprocess.CalledProcessError(1, cmd=full_cmd, output=error)  # nosec
-    else:
-        return result
+    return result
 
 
 def get_untar_dir(layer_tarfile):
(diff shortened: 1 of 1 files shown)


tern/utils/rootfs.py
```diff
@@ -86,8 +86,8 @@ def set_up():
 def extract_layer_tar(layer_tar_path, directory_path):
     '''Assuming all the metadata for an image has been extracted into the
     temp folder, extract the tarfile into the required directory'''
-    with tarfile.open(layer_tar_path) as tar:
-        tar.extractall(directory_path)
+    os.makedirs(directory_path)
+    root_command(extract_tar, layer_tar_path, '-C', directory_path)
```
nishakm: Good catch! The trouble with shelling out to the system's tar utility is that it won't check for empty tarballs as we have seen in #430. I wonder if we can use `tarfile` to see if the tarball is empty or not before shelling out to it. If you can do this, I can close #226 as this change will resolve that one as well.

nishakm:
@PrajwalM2212 on line 61
``` 
else:
        return result
```
Can you remove the `else:` and just `return result`?

tern/utils/rootfs.py
PrajwalM2212: Yes. I will check for empty tarballs.

PrajwalM2212: @nishakm tarfile seems to not help. But `tar -tf` will help in this case. 
But bandit is throwing some security issues. Any clue on how to fix them ?

PrajwalM2212: @nishakm `tarfile` module is not helping. `tar -tf` helps.  But there are some security warnings associated with this. Any clue how to resolve them ?

If they can't be resolved then we may have to find another way to validate tar files.

nishakm: @PrajwalM2212 do you get any warnings when running `tar -tf` on an ubuntu image? I don't see any.

tern/utils/rootfs.py
```diff
@@ -86,8 +85,12 @@ def set_up():
 def extract_layer_tar(layer_tar_path, directory_path):
     '''Assuming all the metadata for an image has been extracted into the
     temp folder, extract the tarfile into the required directory'''
-    with tarfile.open(layer_tar_path) as tar:
-        tar.extractall(directory_path)
+    os.makedirs(directory_path)
+    with open(os.devnull, 'w') as test:
+        result = subprocess.call(['tar', '-tf', layer_tar_path],
```
nishakm: Oh! This is what you meant by security warning. You can run `root_command` here as well and check to see if there is an error or if the result is an empty string.

nishakm: `root_command` will take care of the pipe redirection for you and will return the result and an error message.

nishakm:
Closing this as the commit is in the PR #440

PR #439, opened 2019-09-04
classes: Changed separator for package id

nishakm:
When reading the package id, it is sometimes hard to distinguish
between the package name and the package version, especially if
the package version is of the Debian form:
[epoch:]upstream_version[-debian_revision]

- Replaced the separator '.' with '-'.
- Adjusted the unit test for the Package class accordingly
- Removed stray semicolon
- Shortened description line to less than 80 characters

### DIFF
--- a/tern/classes/package.py
+++ b/tern/classes/package.py
@@ -88,7 +88,7 @@ class Package:
 
     @checksum.setter
     def checksum(self, checksum):
-        self.__checksum = checksum;
+        self.__checksum = checksum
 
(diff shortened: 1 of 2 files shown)


rnjudge:
PR [#437](https://github.com/vmware/tern/pull/437) moves get_package_id functionality to the SPDX plugin since this value is explicitly associated with SPDX reports. I incorporated the changes from Nisha's pull request here to PR #437 so the reporting of package ID will use `name-version` instead of `name.version`.

PR #442, opened 2019-09-06
Check tarballs before extracting

### LINKED ISSUE #226: Usage of tarfile.extractall() function without validating tarfile members

nishakm:
This resolves #226

This change does the following:
1. Replace python's tarfile with a call to the system's tar
   utility. We do this to take advantage of the CVE-2013-4420 fix
   to libtar. Python's tarfile module has a fix in the workings
   but is yet to be merged as of this change.
2. We check for EOF errors and empty tarballs. This is mostly to
   address a few instances where we have seen Docker images that
   were malformed.
3. We modify some functions around loading and analyzing Docker
   images and layers including catching the extra errors that we
   raise for No. 2.

- rootfs: Moved some functionality from check_tar_permissions into
  a new function called shell_command. This function simply runs
  shell commands as the current user and returns the result and
  error to be dealt by the calling function.
- rootfs: check_tar_permissions will now use shell_command.
- rootfs: Created a new function called check_tar_members which
  will list the elements in the tarball to see if there are any
  EOF or empty tarballs.
- rootfs: Repurposed extract_layer_tar to be a general purpose
  extract_tarfile function which can be used throughout the code.
- container: Use extract_tarfile to extract image metadata.
- image_layer: Use extract_tarfile to extract image layer tarballs.
- analyze: Set up mount points after the image is loaded.
- report: In general setup, don't create directories. extract_tarfile
  will now do it.
- report: Catch all the appropriate errors that might get thrown when
  trying to load an image.

This was tested against an image from #430 and it reports the correct
failure and exists gracefully.

### DIFF
--- a/tern/report/report.py
+++ b/tern/report/report.py
@@ -65,11 +65,6 @@ def setup(dockerfile=None, image_tag_string=None):
                 logger.fatal("%s", errors.cannot_find_image.format(
                     imagetag=image_tag_string))
                 sys.exit()
-    # create temporary working directory
-    if not os.path.exists(constants.temp_folder):
-        os.mkdir(constants.temp_folder)
-    # set up folders for rootfs operations
-    rootfs.set_up()
 
 
 def teardown():
@@ -117,18 +112,13 @@ def load_base_image():
                 imagetag=base_image.repotag))
     try:
         base_image.load_image()
-    except NameError as error:
+    except (NameError,
+            subprocess.CalledProcessError,
+            IOError,
+            ValueError,
+            EOFError) as error:
         logger.warning('Error in loading base image: %s', str(error))
         base_image.origins.add_notice_to_origins(
-            dockerfile_lines, Notice(str(error), 'error'))
-    except subprocess.CalledProcessError as error:
-        logger.warning(
-            'Error in loading base image: %s', str(error.output, 'utf-8'))
-        base_image.origins.add_notice_to_origins(
-            dockerfile_lines, Notice(str(error.output, 'utf-8'), 'error'))
-    except IOError as error:
-        logger.warning('Error in loading base image: %s', str(error))
-        base_image.origins.add_notice_to_origin(
             dockerfile_lines, Notice(str(error), 'error'))
     return base_image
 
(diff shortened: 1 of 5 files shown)


tern/report/report.py
```diff
@@ -140,13 +130,12 @@ def load_full_image(image_tag_string):
...
     try:
         test_image.load_image()
-    except NameError as error:
-        test_image.origins.add_notice_to_origins(
-            failure_origin, Notice(str(error), 'error'))
-    except subprocess.CalledProcessError as error:
-        test_image.origins.add_notice_to_origins(
-            failure_origin, Notice(str(error.output, 'utf-8'), 'error'))
-    except IOError as error:
+    except (NameError,
```
rnjudge: This is nice.

tern/report/report.py
```diff
@@ -117,19 +112,14 @@ def load_base_image():
                 imagetag=base_image.repotag))
     try:
         base_image.load_image()
-    except NameError as error:
+    except (NameError,
```
rnjudge: This is nice!

rnjudge:
FYI, I tested `yarara/base:v1` image and got an empty tarball error message. I believe this is expected. I also tested `ubuntu:14.04` and it works. Finally, I tested against `nubs/arch-build:latest` and got the empty tarball message. So I believe this commit will work.

PR #452, opened 2019-09-19
Add files_analyzed property for Image Layer class

### LINKED ISSUE #444: Create a files_analyzed property in the ImageLayer class
**Description**
Create a property called `files_analyzed` in the ImageLayer class located in `tern/classes/image_layer.py`.

**Implementation**
In `tern/classes/image_layer.py`
1. Add `files_analyzed` to class docstring under `class ImageLayer:`
`files_analyzed: whether the files in this layer are analyzed or not`.
2. Add the property to `__init__` function and set it to False.
3. Add a getter and setter. The setter should first check if the input is a boolean (`True` or `False` in python) and if not then raise a ValueError with a message like "files_analyzed should be boolean".

In `tests/test_class_image_layer.py`, add tests for `files_analyzed` in `testInstance`:
1. Test if the default value is False
2. Test if the value can be set to True
3. Test if it raises a ValueError if you try to set it with a string 

**References**
Look in `tern/classes/image_layer.py` and `tests/test_class_image_layer.py` to see examples of how to implement this.

**Super Issues**
#284

manaswinidas:
Fixes #444


manaswinidas:
@nishakm I tried the exact steps. At first, I could see my commits only, then I'm still seeing the other commits :disappointed:

PR #453, opened 2019-09-19
Add files_analyzed property for ImageLayer class

### LINKED ISSUE #444: Create a files_analyzed property in the ImageLayer class

manaswinidas:
Fixes #444 

### DIFF
--- a/tern/classes/image_layer.py
+++ b/tern/classes/image_layer.py
@@ -28,6 +28,7 @@ class ImageLayer:
         last layer of the image that was imported
         import_str: The string from a build tool (like a Dockerfile) that
         created this layer by importing it from another image
+        files_analyzed: whether the files in this layer are analyzed or not
     methods:
         add_package: adds a package to the layer
         remove_package: removes a package from the layer
@@ -45,6 +46,7 @@ class ImageLayer:
         self.__import_str = ''
         self.__pkg_format = ''
         self.__os_guess = ''
+        self.__files_analyzed = False
 
     @property
     def diff_id(self):
@@ -69,7 +71,7 @@ class ImageLayer:
     @property
     def created_by(self):
         return self.__created_by
-
+        
     @created_by.setter
     def created_by(self, create_string):
         self.__created_by = create_string
@@ -106,6 +108,17 @@ class ImageLayer:
     def os_guess(self, os_guess):
         self.__os_guess = os_guess
 
+    @property
+    def files_analyzed(self):
+        return self.__files_analyzed
+        
+    @files_analyzed.setter
+    def files_analyzed(self, x):
+        if isinstance(x, bool):
+            self.__files_analyzed = x
+        else:
+            raise ValueError('files_analyzed should be boolean')
+    
     def add_package(self, package):
         if isinstance(package, Package):
             if package.name not in self.get_package_names():
--- a/tests/test_class_image_layer.py
+++ b/tests/test_class_image_layer.py
@@ -34,6 +34,11 @@ class TestClassImageLayer(unittest.TestCase):
         self.assertEqual(self.layer.pkg_format, 'rpm')
         self.layer.os_guess = 'operating system'
         self.assertEqual(self.layer.os_guess, 'operating system')
+        self.assertFalse(self.layer.files_analyzed)
+        self.layer.files_analyzed = True
+        self.assertTrue(self.layer.files_analyzed)
+        self.assertRaises(ValueError, setattr, self.layer,
+                          'files_analyzed', 'some string')
 
     def testAddPackage(self):
         err = "Object type String, should be Package"


manaswinidas:
@nishakm I'm aware that I forgot to add the body to the commit. But I want to commit it along with the changes I'll make to the pull request, I also wanted to ask whether force-pushing will bring all those commits here again. I don't want to mess up again.

nishakm:
@manaswinidas you did it!! I took a look at your branch on your fork and it looks like you are only one commit difference between master, so I don't think that amending your commit is going to pull any extra changes.
Let me do a review and post instructions here on how to update your commit with the required changes.

tern/classes/image_layer.py
```diff
@@ -106,6 +112,13 @@ def os_guess(self):
     def os_guess(self, os_guess):
         self.__os_guess = os_guess
 
+    @files_analyzed.setter
+    def files_analyzed(self, x):
+        if bool(x):
+            self.__files_analyzed = x
+        else:
+            raise ValueError('files_analyzed should be boolean')
+
```
nishakm: Could you move the setter with the `@property` so it is easier to find when looking at this code later?

tests/test_class_image_layer.py
```diff
@@ -34,6 +34,10 @@ def testInstance(self):
         self.assertEqual(self.layer.pkg_format, 'rpm')
         self.layer.os_guess = 'operating system'
         self.assertEqual(self.layer.os_guess, 'operating system')
+        self.assertFalse(self.layer.files_analyzed)
+        self.assertEqual(self.layer.files_analyzed, True)
```
nishakm: The way you do this is first set `self.layer.files_analyzed` to True and then `self.assertEqual` if it is True.

nishakm:
@manaswinidas Excellent work! There's a couple of changes I have asked to the code and, of course, a body for the commit message. Here are the steps to fix your commit:
1. Make the changes I had asked for in the code
2. Once you are satisfied with them, run `git add tern/classes/image_layer.py tests/test_class_image_layer.py` and `git commit --amend`. This will update your existing commit.
3. Then do `git push -f origin patch4` to update your PR.

tests/test_class_image_layer.py
manaswinidas: @nishakm thanks for the review. Should I add `assertTrue` in place of `assertFalse`? or keep both of that?

nishakm: First do:
`self.layer.files_analyzed = True`
Then do:
`self.assertEqual(self.layer.files_analyzed, True)`

manaswinidas: I added the `assertFalse` due to this tests that was mentioned in issue: Test if the default value is False. Then `assertFalse` isn't required?

nishakm: Oh, no the assertFalse is still required and it works because we've set it to `False` by default and it works.
The second bit we need is to see whether we can toggle `files_analyzed` to `True`.

So first set it to true and then you can do the `assertEqual`.

BTW, you can also say `self.assertTrue(self.layer.files_analyzed)` just like the `assertFalse` statement. But the test will fail if you don't set it to `True` first ;)

manaswinidas: Ok, thanks!

manaswinidas:
@nishakm I have made the changes. Please let me know if there are any other changes that I have to make.

tests/test_class_image_layer.py
nishakm:
``` 
self.assertFalse(self.layer.files_analyzed)
self.layer.files_analyzed = True <-- please add this line
self.assertTrue(self.layer.files_analyzed)
self.assertEqual(self.layer.files_analyzed, True) <-- please remove this line
```

Explanation: All the tests in the `tests` folder are python unit tests. To learn more about what unit tests are, check out: https://en.wikipedia.org/wiki/Unit_testing. We want to test if we can assign `True` to this value and it does stay `True` after assigning. I hope it makes sense.

manaswinidas: @nishakm updated the PR. Thanks for sharing the link.

tern/classes/image_layer.py
```diff
@@ -106,6 +108,17 @@ def os_guess(self):
     def os_guess(self, os_guess):
         self.__os_guess = os_guess
 
+    @property
+    def files_analyzed(self):
+        return self.__files_analyzed
+        
+    @files_analyzed.setter
+    def files_analyzed(self, x):
+        if bool(x):
```
nishakm: `bool(x)` casts the input `x` to a boolean value, so this will always be assigned.
Instead do: `if isinstance(x, bool):`

manaswinidas: Thanks a lot for pointing this out @nishakm

nishakm:
@manaswinidas One last request: It would be really nice if you could describe what you are changing. You can refer to #444 for commit message body content. Thanks again for working on this!

nishakm:
@manaswinidas All good! I will merge this. Thanks again!

PR #456, opened 2019-09-20
Set PackageDownloadLocation as NOASSERTION

### LINKED ISSUE #451: Set PackageDownloadLocation as NOASSERTION in SPDX tag:value report formatting
**Description**
Currently, Tern reports the SPDX `PackagetDownloadLocation` of an image layer as the docker image repotag. This does not reflect the actual download location and there is no way for Tern to pinpoint the download location of an image layer. Therefore, we should assign the SPDX `PackageDownloadLocation` value as NOASSERTION in accordance with the 2.1 spec. We can also remove the `get_download_location` function in `docker_image.py`

**Implementation**
1) In `tern/formats/spdx/spdxtagvalue/generator.py`, remove the `registry_repotag` variable assignment in the following lines:
```
235         # The image's PackageDownloadLocation is from a container registry      
236         # This includes all the layers but the packages' download location      
237         # is unknown if the download_url is blank                               
238         registry_repotag = image_obj.get_download_location() if hasattr(        
239             image_obj, 'repotag') else 'NOASSERTION'
```
2) In `tern/formats/spdx/spdxtagvalue/generator.py`, hardcode the `PackageDownloadLocation` value as `NOASSERTION`:
```
265                 PackageDownloadLocation=NOASSERTION,
```
3) In `tern/classes/docker_image.py` remove the definition for `get_download_location()` since it is not possible to get the actual location (dockerhub.io is not the actual location).
```
171     def get_download_location(self):                                            
172         '''A docker image's download location is in the repotags. We will       
173         return a string of repotags joined by a hyphen -'''                     
174         return '-'.join(self.repotags)
```

**References**
See above.

manaswinidas:
set PackageDownloadLocation

Fixes #451 

### DIFF
--- a/tern/formats/spdx/spdxtagvalue/generator.py
+++ b/tern/formats/spdx/spdxtagvalue/generator.py
@@ -232,11 +232,6 @@ class SpdxTagValue(generator.Generate):
         licenses_found = []  # This is needed for unrecognized license strings
         image_obj = image_obj_list[0]
         template = SPDX()
-        # The image's PackageDownloadLocation is from a container registry
-        # This includes all the layers but the packages' download location
-        # is unknown if the download_url is blank
-        registry_repotag = image_obj.get_download_location() if hasattr(
-            image_obj, 'repotag') else 'NOASSERTION'
 
         # first part is the document tag-value
         # this doesn't change at all
@@ -262,7 +257,7 @@ class SpdxTagValue(generator.Generate):
                 layer_obj.to_dict(template),
                 layer_obj.origins.origins,
                 SPDXID=get_layer_spdxref(layer_obj),
-                PackageDownloadLocation=registry_repotag,
+                PackageDownloadLocation='NOASSERTION',
                 PackageLicenseDeclared='NOASSERTION',
                 PackageLicenseConcluded='NOASSERTION',
                 PackageCopyrightText='NOASSERTION',
--- a/tern/classes/docker_image.py
+++ b/tern/classes/docker_image.py
@@ -167,8 +167,3 @@ class DockerImage(Image):
             raise
         except IOError:  # pylint: disable=try-except-raise
             raise
-
-    def get_download_location(self):
-        '''A docker image's download location is in the repotags. We will
-        return a string of repotags joined by a hyphen -'''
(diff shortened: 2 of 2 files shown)


nishakm:
@manaswinidas the test is failing for some weird reason. I'll check it out.

tern/formats/spdx/spdxtagvalue/generator.py
```diff
@@ -235,8 +235,6 @@ def generate(self, image_obj_list):
         # The image's PackageDownloadLocation is from a container registry
         # This includes all the layers but the packages' download location
         # is unknown if the download_url is blank
-        registry_repotag = image_obj.get_download_location() if hasattr(
```
rnjudge: @manaswinidas Since we are removing the registry_repotag assignment, could you also remove the comment above it?

rnjudge:
@manaswinidas Overall this commit looks good and we really appreciate your time working on it. Could you please take a look at your commit message and try to flush it out with a little more detail about the changes you are making? If it helps, you can take a look at the [issue description](https://github.com/vmware/tern/issues/451) use some of the text from it in your commit message.

rnjudge:
@manaswinidas This PR looks good and is ready for merge after a few small tweaks to the commit message. Primarily, adding a little more context for your changes will be super helpful to future contributors. You can pretty much copy and paste from the issue. Lastly, including the issue that this is resolving is important. Something like this would be great (feel free to copy and paste and use as your commit message:

```
 Set PackageDownloadLocation as NOASSERTION

Currently, Tern reports the SPDX PackagetDownloadLocation of an image 
layer as the docker image repotag. This does not reflect the actual 
download location and there is no way for Tern to pinpoint thei 
download location of an image layer. Therefore, we should assign the 
SPDX PackageDownloadLocation value as NOASSERTION in accordance with 
the 2.1 spec. This commit also removes the get_download_location 
function in docker_image.py

Resolves #451 

```

Once the commit message is updated I think this PR is ready for merge! :)

rnjudge:
Thank you @manaswinidas!

rnjudge:
NOTE to my future self: merging this PR even though test_changes failed because the test was fixed via PR #457 which has already been merged.

PR #459, opened 2019-09-24
Report package information by layer

rnjudge:
Currently, Tern summarizes all the package names, versions and licenses
found throughout the container image at the end of the report
with no way to discern which layer the package and license information
was found in. This commit changes the output of the default
format to report the package + license information by image layer. It
also adds a summary section to the bottom of the report summarizing the
full list of licenses found throughout all the layers of the container
image. This separation of information between layers will be helpful
for larger container images with multiple layers.

### DIFF
--- a/tern/formats/default/generator.py
+++ b/tern/formats/default/generator.py
@@ -23,8 +23,7 @@ def print_full_report(image):
     '''Given an image, go through the Origins object and collect all the
     notices for the image, layers and packages'''
     notes = ''
-    pkg_list = []
-    license_list = []
+    full_license_list = []
     for image_origin in image.origins.origins:
         notes = notes + content.print_notices(image_origin, '', '\t')
     for layer in image.layers:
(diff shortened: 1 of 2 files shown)


rnjudge:
@nishakm This commit yields an ouput report that lists the packages and licenses by layer but also summarizes the full license list at the bottom of the report. I decided to keep the list of licenses found in each layer for a circumstance where a user might typically only look at the full license list, but then comes across a prohibited or concerning license in the summary. They may want to be able to go back and determine which layer pulls in that license.

Do you think this use case is too specific? If so, I can remove the licenses summary by layer and just keep the one at the end.

nishakm:
> @nishakm This commit yields an ouput report that lists the packages and licenses by layer but also summarizes the full license list at the bottom of the report. I decided to keep the list of licenses found in each layer for a circumstance where a user might typically only look at the full license list, but then comes across a prohibited or concerning license in the summary. They may want to be able to go back and determine which layer pulls in that license.
> 
> Do you think this use case is too specific? If so, I can remove the licenses summary by layer and just keep the one at the end.

No, I think that's a good use case to have the license information repeated per layer.

PR #460, opened 2019-09-25
Add analyzed_output property

### LINKED ISSUE #445: Create an analyzed_output property for each layer with appropriate tests
***Description***
Create a property called `analyzed_output` in the ImageLayer class located in `tern/classes/image_layer.py`.

***Implementation***
In `tern/classes/image_layer.py`:

1. Add `analyzed_output` to class docstring under class ImageLayer:
`analyzed_output: the result of the file analysis`.
2. Add the property to `__init__` function and set it to an empty string.
3. Add a getter and setter. The setter should check if the input is a string and if not, then raise a ValueError with a message like "analyzed_output should be a string"

In `tests/test_class_image_layer.py`, add tests for files_analyzed in `testInstance`:
1. Test if the default value is and empty string.
2. Test if the value can be set to any string.
3. Test if a ValueError is raised if you set it to an `int`.

References
Look in tern/classes/image_layer.py and tests/test_class_image_layer.py to see examples of how to implement this.

Super Issues
#284

manaswinidas:
Create a property called analyzed_output in the
 ImageLayer class located in tern/classes/image_layer.py

Resolves #445

### DIFF
--- a/tern/classes/image_layer.py
+++ b/tern/classes/image_layer.py
@@ -29,6 +29,7 @@ class ImageLayer:
         import_str: The string from a build tool (like a Dockerfile) that
         created this layer by importing it from another image
         files_analyzed: whether the files in this layer are analyzed or not
+        analyzed_output: the result of the file analysis
     methods:
         add_package: adds a package to the layer
         remove_package: removes a package from the layer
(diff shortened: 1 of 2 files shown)


rnjudge:
This looks good @manaswinidas. Nice work!

PR #467, opened 2019-09-30
WIP: Precautions against external interference

nishakm:
Sometimes, a user may issue a keyboard interrupt during container
image analysis if it takes too long. In this situation, the working
folder will remain with all mount points still mounted. This
includes /sys, /dev, and /proc which are bind mounted to the host
machine. If the user tries to remove this folder using elevated
privileges, some of the devices and processor information will get
deleted as well and the host machine will be corrupted for that
session. Although this state is recoverable on reboot, it isn't
an ideal situation for a user to be in.

We deal with this problem by checking for the KeyboardInterrupt
exception during analysis. There are two places for this - analysis
using the system level tools and analysis using snippets.

- Added a function in rootfs.py to recover the mounted filesystems
- Added a "Keyboard Interrupt" error to the list of errors
- Added a function in analyze.py to use the recover function and
  exit gracefully

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -58,8 +58,17 @@ def get_version():
     return message
 
 
+def create_top_dir():
+    '''Create the top level working directory'''
+    top_dir = general.get_top_dir()
+    if not os.path.isdir(top_dir):
+        os.mkdir(top_dir)
+
+
 def do_main(args):
     '''Execute according to subcommands'''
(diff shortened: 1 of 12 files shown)


rnjudge:
I tested this PR code and it looks good. after I ctrl-c I can `rm -rf` the temp directory without deleting any important mounted `/dev` resources.

PR #468, opened 2019-10-01
Enable cli to use raw container image

### LINKED ISSUE #64: Enable CLI to use raw container image
Something like 'tern report --image/-i point to a local tarball

rnjudge:
This commit adds functionality so that Tern can accept a local raw
image tarball as an alternative to the current image:tag string option.
This functionality is implemented using an additional command line
option 'report -w' or 'report --raw-image'. If the user invokes Tern
with the --raw-image command line option and the input is corrupted or
does not exist at the specified location, Tern will abruptly and
gracefully exit.

Resolves #64

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -14,10 +14,12 @@ import logging
 import os
 import sys
 
+from tern.analyze import common
 from tern.analyze.docker import run
 from tern.utils import cache
 from tern.utils import constants
 from tern.utils import general
+from tern.report import errors
 
 
 # global logger
@@ -82,8 +84,18 @@ def do_main(args):
         if args.dockerfile:
             run.execute_dockerfile(args)
         if args.docker_image:
-            run.execute_docker_image(args)
-        logger.debug('Report completed.')
+            if common.check_tar(args.docker_image):
+                logger.error("%s", errors.incorrect_raw_option)
+            else:
+                run.execute_docker_image(args)
+                logger.debug('Report completed.')
+        if args.raw_image:
+            if not common.check_tar(args.raw_image):
+                logger.error("%s", errors.invalid_raw_image.format(
+                    image=args.raw_image))
+            else:
+                run.execute_docker_image(args)
+                logger.debug('Report completed.')
     logger.debug('Finished')
 
 
@@ -128,6 +140,9 @@ def main():
                                " The option can be used to pull docker"
                                " images by digest as well -"
                                " <repo>@<digest-type>:<digest>")
+    parser_report.add_argument('-w', '--raw-image', metavar='FILE',
+                               help="Raw container image that exists locally "
+                               "in the form of a tar archive.")
     parser_report.add_argument('-f', '--report-format',
                                metavar='REPORT_FORMAT',
                                help="Format the report using one of the "
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -9,6 +9,7 @@ Common functions
 
 import logging
 import os
+import tarfile
 
 from tern.classes.package import Package
 from tern.classes.notice import Notice
@@ -446,3 +447,11 @@ def get_os_style(image_layer, binary):
                     package_manager=binary,
                     package_format=image_layer.pkg_format,
                     os_list=image_layer.os_guess), 'info'))
+
+
+def check_tar(tar_file):
+    '''Check if provided file is a valid tar archive file'''
+    if os.path.exists(tar_file):
+        if tarfile.is_tarfile(tar_file):
+            return True
+    return False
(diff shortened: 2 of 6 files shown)


tern/analyze/common.py
```diff
@@ -445,3 +446,13 @@ def get_os_style(image_layer, binary):
...
+
+
+def check_tar(tar_file):
+    '''Check if provided file is a valid tar archive file'''
+    try:
+        if tarfile.is_tarfile(tar_file):
+            return True
+    except FileNotFoundError:
+        pass
+    return False
```
nishakm: Can we not `return False` in the exception handling?

tern/__main__.py
```diff
@@ -73,8 +75,18 @@ def do_main(args):
...
-        logger.debug('Report completed.')
+            if common.check_tar(args.docker_image):
+                logger.error("%s", errors.incorrect_raw_option)
+            else:
+                run.execute_docker_image(args)
+                logger.debug('Report completed.')
+        if args.raw_image:
+            if not common.check_tar(args.raw_image):
+                logger.error("%s", errors.invalid_raw_image.format(
+                    image=args.raw_image))
```
nishakm: What do you think about adding an option `-t, --type` and one can enter a type like `docker` or `oci` and then one can provide either `-i, --image` and for this input you can detect whether the input is an existing tar archive or a string ?

tern/report/errors.py
```diff
@@ -52,6 +52,10 @@
     '''{default_shell}\n'''
 unknown_content = '''Unknown content included in layer {files}. Please ''' \
     '''analyze these files separately\n'''
+invalid_raw_image = '''Invalid raw image provided: '{image}' - Check ''' \
+    '''that path to raw image is correct and in tar archive format.'''
+incorrect_raw_option = '''Raw image detected. Expected docker image ''' \
+    '''with image:tag format.'''
```
nishakm: The API will actually take an image string of the form `image@<digest-type>:<digest>` so maybe make this more generic?

tern/analyze/common.py
rnjudge: I originally had the `return false` inside the exception handling, however, pylint complained:
```
pylint: inconsistent-return-statements / Either all return statements in a function should return an expression, or none of them should.
```

nishakm: Oh, it's because there is an `if` statement after the `try`. What about putting a `logger.error` statement instead of `pass`?

tern/__main__.py
rnjudge: @nishakm I like the `--type` suggestion to determine if the image type is oci or docker, especially for future support of other image formats. The issue I believe lies in automatically detecting image type. I originally designed this PR to use the existing `--image` option to automatically detect whether the input was a tarfile or a string. The trouble I see with doing it this way is that it's impossible to detect the users intent for the image they provided.

If the user enters a typo in the name or path of their tarfile while running Tern, the `check_tar`  function to see if the tarfile is valid will return false, which would thn indicate that it is a string/docker image type input. This would be incorrect but Tern would continue to try to pull "typo.tar" from dockerhub. 

If we want to automatically detect raw tar format vs docker image format, we can do it but we would have to accept the following error to the terminal when we assume that anything not a tarball is a docker hub image tag string (even in the case of typos).

```
2019-10-02 15:06:38,077 - DEBUG - container - Checking if image "typo.tar" is available on disk...
2019-10-02 15:06:38,098 - DEBUG - container - Attempting to pull image "typo.tar"
2019-10-02 15:06:40,404 - WARNING - container - No such image: "typo.tar"
2019-10-02 15:06:40,404 - CRITICAL - report - Cannot find image typo.tar locally or from remote registry.
```

My decision for implementing an explicit `raw` option was to allow Tern to provide more helpful and intuitive error messages. If the user selects the raw option and it is not a valid tarball, Tern can provide specific feedback that it is not of proper tar format.

tern/analyze/common.py
rnjudge: Because there are a few places I use this check_tar() functionality I am hesitant to put a specific error message in the log when it is called. I was hoping to keep this function as a simple utility and we can handle the error message on a case by case basis depending on how/where in the code the function is invoked.

nishakm: What about:
```
if os.path.exists(tar_file):
    if tarfile.is_tarfile(tar_file):
        return True
return False
```

tern/__main__.py
nishakm: OK, this makes sense considering "STRING" is a different data type than "FILE"

tern/__main__.py
```diff
@@ -119,6 +131,9 @@ def main():
                                " The option can be used to pull docker"
                                " images by digest as well -"
                                " <repo>@<digest-type>:<digest>")
+    parser_report.add_argument('-w', '--raw-image', metavar='RAW_IMAGE',
```
nishakm: Can `metavar` here be `FILE`?

tern/report/errors.py
rnjudge: @nishakm How does "Expected docker image but detected file in tar archive format." sound?

PR #469, opened 2019-10-02
spdx: Remove invalid characters from LicenseRef

### LINKED ISSUE #465: Remove comma seperators from SPDX LicenseRef values
**Describe the bug**

**To Reproduce**
1. Run an spdxtagvalue report with `photon:3.0`
2. Look at the output report and try to validate it.
3. The output report will fail because of a comma separator between a LicenseRef value.

**Proposed Solution***
Substitute any commas in the LicenseRef value with a `-` per the SPDX 2.1 spec

**Expected behavior**
The resulting spdxtagvalue report should validate.

Enter all that apply
- Working off of latest commit
- Ubuntu VMware fusion VM

rnjudge:
Currently, if you try to generate an spdxtagvalue report from a photon
image, the resulting report will not validate in the SPDX online tool.
This is because when the SPDX tool parses the PackageLicenseDeclared
value for LicenseRef-[idstring] the idstring value may only be
"a unique string containing letters, numbers, “.” or “-”" according to
the 2.1 SPDX spec.

Instead of doing a bunch of string manipulation to get the license
[idstring] value to conform to SPDX standards, this commit takes the
license string value and creates a unique SHA-256 hash value to use
as the LicensesRef-[idstring] value. Using a unique hash value instead
of a human-readable string is simpler and removes the need for a
handful of helper functions (get_package_licenses(), format_license(),
and update_license_list()). A human-readable description of the license
text can still be found at the end of the spdxtagvalue document.

Resolves #465

### DIFF
--- a/tern/formats/spdx/spdxtagvalue/generator.py
+++ b/tern/formats/spdx/spdxtagvalue/generator.py
@@ -8,6 +8,7 @@ SPDX document generator
 """
 
 import datetime
+import hashlib
 
 from tern.formats.spdx.spdx import SPDX
 from tern.utils.general import get_git_rev_or_version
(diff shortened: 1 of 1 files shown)


tern/formats/spdx/spdxtagvalue/generator.py
```diff
@@ -143,6 +143,9 @@ def format_license(license_string):
     We will split up the licenses by spaces, prepend each string with a
     "LicenseRef-" and then join the strings with an " AND "'''
     license_list = get_package_licenses(license_string)
+    for l in license_list:
+        if 'and' in l or 'or' in l:
+            license_list.remove(l)
```
nishakm: How about using `re.sub` in case there are more unacceptable words we will have to deal with in the future?

tern/formats/spdx/spdxtagvalue/generator.py
```diff
@@ -116,7 +116,7 @@ def get_layer_relationships(layer_obj, prev_layer_spdxref=None):
 
 def get_license_ref(pkg_license):
     '''Given one license, return a LicenseRef'''
-    return 'LicenseRef-' + pkg_license
+    return 'LicenseRef-' + pkg_license.replace(',', '-')
```
nishakm: How about using re.sub here for both `,` and `+` and for any other unacceptable characters we may have to deal with in the future? This way, we will not have to do it again in license block function and the generator function.

tern/formats/spdx/spdxtagvalue/generator.py
rnjudge: @nishakm I'm confused how I would use `re.sub` here. Do you mean use `re.search` to find the elements in the list? Not sure how re.sub would help me here as I want to compare the full license string and if it matches completely with and or or, remove the complete license element from the list. I think re.sub will manipulate just the string. But if we manipulate just the license string, even by substituting it with an empty string it would end up as an empty element in the list which we wouldn't want either. Thanks in advance :)

nishakm: Oh I see what you mean. So I guess what's going on is that `get_package_licenses` splits by `' '` but some of these licenses are true license expressions using 'and' and 'or' (just not in caps).

OK, we can look at a more sophisticated method of weeding out true license expressions and old ones later. For now, can you move this change into the get_package_licenses function?

PR #486, opened 2019-11-05
release: Prep for Release 1.0.0

nishakm:
* We're using 2FA to upload PyPI packages. Since there is no
  support for 2FA for headless accounts yet, we will be uploading
  manually. Hence we disable the Release workflow in CircleCI.
* The release checklist has steps to upload a PyPI release manually.
  It also has some clarifications on what the release number looks
  like and how to get the changelog and authors information.
* Added new requirements to requirements.in
* Bumped up versions for docker and stevedore in requirements.txt
* Added freeze file for requirements.txt.
* Added release notes.


docs/releases/release_checklist.md
```diff
@@ -25,13 +25,23 @@ This is a checklist for cutting a release
...
 - [ ] Commit release notes, `v<release>-requirements.txt`, and any changes to `requirements.txt`.
 
-- [ ] Tag release on GitHub. Check to see if release automation works.
+- [ ] Tag release on GitHub.
+    * Add new tag
+    * Provide a link to the release notes.
+
+- [ ] Deploy to PyPI
+    * Run `git fetch --tags` to get the release tag
+    * Run `git checkout <release tag>
```
nishakm: should be `git checkout -b release <release tag>`

.circleci/config.yml
```diff
@@ -101,13 +101,3 @@ workflows:
       - commit_check
       - security
       - test_changes
-  Release:
-    jobs:
-      - pypi_deploy:
```
rnjudge: There is a pypi_deploy definition a few lines up from here. We should probably remove it as well:

```
 82   # Deploy to PyPi
 83   pypi_deploy:
 84    executor: ubuntu1604
 85    steps:
 86       - setup
 87       - run: pip install .
 88       - pre_deploy
 89       - run: pip install twine
 90       - run: pip install wheel
 91       - create_pypirc
 92       - run: python setup.py sdist bdist_wheel
 93       - run: twine check dist/*
 94       - run: twine upload dist/*
 95 
```

rnjudge: @nishakm and while we are at it... should probably remove the `create_pypirc` and `pre_deploy` commands if we're not using it to automatically deploy to pypi:

```
 23   create_pypirc:
 24     steps:
 25       - run: 
 26           name: Create pypirc file
 27           command: |
 28            echo -e "[distutils]" >> ~/.pypirc
 29            echo -e "index-servers =" >> ~/.pypirc
 30            echo -e "    pypi\n" >> ~/.pypirc
 31            echo -e "[pypi]" >> ~/.pypirc
 32            echo -e "username: $R_PYPI_UN" >> ~/.pypirc
 33            echo -e "password: $PYPI_PW" >> ~/.pypirc
 34   pre_deploy:
 35     steps:
 36       - run:
 37           name: Ensure that PyPI packages Tern with the same requirements used when the release was cut
 38           command: cp docs/releases/${CIRCLE_TAG//./_}-requirements.txt requirements.txt
```

nishakm: I don't think we should. If PyPI enables 2FA for headless accounts, then it would only be a small change to revert. It's just workflows that get executed by circleci right?

PR #489, opened 2019-11-06
release: New post release flow and corrections

nishakm:
* Added post release steps to create a tarball of the sources of
  the dependencies and to upload the wheel package and the sources
  tarball to the GitHub release page.
* Made a correction when doing the creation of the pip package
  because setup.py doesn't understand the backslash in the snapshot
  requirements file with package checksums.

### DIFF
--- a/docs/releases/release_checklist.md
+++ b/docs/releases/release_checklist.md
@@ -38,7 +38,7 @@ This is a checklist for cutting a release
 - [ ] Deploy to PyPI
     * Run `git fetch --tags` to get the release tag
     * Run `git checkout -b release <release tag>
-    * Overwrite `requirements.txt` with `docs/releases/v<release>-requirements.txt`
+    * Run `pip compile`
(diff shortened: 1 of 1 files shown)


docs/releases/release_checklist.md
```diff
@@ -47,3 +47,10 @@ This is a checklist for cutting a release
     * Create a fresh environment.
     * Pip install tern.
     * Run appropriate tests.
+
+- [ ] Prepare sources tarball.
+    * In the release environment, create a new directory called `vendor`
+    * Run `pip install -d vendor --require-hashes --no-binary :all: -r docs/releases/v1_0_0-requirements.txt`
```
rnjudge: Do you want to change this to be more generic? i.e.
```
docs/releases/v<release>-requirements.txt
```

nishakm: Oops! Absolutely! Thanks for catching this.

PR #495, opened 2019-11-19
Update Dockerfile to include Tar

brphelps:
In our testing of the tern repo, we saw failures wherever tar was used. Added it to the list of installed packages and the errors went away.


nishakm:
@brphelps Thanks for submitting a PR! There is another PR that fixes this issue that was submitted before that one, so we will merge that one. We are also working on an underlying issue with running the resulting Docker image, so watch for another release. Meanwhile, if you are playing around with the tool, please use the Vagrant image instead. Thanks again!

PR #511, opened 2019-12-19
executor: scancode: Use pip package

nishakm:
Scancode now has a working pip package. Update the executor to
use it instead of running it from the cloned git repository.

- Updated the executor's documentation.
- Called the general scancode package which should be installed
either on the system or in a virtual environment.
- Updated the README with instructions on how to run tern with
scancode


nishakm:
@rnjudge we'll have to cut another release for the documentation to be valid.

rnjudge:
> @rnjudge we'll have to cut another release for the documentation to be valid.

@nishakm should we wait for the other scancode fix that's coming?

nishakm:
> > @rnjudge we'll have to cut another release for the documentation to be valid.
> 
> @nishakm should we wait for the other scancode fix that's coming?

We don't have to. But the script will fail if we merge this in and don't release an equivalent PyPI package. I say we wait till after the holidays to merge this change in.

Issue #513, opened 2020-01-07
Prospector issues warnings on freshly cloned repository


oc37ejuc:
**Describe the bug**
Cloning the Repository, and installing dependencies as described in the contribution guidelines, yields a codebase that does not pass the prospector tests.

**To Reproduce**
1. `pip install wheel`
2. `pip install -e.[dev]`
3. `prospector .`

**Error in terminal**
```
Messages
========

tern/analyze/docker/analyze.py
  Line: 122
    pylint: too-many-branches / Too many branches (18/15)

tern/analyze/docker/dockerfile.py
  Line: 50
    pylint: no-else-continue / Unnecessary "else" after "continue" (col 12)
    pylint: no-else-continue / Unnecessary "else" after "continue" (col 12)

tern/analyze/docker/run.py
  Line: 102
    pylint: too-many-branches / Too many branches (24/15)

tern/classes/image.py
  Line: 111
    pylint: unnecessary-pass / Unnecessary pass statement (col 8)
  Line: 153
    pylint: unnecessary-pass / Unnecessary pass statement (col 8)

tern/classes/notice.py
  Line: 11
    pylint: unnecessary-pass / Unnecessary pass statement (col 4)

tern/formats/default/generator.py
  Line: 22
    pylint: too-many-branches / Too many branches (16/15)

tern/formats/json/generator.py
  Line: 1
    pylint: cyclic-import / Cyclic import (tern.analyze.common -> tern.report.content -> tern.command_lib.command_lib -> tern.analyze.docker.container)
    pylint: cyclic-import / Cyclic import (tern.analyze.common -> tern.command_lib.command_lib -> tern.analyze.docker.container)

tern/formats/spdx/spdxtagvalue/generator.py
  Line: 142
    pylint: too-many-branches / Too many branches (16/15) (col 4)



Check Information
=================
         Started: 2020-01-06 14:50:48.797866
        Finished: 2020-01-06 14:50:59.745797
      Time Taken: 10.95 seconds
       Formatter: grouped
        Profiles: default, no_doc_warnings, no_test_warnings, strictness_medium, strictness_high, strictness_veryhigh, no_member_warnings
      Strictness: None
  Libraries Used: 
       Tools Run: dodgy, mccabe, pep8, profile-validator, pyflakes, pylint
  Messages Found: 11
```

**Expected behavior**
prospector should not issue warnings.

PR #516, opened 2020-01-08
circleci: Update prospector profile

rnjudge:
This commit makes three changes:

1) Bumps the minimum version of prospector installed on the circleci VMs
   from 1.1 to 1.2.

2) Adds a prospector.yaml profile to increase prospector code linting
   and control the checks and their strictness.
   This yaml file takes suggestions from Michael Rohan in
   https://github.com/vmware/tern/pull/356.

See also #513

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
     # steps to run Prospector
     steps:
       - setup
-      - run: pip install prospector>=1.1 GitPython~=2.1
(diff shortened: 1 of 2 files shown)


nishakm:
@rnjudge I think the "too many branches" errors are valid in the sense it helps us reduce function complexity. So I think we still need to address those errors if they arise. How about we file a technical debt bug for those errors and fix them later?

PR #526, opened 2020-01-22
Bump prospector version and disable pep8 N802

nishakm:
- Bump up prospector version in the develpment requirements so
it matches the version that we use on CI.
- Disable N802 check - function name should be lower case. We do
this because unittest in particular wants the functions to be in
CamelCase or else it is unable to identify them during testing.

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -7,6 +7,8 @@ test-warnings: false
 
 pep8:
   full: true
+  disable:
+    - N802 # pep8-naming: function name should be lower case
 
 pep257:
   run: false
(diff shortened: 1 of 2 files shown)


.prospector.yaml
```diff
@@ -7,6 +7,8 @@ test-warnings: false
 
 pep8:
   full: true
+  disable:
+    - N802
```
rnjudge: @nishakm do you think it could be helpful to add a comment above or after this error code specifying what it is exactly?

nishakm: Let me try...

nishakm: @rnjudge I've added a comment

PR #529, opened 2020-02-02
setup.py: Discover tests

PrajwalM2212:
This PR makes it possible for the tests
to be discovered correctly by adding
a function get_test_suite in setup.py.

Closes https://github.com/vmware/tern/issues/406

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,15 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from setuptools import setup
+import unittest
+
+
+def get_test_suite():
+    return unittest.TestLoader().discover('tests', 'test_*.py')
 
 
 setup(
     setup_requires=['pbr'],
     pbr=True,
-    test_suite="tests.runtests",
+    test_suite="setup.get_test_suite",
 )


PrajwalM2212:
Earlier there was an error in loading test suite
Screen Shot 2020-02-02 at 11 24 32 AM

Now the tests are discovered ad run correctly
Screen Shot 2020-02-02 at 11 38 15 AM

rnjudge:
This PR looks good to me. Will merge after a review from @nishakm.

rnjudge:
@PrajwalM2212 Thanks for this PR! This will be helpful to run all the unit tests at once. When I run `python setup.py test` this change works for me to run the test suite.

However, I was looking through setuptools [documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#test) and saw that `test` is deprecated in version 41.5.0 in favor of [tox](https://tox.readthedocs.io/en/latest/) and the `test` command will be removed in future versions of the package. Therefore, I wonder if this change should be implemented with tox instead of setuptools test option to save ourselves trouble in the future? @PrajwalM2212 @nishakm thoughts?

nishakm:
I agree with @rnjudge that it would be worth it to move to tox as that seems to be the testing standard for Python projects. Also, this would be a nice GSoC project to work on, unless @PrajwalM2212 wants to take it on now :)

rnjudge:
@PrajwalM2212 is working on a PR to use tox. Closing this PR.

Issue #405, opened 2019-08-25
Fix DockerImage class unit tests


PrajwalM2212:
**Problem Statement**
The import statements on test files are wrong and need to be rearranged.

**Describe the Proposal**
Describe the change you would like to see in the project.
https://github.com/vmware/tern/blob/master/tests/test_class_docker_image.py
uses 
```python
from classes.docker_image import DockerImage
from utils.container import docker_command
from utils.container import pull
from utils.container import check_image
```
but 
https://github.com/vmware/tern/commit/8092d0fde3059de9115ae485fd7d09cf67642e22 along with a number of PR's had made refactors a lot of code making these import statements outdated. 
This has to be fixed for the tests to run

**Steps to Implement Proposal**
Refactor the import statements in the test files.

nishakm:
In order to make this a "real" unit test, we need a test Docker image. But we do not have a project Docker repository. We will wait until tern has its own Dockerhub repo.

nishakm:
So we have a repo now `https://hub.docker.com/r/vmware/tern`. Next thing we need is a dummy test image. We can create one this way:
1. Create a simple directory structure:
```
foo
     |_ file1 <-owned by user
     |_ file2
     |_ bar
              |_ file1 <-owned by root
              |_ file2
```
2. Run fs_hash.sh on this directory, get the checksum and save it somewhere
3. Tar up this directory
4. Do a `docker load < testdir.tar`
5. Do a `docker push vmware/tern/testimage:<date>`
6. Do a `docker save` of this image, untar the contents and save all of the checksums, manifest and config information for the test.
7. Write the test that checks to see if all the content is the same for this image.

Note that we should be pulling this image by digest if we are to run tests against it.

PrajwalM2212:
I am working on it.

nishakm:
Fixed with PR #536

PR #530, opened 2020-02-02
config.yml: Add coverage

### LINKED ISSUE #407: Add test-coverage
**Problem Statement**
There must be a way to know the current coverage

**Describe the Proposal**
Currently, tern does not have a feature to know the current test coverage.
Have test coverage is important to make sure that untested code does not get 
merged.

Install and configure `coverage`

PrajwalM2212:
This PR adds coverage so that the current
test coverage can be identified.

This PR removes test_class_docker_image.py,
test_util_commands.py and test_util_metadata.py
as a docker image required to run these tests
is not available at this time.
Closes #407

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 version: 2.1
+
 # set up executor
 executors:
   ubuntu1604:
@@ -63,6 +64,15 @@ jobs:
       - setup
       - run: pip install .
       - run: tern -l report -i photon:3.0
+  test_coverage:
+    executor: ubuntu1604
+    steps:
+      - setup
+      - run: pip install coverage
+      - run: pip install .
+      - run: coverage run -m unittest discover -s tests
+      - run: coverage report
+
 
 workflows:
   version: 2
@@ -72,3 +82,5 @@ workflows:
       - commit_check
       - security
       - test_changes
+      - test_coverage
+


rnjudge:
Thanks for your PR @PrajwalM2212! Since I am not familiar with coverage, could you help me understand why only 3 out of the 12 test files were altered to add the `"""` at the beginning and end of the files?

PrajwalM2212:
@rnjudge The 3 test files that were commented out fail and have to be fixed. https://github.com/vmware/tern/issues/405

rnjudge:
> @rnjudge The 3 test files that were commented out fail at the module import level (they refer to utils.commands that does not exist) and have to be fixed. #405

I wonder if we should have a fix ready before merging this PR. Commenting out the three tests doesn't seem to me like the right approach in the long run. @nishakm thoughts?

nishakm:
@PrajwalM2212 LGTM in general. Could you please make some minor changes?
- Fix spelling in the commit message body: curret -> current
- If you could add an explanation for why the 3 tests were blanket commented out, that would be awesome

@rnjudge We need a test docker image in order to run those 3 tests and we don't run them in our normal CI tests anyway. We should have technical debt issues for each of these 3 files and link them to the blocking issue #405.

PrajwalM2212:
@nishakm @rnjudge The commit message has been amended.

nishakm:
Will merge this after we fix-up #531

rnjudge:
Currently the test coverage is at 63% (see [latest GitHub actions run](https://github.com/vmware/tern/pull/536/checks?check_run_id=453299736)). I propose that we shelf this PR and open an issue to increase the test coverage until we get to 85%. If we merge this now, all PRs will look like they are failing because of the overall coverage, which doesn't actually reflect the content of the PR. @nishakm @PrajwalM2212 thoughts?

PrajwalM2212:
@rnjudge We could also change `coverage report --fail-under 85` to just `coverage report `. This way the PR's wont fail. This will also display the report which can be used as a reference to increase the coverage. 
I am also ok with removing the coverage completely till the test coverage is increased.

nishakm:
> @rnjudge We could also change `coverage report --fail-under 85` to just `coverage report `. This way the PR's wont fail. This will also display the report which can be used as a reference to increase the coverage.

Looking at a coverage report sounds good to me for now. We can add the `--fail-under` flag later. We don't have tests for a number of things right now, but having a test docker image helps us.

rnjudge:
> > @rnjudge We could also change `coverage report --fail-under 85` to just `coverage report `. This way the PR's wont fail. This will also display the report which can be used as a reference to increase the coverage.
> 
> Looking at a coverage report sounds good to me for now. We can add the `--fail-under` flag later. We don't have tests for a number of things right now, but having a test docker image helps us.

OK. I will update the GitHub actions to remove the `--fail-under` flag for now. I also opened issue #539.

rnjudge:
Merging.

Issue #406, opened 2019-08-25
Run tests using `python3 setup.py test`


PrajwalM2212:
**Problem Statement**
The command `python3 setup.py test` is not running the tests.

**Describe the Proposal**
`python3 setup.py test` should be able to run all the tests.
Such functionality will help the developers to know if they are breaking anything with their new features.

**Steps to Implement Proposal**
1. Reorganise import statements in the test files.
2. Add a function in setup.py to load test_suite

nishakm:
@PrajwalM2212 do you want to try incorporating this?

PrajwalM2212:
Yes, Give me a couple of days.

nishakm:
Closing as we can now run tests using `tox`.

PR #541, opened 2020-02-21
tox.ini: configure and run tox

PrajwalM2212:
This PR adds tox.ini file and modifies
pull_request.yml so that github actions
can run tox.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -54,4 +54,4 @@ console_scripts =
     tern = tern.__main__:main
 
 [options.extras_require]
-dev = bandit~=1.6; prospector>=1.2; GitPython~=2.1
+dev = bandit~=1.6; prospector>=1.2; GitPython~=2.1; tox>=3.14
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,38 @@
+[tox]
+envlist = py35
+          py36
+          py37
+          commit
+          prospector
+          security
+          changes
+          coverage
+
+[testenv]
+description = Run tests
+deps = GitPython~=2.1
+
+[testenv:prospector]
+description = Run the Prospector test
+deps = prospector>=1.2
+commands = prospector
+
+[testenv:commit]
+description = Run the Commit test
+commands = python ci/test_commit_message.py
+
+[testenv:security]
+description = Run bandit
+deps = bandit~=1.6
+commands = bandit .
+
+[testenv:tests]
+description = Run unittests
+commands = python -m unittest discover -s tests
+
+[testenv:coverage]
+description = Get current coverage
+deps = coverage
+commands = coverage run -m unittest discover -s tests
+           coverage report --omit .tox/**,tests/**
+
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,4 +83,3 @@ workflows:
       - security
       - test_changes
       - test_coverage
-
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -21,14 +21,14 @@ jobs:
           c=`python ci/evaluate_docs.py`; if [ -z $c ]; then echo "No .py files to lint"; else echo $c | xargs prospector; fi
       - name: Commit Message Linting
         if: always() # run even if above step fails
-        run: python ci/test_commit_message.py    
+        run: python ci/test_commit_message.py
       - name: Security Linting
-        if: always() # run even if above step fails 
+        if: always() # run even if above step fails
         run: |
           pip install bandit~=1.6
           c=`python ci/evaluate_docs.py`; if [ -z $c ]; then echo "No .py files to lint"; else echo $c | xargs bandit; fi
       - name: Test Changes
-        if: always() # run even if above step fails 
+        if: always() # run even if above step fails
         run: python ci/test_files_touched.py
       - name: Test Coverage
         if: always() # run even if above step fails
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,6 +120,9 @@ This will install tern in development mode on windows and mac.
 4. Fix any issues bandit brings up.
 5. Test your changes.
 
+### Testing you changes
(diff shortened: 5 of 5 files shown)


PrajwalM2212:
I will be deliberately making the PR to fail to see if it is caught by tox.

PrajwalM2212:
I have got `Test Changes` and `Test Coverage` to work on tox. 3 more to go.

PrajwalM2212:
@nishakm @rnjudge 
I have added `tox`, the recommended way for testing . This took a lot of effort to complete. I even cloned the tox repo and had to read it some extent to understand how it works.

tern/analyze/common.py
```diff
@@ -164,6 +164,20 @@ def convert_to_pkg_dicts(pkg_dict):
...
+
+
+
+
+
+
+
+
+
+
```
nishakm: I think this was unintentional?

PrajwalM2212: Did that to check if prospector in tox caches it.

PrajwalM2212:
I was reading the docs and found that we can pass around environment variables. Have used that to fix the problem.

### Prospector failing example
Screen Shot 2020-02-22 at 8 48 50 PM
Screen Shot 2020-02-22 at 8 48 27 PM

### Bandit running example
Screen Shot 2020-02-22 at 8 48 40 PM

So tox is functioning correctly.

nishakm:
@PrajwalM2212 Thanks again for all the hard work! I think this ended up being more complicated than it should have :). The tests in the `ci` directory are for the project's CI and not for regular development. So we don't have to replace the CircleCI or the GHA configurations.
However, we do need tox (as you had filed in #406) to help developers before they submit the PR. For this purpose, just running `python tests/...` for each of the tests would be enough. I still don't understand how tox would behave from within a virtualenv. I'll check out the behaviour. Another place we would need tox is for deployments. In this case we will be running the whole battery of tests rather than just what the PR changed. But we can get to that part later. For now, just running all the tests in `tests` will suffice.

nishakm:
@PrajwalM2212 I tested with a simple `tox.ini` file:
```
[tox]
envlist = py35, py36, py37

[testenv]
deps =
    bandit~=1.6
    GitPython~=2.1
    prospector>=1.2
commands =
    python -m unittest discover -s tests
```
This works within the virtualenv with errors that python interpreters were not found:
```
ERROR:  py35: InterpreterNotFound: python3.5
  py36: commands succeeded
  py37: commands succeeded
```
I think this is OK.

tox.ini
```diff
@@ -0,0 +1,38 @@
+[tox]
+envlist = commit
+          prospector
+          security
+          changes
+          coverage
```
nishakm: Would it be possible to run this on multiple python environments?
```
envlist = py35, py36, py37
```

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+
+[testenv]
+description = Run tests
+basepython = python3.6
+deps = GitPython~=2.1
+passenv = FILES_TO_LINT
```
nishakm: I don't think we need to pass environment variables here.

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+
+[testenv:prospector]
+description = Run the Prospector test
+deps = prospector>=1.1
```
nishakm: prospector>=1.2

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+commands = /bin/sh -c 'if [ -z $FILES_TO_LINT ]; then echo "No .py files to lint"; else echo $FILES_TO_LINT | xargs prospector; fi'
+           /bin/sh -c 'echo $FILES_TO_LINT'
```
nishakm: Just run `prospector` here

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+
+[testenv:commit]
+description = Run the Commit test
+commands = python ci/test_commit_message.py
+
+[testenv:security]
+description = Run bandit
+deps = bandit~=1.6
+commands = /bin/sh -c 'if [ -z $FILES_TO_LINT ]; then echo "No .py files to lint"; else echo $FILES_TO_LINT | xargs bandit; fi'
```
nishakm: Just run `bandit` here.

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+
+[testenv:changes]
+description = Run changes
+commands = python ci/test_files_touched.py
+
```
nishakm: Just run tests here: `python -m unittest discover -s tests`

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+[testenv:coverage]
+description = Get current coverage
+deps = coverage
+       pytest
+commands = coverage run -m pytest tests
+           coverage report --omit .tox/**,tests/**
```
nishakm: No need to add the `pytest` dependency. You can run `coverage run -m unittest discover -s tests`.

tox.ini
nishakm: `commands = bandit .`

PrajwalM2212:
@nishakm Changes done. Tests can now be run easily. It is really nice to be able to run these tests before submitting the PR.

setup.cfg
```diff
@@ -54,4 +54,4 @@ console_scripts =
     tern = tern.__main__:main
 
 [options.extras_require]
-dev = bandit~=1.6; prospector>=1.2; GitPython~=2.1
+dev = bandit~=1.6; prospector>=1.2; GitPython~=2.1; tox
```
nishakm: On my machine, the tox version is 3.14.5 Is that the same for you? If yes, then add `tox>=3.14`

tox.ini
```diff
@@ -0,0 +1,38 @@
...
+[testenv:commit]
+description = Run the Commit test
+commands = python ci/test_commit_message.py
+
+[testenv:security]
+description = Run bandit
+deps = bandit~=1.6
+commands = bandit .
+
+[testenv:changes]
```
nishakm: Change to `tests`

PrajwalM2212:
@nishakm Done

PR #542, opened 2020-02-21
Add new class FileData

### LINKED ISSUE #476: Create new class: File
**Description**
Create a new class called File. The SPDX spec has a good model of what a File is: https://spdx.org/sites/cpstandard/files/pages/files/spdxversion2.1.pdf.

**To Do**
1. Create the new class with all the metadata using the SPDX spec as a reference
2. Create tests for the new class
3. Add tests to the ci pipeline

**Background**
This is required in order to support file level scanning. This also helps with recording file level data like checksums and extended attributes. The ImageLayer class will contain a list of File objects.

**Super Issues**
#474

nishakm:
This begins work towards adding file level data to generated
reports.

FileData represents metadata for a file found in an image layer
filesystem. It is called FileData so as to not interfere with
python's builtin 'file'. We use scancode's data model for the
initial set of properties. These include: name, path, date, file_type,
licenses, license_expressions, copyrights, packages, urls,
checksum_type and checksum. We have additionally included
version_control and version to generally point to where the file
may have come from. This is useful if the file was added from a git
repository cloned on the host machine. We have also included an
'Origins' object to record any issues that may have happened during
analysis.

We also added a mapping in the Template class called file_data to
allow templating for reports.

Appropriate tests were created and added to the ci automation.

Resolves #476

### DIFF
--- a/tern/classes/file_data.py
+++ b/tern/classes/file_data.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import datetime
+
+from tern.classes.origins import Origins
+from tern.utils.general import prop_names
+
+
+class FileData:
+    '''A file existing within a container image layer
+    If any file level analysis is done then most of the attributes can be
+    filled up. If no analysis is done, we would still have at the minimum the
+    file name and path
+    attributes:
+        name: file name
+        path: path to the file
+        date: the creation date of the file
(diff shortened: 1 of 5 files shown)


tern/classes/file_data.py
```diff
@@ -0,0 +1,142 @@
...
+        self.__checksum = checksum
+
+    def set_version(self, version_control='', version=''):
+        '''Set the version control and version of the file'''
+        # TODO: find the version given the version control system
+        self.__version_control = version_control
+        self.__version = version
+
+    def to_dict(self, template=None):
+        '''Return a dictionary version of the Package object
```
rnjudge: As discussed in slack, change to reflect FileData class :)

Issue #507, opened 2019-12-17
Extract the digest of a Docker image to pin for Dockerfile "freeze" functionality


rnjudge:
**Description**
Extract the digest of the image provided in the Dockerfile and pin it in the "frozen" Dockerfile.

**To Do**
1) This task will take some investigation as it is unclear at the time of filing this issue how we get the exact transport digest of a docker image. As far as we know, the Docker API has no way of getting the image digest if the image already exists on disk and the digest that you pull is not the same as the digest that exists when you do a docker save that you might find in the container.

2) Once (1) is determined, we will need to update the `docker_image` class to store the digest value. Currently we only store the image repo tag but we should have something to represent the digest that we pull in the Dockerfile. 

3) Add a parser in the `tern/analyze/docker/dockerfile.py` class to determine if the image digest is provided in the Dockerfile. If it is, we don't need to determine the digest in step (2) for the resulting "frozen" Dockerfile.

There is some prototyping done for this here: https://gist.github.com/joshuagl/2d30d79373f87858db9863d2a44d6150

**Super Issues**
#454

nishakm:
Fixed in 3dbd5a037d92f00dbb7f8bf0bf72404b16bc007b

PR #548, opened 2020-02-25
Extract sha digest of Docker image:tag

rnjudge:
This commit adds two functions in dockerfile.py that will extract the
sha256 digest of an image given its name and tag using the Docker SDK
for Python. All changes are in tern/analyze/docker/dockerfile.py.

1) find_image_digest() takes an image name and tag and outputs the
   associated tag@sha:digest value from Docker. There are two digests
   that associate to an image -- a) the transport digest that shows
   when you run 'docker pull <image> on your local command line and
   b) The digest that you see if you were to navigate the dockerhub
   webui repository and look for the image. find_image_digest() will
   return the digest from option (a).

2) expand_from_images() will take a dockerfile object (created using
   get_dockerfile_obj() in dockerfile.py) and update the structure
   dictionary and parent_images list with the full tag@sha:digest
   value. It uses find_image_digest() to get the associated digest
   shas. This will be used later to "pin" a base image in a
   dockerfile. This works for multistage builds as well.

Works towards #507

### DIFF
--- a/tern/analyze/docker/container.py
+++ b/tern/analyze/docker/container.py
@@ -77,23 +77,39 @@ def check_image(image_tag_string):
     logger.debug(
         "Checking if image \"%s\" is available on disk...", image_tag_string)
     try:
-        client.images.get(image_tag_string)
+        image = client.images.get(image_tag_string)
         logger.debug("Image \"%s\" found", image_tag_string)
-        return True
+        return image
     except docker.errors.ImageNotFound:
-        return False
+        return None
 
 
 def pull_image(image_tag_string):
     '''Try to pull an image from Dockerhub'''
     logger.debug("Attempting to pull image \"%s\"", image_tag_string)
     try:
-        client.images.pull(image_tag_string)
+        image = client.images.pull(image_tag_string)
         logger.debug("Image \"%s\" downloaded", image_tag_string)
-        return True
+        return image
     except docker.errors.ImageNotFound:
         logger.warning("No such image: \"%s\"", image_tag_string)
-        return False
+        return None
+
+
+def get_image(image_tag_string):
+    '''Try to get an image from Dockerhub.
+    image_tag_string: can be in image:tag or image@digest_type:digest format'''
+    check_docker_setup()
+    image = check_image(image_tag_string)
+    if image is None:
+        return pull_image(image_tag_string)
+    return image
+
+
+def get_image_digest(docker_image):
+    '''Given a docker image object return the digest information of the
+    unique image in 'image@sha_type:digest' format.'''
+    return docker_image.attrs['RepoDigests'][0]
 
 
 def build_container(dockerfile, image_tag_string):
--- a/tern/analyze/docker/dockerfile.py
+++ b/tern/analyze/docker/dockerfile.py
@@ -9,8 +9,15 @@ Dockerfile information retrieval and modification
 
 from dockerfile_parse import DockerfileParser
 import re
+import logging
 
 from tern.utils import general
+from tern.utils import constants
+from tern.analyze.docker import container
+
+# global logger
+logger = logging.getLogger(constants.logger_name)
+
 
 directives = ['FROM',
               'ARG',
@@ -116,6 +123,35 @@ def parse_from_image(dfobj):
     for image_string in dfobj.parent_images:
         image_list.append(general.parse_image_string(image_string))
     return image_list
+
+
+def expand_from_images(dfobj):
+    '''Replace all parent_image values with their full image@digest_type:digest
+    value. Update the structure dictionary with the same information.
+    dfobj: the Dockerfile object created using get_dockerfile_obj'''
+    # update parent_images
+    images = parse_from_image(dfobj)
+    for i, img in enumerate(images):
+        # don't re-pull digest if already available
+        if not img['digest_type'] and not img['digest']:
+            if not img['tag']:
+                img['tag'] = 'latest'
+            image = container.get_image(img['name'] + tag_separator +
+                                        img['tag'])
+            if image is not None:
+                dfobj.parent_images[i] = container.get_image_digest(image)
+            else:
+                logger.error("Error pinning digest to '%s'. Image not found.",
+                             dfobj.parent_images[i])
+    # update structure
+    counter = 0
+    for i, command_dict in enumerate(dfobj.structure):
+        if command_dict['instruction'] == 'FROM':
+            # Pull digest in order of parent_images
+            dfobj.structure[i]['content'] = command_dict['instruction'] + \
+                ' ' + dfobj.parent_images[counter] + '\n'
+            dfobj.structure[i]['value'] = dfobj.parent_images[counter]
+            counter = counter + 1
 
 
 def get_command_list(dockerfile_name):
--- a/tern/report/report.py
+++ b/tern/report/report.py
@@ -52,9 +52,9 @@ def setup(dockerfile=None, image_tag_string=None):
         dhelper.load_docker_commands(dockerfile)
     # check if the docker image is present
     if image_tag_string and general.check_tar(image_tag_string) is False:
-        if not container.check_image(image_tag_string):
+        if container.check_image(image_tag_string) is None:
             # if no docker image is present, try to pull it
-            if not container.pull_image(image_tag_string):
+            if container.pull_image(image_tag_string) is None:
                 logger.fatal("%s", errors.cannot_find_image.format(
                     imagetag=image_tag_string))
                 sys.exit()
@@ -88,9 +88,9 @@ def load_base_image():
     '''Create base image from dockerfile instructions and return the image'''
     base_image, dockerfile_lines = dhelper.get_dockerfile_base()
     # try to get image metadata
-    if not container.check_image(base_image.repotag):
+    if container.check_image(base_image.repotag) is None:
         # if no base image is found, give a warning and continue
-        if not container.pull_image(base_image.repotag):
+        if container.pull_image(base_image.repotag) is None:
             logger.warning("%s", errors.cannot_find_image.format(
                 imagetag=base_image.repotag))
     try:
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -143,6 +143,8 @@ def check_root():
 
 
 def check_image_string(image_str: str):
+    '''Check if the image string is in the format image:tag or
+    image@digest_type:digest format. If not, return False.'''
     tag_format = r'.+:.+'
     digest_format = r'.+@.+:.+'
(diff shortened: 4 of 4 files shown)


tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +126,38 @@ def parse_from_image(dfobj):
...
+    '''Given an image name and tag from the FROM instruction, return the
+    digest information of the unique image in 'tag@sha:digest' format.'''
+    container.check_docker_setup()
+    image_tag_string = image_name + tag_separator + image_tag
+    logger.debug("Attempting to pull image \"%s\"", image_tag_string)
+    try:
+        return container.client.images.pull(image_tag_string).attrs[
+            'RepoDigests'][0]
+    except docker.errors.ImageNotFound:
+        logger.warning("No such image: \"%s\"", image_tag_string)
```
nishakm: I wonder if this should live in `container.py` instead of `dockerfile.py`.

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +126,38 @@ def parse_from_image(dfobj):
...
+        dfobj.parent_images[i] = find_image_digest(img['name'], img['tag'])
+    # update structure
+    for i, command_dict in enumerate(dfobj.structure):
+        if command_dict['instruction'] == 'FROM':
+            dfobj.structure[i]['content'] = 'FROM ' + find_image_digest(
+                command_dict['value'].split(tag_separator, 2)[0],
+                command_dict['value'].split(tag_separator, 2)[1])
+            dfobj.structure[i]['value'] = find_image_digest(
+                command_dict['value'].split(tag_separator, 2)[0],
+                command_dict['value'].split(tag_separator, 2)[1])
```
nishakm: I would rather pass in the results from `find_image_digest`. Either that or move the whole function into `container.py`. I'm mostly thinking in terms of the dependency on the external image for testing purposes. If this gets moved to `container.py` then we can use the test image(s) on Dockerhub for all of those tests.

rnjudge:
@nishakm I was going to open another PR  adding test coverage for these but agree that this could be a good first issue. I will open it.

tern/analyze/docker/dockerfile.py
rnjudge: Pass the results of `find_image_digest` to the `expand_from_images` instead of the dfobj or in addition?

tern/analyze/docker/dockerfile.py
rnjudge: Sure, I can move it.

tern/analyze/docker/dockerfile.py
rnjudge: I also think that this should stay in `dockerfile.py`, as it is a parallel function to `expand_vars` and will only be used in the dockerfile freeze functionality.

nishakm: OK. If we need to move it somewhere else, we can do that.

tern/analyze/docker/container.py
```diff
@@ -96,6 +96,24 @@ def pull_image(image_tag_string):
...
+def get_image(image_tag_string):
+    '''Try to get an image from Dockerhub. If image is not valid
+    or unavailable return blank string.
+    image_tag_string: can be in image:tag or image@digesttype:digest format'''
+    check_docker_setup()
+    # Can only docker 'get' a local image. If not available, pull it first
+    if not check_image(image_tag_string):
+        if not pull_image(image_tag_string):
+            return ''
+    return client.images.get(image_tag_string)
```
nishakm: I was actually thinking you could modify the `check_image` function above to return the docker_image object instead of returning `True`. If no object is returned then return `None`. Same thing with the `pull_image` function (although I haven't tested whether it will return the same digest).
Also, this function would be useful in the `setup` function in `report.py`. Alternatively, you could also assume that the check for the image is already done and it is available (or the script has errored out).

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +125,27 @@ def parse_from_image(dfobj):
...
 
+def expand_from_images(dfobj):
+    '''Replace all parent_image values with their full tag@sha:digest value.
+    Update the structure dictionary with the same information.
+    dfobj: the Dockerfile object created using get_dockerfile_obj'''
+    # update parent_images
+    images = parse_from_image(dfobj)
+    for i, img in enumerate(images):
+        if general.check_image_string(img['name'] + tag_separator +
+                                      img['tag']):
```
nishakm: IIRC the dictionary returned by `parse_from_image` will have a non-empty `digest_type` and `digest` if that was the image in the Dockerfile, in which case, you don't have to get the digest. So you can just check if the dictionary has an `image` and a `tag`. If there is no `tag` then you can substitute that value with `latest` and run your `get_image` function.

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +125,27 @@ def parse_from_image(dfobj):
...
+            dfobj.parent_images[i] = container.get_image_digest(
+                container.get_image(img['name'] + tag_separator + img['tag']))
+    # update structure
+    for i, command_dict in enumerate(dfobj.structure):
+        if command_dict['instruction'] == 'FROM':
+            dfobj.structure[i]['content'] = command_dict['instruction'] + \
+                ' ' + container.get_image_digest(container.get_image(
+                    command_dict['value'])) + '\n'
```
nishakm: Why are you trying to `get_image` again?

rnjudge: The value of the digest needs to be assigned twice. But I can assign it to a variable and assign it that way if you prefer.

rnjudge: The value of the digest needs to be assigned twice and therefore needs the image object twice as input. But I can assign the image object to a variable and input it to get_image_digest that way if you prefer.

rnjudge: Additionally, this is the first time `get_image` is called in this loop -- so not sure what you mean by "again" unless you're talking about line 145-146? Or are you saying again from the loop above?

tern/analyze/docker/container.py
```diff
@@ -96,6 +96,24 @@ def pull_image(image_tag_string):
...
+
+
+def get_image_digest(docker_image):
+    '''Given a docker image object return the digest information of the
+    unique image in 'tag@sha:digest' format.'''
```
nishakm: Small nit: `image@digest_type:digest` format

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +125,27 @@ def parse_from_image(dfobj):
...
+    '''Replace all parent_image values with their full tag@sha:digest value.
+    Update the structure dictionary with the same information.
+    dfobj: the Dockerfile object created using get_dockerfile_obj'''
+    # update parent_images
+    images = parse_from_image(dfobj)
+    for i, img in enumerate(images):
+        if general.check_image_string(img['name'] + tag_separator +
+                                      img['tag']):
+            dfobj.parent_images[i] = container.get_image_digest(
+                container.get_image(img['name'] + tag_separator + img['tag']))
```
nishakm: It looks here that you are overwriting the `dfobj.parent_images` with the image in digest form. Nice!

tern/analyze/docker/dockerfile.py
nishakm: Why not use what is stored in `dfobj.parent_images`? Or if you are worried about losing association, how about making a dictionary with `{'image:tag': 'image@digest_type:digest'}` in the first loop and using it in the second loop?

rnjudge: I thought about that but since we are iterating through a structure with many instructions, the count of the second loop will be extremely difficult to keep up with the one or two images in the images dictionary.

tern/analyze/docker/container.py
```diff
@@ -77,23 +77,39 @@ def check_image(image_tag_string):
...
+
+
+def get_image(image_tag_string):
+    '''Try to get an image from Dockerhub.
+    image_tag_string: can be in image:tag or image@digesttype:digest format'''
+    check_docker_setup()
+    image = check_image(image_tag_string)
+    if image is None:
+        return pull_image(image_tag_string)
+    return image
```
nishakm: `def setup()` in `report.py` has some error checking if neither one of these work:
``` 
if not container.check_image(image_tag_string):
            # if no docker image is present, try to pull it
            if not container.pull_image(image_tag_string):
                logger.fatal("%s", errors.cannot_find_image.format(
                    imagetag=image_tag_string))
                sys.exit()
```
Have you thought about rolling this code in here and using this function in `setup()`?

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +120,31 @@ def parse_from_image(dfobj):
...
+    # update parent_images
+    images = parse_from_image(dfobj)
+    for i, img in enumerate(images):
+        # don't re-pull digest if already available
+        if not img['digest_type'] and not img['digest']:
+            if not img['tag']:
+                img['tag'] = 'latest'
+            dfobj.parent_images[i] = container.get_image_digest(
+                container.get_image(img['name'] + tag_separator +
+                                    img['tag']))
```
nishakm: So it looks like here `dfobj.parent_images` has the image digests all set. Why not use these values in the loop below? To maintain association, you can create a dictionary with the original name and the digest name.

tern/analyze/docker/container.py
nishakm: My bad - didn't read your changes to report.py

tern/analyze/docker/dockerfile.py
nishakm: Also, what happens when `get_image` returns `None` here?

rnjudge: I already made changes in the loop below to use the `dfobj.parent_images`.

I don't have a check to account for None image types... was thinking this would happen at a higher level but can add one here if you like.

nishakm: Sorry, it looks like it preserved my original review after you made the change. I would suggest checking if what is returned by `check_image` is not None and then doing the replacement and if not, log the error and don't expand.

tern/analyze/docker/dockerfile.py
```diff
@@ -118,6 +125,35 @@ def parse_from_image(dfobj):
     return image_list
 
 
+def expand_from_images(dfobj):
+    '''Replace all parent_image values with their full tag@sha:digest value.
```
nishakm: Small nit: `image@digest_type:digest`

rnjudge: Ah shoot, thought I had fixed this.

PR #549, opened 2020-02-25
container: close the client socket

nishakm:
The docker client socket is still open after running. We don't
want to leave this open. So closing it during teardown.


nishakm:
From the test coverage - the open resource warning is no longer there:
```
#!/bin/bash -eo pipefail
coverage run -m unittest discover -s tests

.....1
...1
.......................................................
----------------------------------------------------------------------
Ran 63 tests in 0.847s

OK
```

PR #558, opened 2020-02-27
classes: Added FileData object list to ImageLayer

nishakm:
We add a list of FileData objects representing a list of files
found in a container image layer. The property, and methods closely
follow the ones for the Package list. We also take into account
the addition of the list of files in the dictionary representation
with and without templates. We modify the test and test fixtures
accordingly.

### DIFF
--- a/tern/classes/image_layer.py
+++ b/tern/classes/image_layer.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2017-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
 
 from tern.classes.package import Package
+from tern.classes.file_data import FileData
 from tern.classes.origins import Origins
 from tern.utils import rootfs
 from tern.utils.general import prop_names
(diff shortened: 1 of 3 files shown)


tern/classes/image_layer.py
```diff
@@ -152,11 +162,42 @@ def remove_package(self, package_name):
...
+    def add_file(self, filedata):
+        if isinstance(filedata, FileData):
+            if filedata.path not in self.get_file_paths():
+                self.__files.append(filedata)
+        else:
+            raise TypeError('Object type is {0}, should be FileData'.format(
+                type(filedata)))
+
+    def get_file_paths(self):
+        '''Get the list of package names in this layer'''
```
PrajwalM2212: The documentation string is referring to packages instead of file paths

nishakm: Good catch. Thanks!

tern/classes/image_layer.py
```diff
@@ -30,18 +31,23 @@ class ImageLayer:
...
     methods:
         add_package: adds a package to the layer
         remove_package: removes a package from the layer
         to_dict: returns a dict representation of the instance
         get_package_names: returns a list of package names
-        gen_fs_hash: calculate the filesystem hash'''
+        gen_fs_hash: calculate the filesystem hash
+        add_file: adds a file to the layer
+        remove_file: given the file path, remove a file object from the
+        list of files in this layer'''
```
rnjudge: Did you want to add `get_file_paths` to this list?

PR #560, opened 2020-02-28
classes: Add FileData object list to package class

PrajwalM2212:
This PR adds files property to package class. It adds add_file, remove_file, get_file_paths methods. Tests have been added for the changes made. The to_dict method has also been changed to account for the new files property

Closes https://github.com/vmware/tern/issues/559

### DIFF
--- a/tern/classes/package.py
+++ b/tern/classes/package.py
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
-
+from tern.classes.file_data import FileData
 from tern.classes.notice import Notice
 from tern.classes.origins import Origins
 from tern.utils.general import prop_names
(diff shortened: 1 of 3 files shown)


tern/classes/package.py
```diff
@@ -90,11 +100,33 @@ def checksum(self):
...
         self.__checksum = checksum
 
+    def get_file_paths(self):
+        """Return a list of paths of all the files in a package"""
+        return list(map(lambda f: f.path, self.__files))
+
+    def add_file(self, file):
+        """Add a file to the list of files present in a package"""
+        if isinstance(file, FileData):
+            if file.path not in self.get_file_paths():
```
nishakm: I don't think I added this extra check in the `add_file` function in the `ImageLayer` class. Nice!

tern/classes/package.py
```diff
@@ -90,11 +100,33 @@ def checksum(self):
     def checksum(self, checksum):
         self.__checksum = checksum
 
+    def get_file_paths(self):
+        """Return a list of paths of all the files in a package"""
+        return list(map(lambda f: f.path, self.__files))
```
nishakm: I think this is a bit overkill and hurts readability, particularly if you are casting the returned object multiple times. How about using list comprehension if you want a one-liner: `return [f.path for f in self.__files]`

nishakm:
@PrajwalM2212 I asked for a small change in the code. Other than that, it looks good to me.

tern/classes/package.py
PrajwalM2212: list comprehension is way better. Thanks

PrajwalM2212:
@nishakm Changes made

Issue #506, opened 2019-12-17
Update command parsing to detect package to be installed if version is specified


rnjudge:
**Description**
Each Linux package manager specifies how to install a particular version of a package differently. For example, when installing a package on a Debian/Ubuntu system, you would run `sudo apt install package=version.release` and if you were using Alpine you would similarly run `apk add packagename=1.2.3-suffix`. If you were using a Photon/CentOS/Fedora package manager like `tdnf` or `rpm` you would use `rpm -i package-version.release`. 

This parsing works towards the Dockerfile "freeze" functionality. The reason we want to be able to detect package versions that are specified is so we can automatically report the package version in the "frozen" Dockerfile without having to do the work of looking for it in the built container. 

**To Do**
1) Do some research to compare the python shlex and getopt library to see if there is a good way to use these for this bug.
2) Add a section under each package manager in `tern/command_lib/base.yml` that denotes how the package manger specifies package versions for install (i.e. `-` or `=`)
3) Update the `parse_command` function in `tern/utils/general` to recognize and record package versions when they are present after an install command.

**Super Issues**
#454

nishakm:
This should also fix #2 - a really old bug :)

nishakm:
Fixed in a4c107f447621defecfc227d31d043e2c3a91326

PR #570, opened 2020-03-05
Parse package name from version

### LINKED ISSUE #2: Unix command parser does not parse version or arch specified in the package name
For example: apt-get install package:arch=version

If the package has a version then record it in the returning dict as 'version': version. If the package has an arch, record it in the returning dict as 'arch':

rnjudge:
This commit adds the following functionality to the following files as
well as updating the copyright years in them.

tern/classes/command.py
   Adds a get_pkg_name() function to the command class. This function
   takes a word from a dockerfile command and separator list to
   separate the packge from its version by. For all but one separator,
   the function will use regex to split the pacakge from its version.
   In the case of the '-' separator, the function will loop through
   the characters in the word to determine package version. In either
   case, it will return the package name. This will be helpful in
   the upcoming dockerfile freeze functionality when we want to use
   the package manager to find the version of a package installed,
   regardless of the package that is specified in the dockerfile.

tests/test_class_command.py
   Add a test to test get_pkg_name() from the command class.

tern/command_lib/base.yml
   Add a package manager separator character list that would be used
   in a dockerfile to to denote a specific package version to install,
   for example, in 'apk add package=version', '=' is the separator
   character. Also add a 'pinning_separator' string associated with
   each package manager that will be used to "freeze" pacakge
   versions in the pinned dockerfile returned to the user.

tern/command_lib/snippets.yml
   Add 'rpm' snippet as a general command to retrieve package
   information.

Resolves #2
Works towards #506

### DIFF
--- a/tern/command_lib/base.yml
+++ b/tern/command_lib/base.yml
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2019 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2017-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 #
 # Known package listings for base images
@@ -6,6 +6,8 @@
 # <package manager>:
 #   pkg_format: <the package format used by the package manager>
 #   shell: <the shell executable> (needed if entering code snippets)
+#   pkg_separators: Potential characters to separate package name from version. Some package managers will have more than one.
+#   pinning_separator: Character to denote package version in pinned Dockerfile (i.e. apk add package=version)
 #   names: <a list of package names>
 #     invoke: (if this is a script to invoke include this)
 #       1:
@@ -22,6 +24,9 @@ tdnf:
   path:
     - 'usr/bin/tdnf' # don't put forward slash here as os.path.join will think it is the root directory on the host
   shell: '/usr/bin/bash'
+  pkg_separators:
+    - '-'
+  pinning_separator: '-'
   names:
     invoke:
       1:
@@ -65,6 +70,9 @@ dpkg:
   path:
     - 'usr/bin/dpkg'
   shell: '/bin/bash'
+  pkg_separators:
+    - '='
+  pinning_separator: '='
   # either enter the package names as a list or a snippet to invoke either within or outside the container
   names:
     invoke:
@@ -98,6 +106,9 @@ apk:
   path:
     - 'sbin/apk'
   shell: '/bin/sh'
+  pkg_separators:
+    - '='
+  pinning_separator: '='
   names:
     invoke:
       1:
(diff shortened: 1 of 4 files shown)


tern/command_lib/base.yml
```diff
@@ -22,6 +23,8 @@ tdnf:
   path:
     - 'usr/bin/tdnf' # don't put forward slash here as os.path.join will think it is the root directory on the host
   shell: '/usr/bin/bash'
+  pkg_delimiter:
+    - '-'
```
nishakm: Is your expectation that there will be more than 1 package delimiter?

tern/command_lib/snippets.yml
```diff
@@ -54,4 +54,7 @@ dnf:
     - 'check-update'
     - 'clean'
   packages: 'rpm'
-
+rpm:
+   install: '-i'
+   remove: '-e'
+   packages: 'rpm'
```
nishakm: Is this change part of this change set? It seems to me this adds some extra package manager enabling.

tern/command_lib/base.yml
rnjudge: Yes, for example in the case of `pip` you can use `>=` or `<=` or various combinations involving `~`.

tern/command_lib/snippets.yml
rnjudge: It is tangentially related to how rpm specifies the package delimiter using `-`.

tern/command_lib/base.yml
rnjudge: But now I am thinking that I may want to pass in a list to `get_pkg_name` instead of just a single string in the case of multiple delimiters.

nishakm: I wonder if we need to make the distinction between what delimiter we use for pinning to a version vs what delimiters get used in the Dockerfile. I would think that the pinning delimiter used would be a string but for parsing out the package name from the package name + version string that might be taken care of by a regex.

nishakm: For example: `RUN pip install -y tern>=1.0` in a Dockerfile will be transcribed as `RUN pip install -y tern==1.0.1`. So I guess `pinning_separator` for `pip` is `==` but the delimiters can be any combination of `>=~^`.

rnjudge: Ah! Ok, I see, that makes sense. I think we can put `pinning_separator` and `pkg_delimiters` both in base.yml and have but have `pkg_delimiters` be list of potential values and `pinning_separator` just be the one string? 

If that is the case, `get_pkg_name` would still need to take a list of delimiters to account for in finding the package name. Somewhere else in the code will take care of swapping the `pkg_delimiter` with the `pinning_separator` for the frozen Dockerfile. Is that similar to what you're thinking as well?

Issue #509, opened 2019-12-17
Record git project name and sha in "pinned" Dockerfile comment


rnjudge:
**Description**
When parsing the commands in a Dockerfile, if the parser comes across the `ADD` directive, record the git project name and sha in a comment after the command. i.e.

```ADD tar /folder --> ADD tar /folder # git project name and sha if any```

**To Do**
Update Tern's parser to parse `ADD` arguments. Determine if the argument after `ADD` (i.e. what is being copied in) is a folder and if there's a any git information. If not, make a comment indicating that no git information was found.

**Super Issues**
#454

ForgetMe17:
I want to fix this issue. Here is my plan. We can add a function called `add_git_info()` in `dockerfile.py`, which recieve a line with ADD instruction. First it will check if the argument is a folder. For a folder, it will check if it contains a `.git` folder. Then it will extract git project name from `config` file in `.git`. If it is a git respo cloned, `config` file will have a `remote` section containing `url`, we can get the project name here. For sha info, we can know where the HEAD is from file `HEAD` (i.e. ref: refs/heads/master), then we can find the sha in ref/heads/master. Finally we can append the info at the end of line.

For testcases, there are no ADD instructions in the dockfiles at `test`. Maybe we should add a test dockerfile for this function.

nishakm:
@rnjudge does the above plan sound good to you?

rnjudge:
@ForgetMe17 I like it! Thank you for taking this on.

rnjudge:
@ForgetMe17  I'm also thinking about how we would get the actual folder -- since the `dfobj.structure` ADD command line will just be a string containing the folder name. @nishakm Would we assume that the directory is local on the users working directory and find it that way?

nishakm:
> @ForgetMe17 I'm also thinking about how we would get the actual folder -- since the `dfobj.structure` ADD command line will just be a string containing the folder name. @nishakm Would we assume that the directory is local on the users working directory and find it that way?

This one is dicey - the folder where the Dockerfile exists may not be in a local git repository. So we first have to check if it is. You can't do that by just looking for a `.git` folder because the Dockerfile may exist in a sub folder. You can try to run a git command and see if it fails. That way you know if you are in a git repo or not. From here on, rather than look into the refs in the `.git` folder, you're better off running git commands to get the url, remote branch and commit. Not that these are guaranteed pins but it's the best we can do.

ForgetMe17:
[...]

PR #571, opened 2020-03-05
Record git project name and sha

ForgetMe17:
This pr adds functions to the dockerfile parser. If the parser
comes across the ADD directive, record the git projeect name and
sha in a comment after the command.

This pr modifies utils/general.py and analyze/docker/dockerfile.py.

In dockerfile.py, function find_git_info recieve a line of ADD
command and the path to the parsed dockerfile. Function
expand_add_command receives a dfobj and iter it with find_git_info.

In general.py, function check_git_folder is used to find git info
in a folder and funtion check_git_tar in a tar file. Function
check_git_src looks into the src_path, and check the file type.

Work towards #509.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -9,6 +9,7 @@ Common functions
 
 import logging
 import os
+import subprocess  # nosec
 
 from tern.classes.package import Package
 from tern.classes.notice import Notice
@@ -452,3 +453,59 @@ def get_os_style(image_layer, binary):
                     package_manager=binary,
                     package_format=image_layer.pkg_format,
                     os_list=image_layer.os_guess), 'info'))
+
+
+def check_git_src(dockerfile_path):
+    '''Given the src_path and the dockerfile path, return the git
+    repository name and sha information in the format of string.
+    Currently we only consider the following situation:
+    - target_git_project
+        - dir1
+        - dir2/dockerfile
+    So we only use dockerfile_path to find the git repo info.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # locate the top level directory
+    path_to_toplevel = get_git_toplevel(dockerfile_folder_path)
+    # get the path of the target folder or file
+    logger.debug('looking into path: %s for git repo.', path_to_toplevel)
+    comment_line = ''
+    if path_to_toplevel:
+        sha_info = get_git_sha(path_to_toplevel)
+        # if path_to_toplevel exists, name_info should be the folder name
+        name_info = os.path.basename(path_to_toplevel)
+        comment_line = ('git project name: ' + name_info +
+                        ', HEAD sha: ' + sha_info)
+    else:
+        comment_line = 'Not a git repository'
+    return comment_line
+
+
+def get_git_sha(path_to_toplevel):
+    '''Given a absolute path to a git repository, return the HEAD sha.'''
+    command = ['git', 'rev-parse', 'HEAD']
+    sha_info = '(not found)'
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=path_to_toplevel)
+        if isinstance(output, bytes):
+            sha_info = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.debug("Cannot find git repo sha, toplevel path is %s",
+                     path_to_toplevel)
+    return sha_info
+
+
+def get_git_toplevel(path):
+    '''Give a path, return the absolute path to the top level directory if it
+    is in a git repository. Empty string will be returned if not.'''
+    command = ['git', 'rev-parse', '--show-toplevel']
+    path_to_toplevel = ''
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=path)
+        if isinstance(output, bytes):
+            path_to_toplevel = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.debug("Cannot find git repo toplevel, path is %s", path)
+    return path_to_toplevel
--- a/tern/analyze/docker/dockerfile.py
+++ b/tern/analyze/docker/dockerfile.py
@@ -14,10 +14,10 @@ import logging
 from tern.utils import general
 from tern.utils import constants
 from tern.analyze.docker import container
+from tern.analyze import common
 
 # global logger
 logger = logging.getLogger(constants.logger_name)
-
 
 directives = ['FROM',
               'ARG',
@@ -46,6 +46,7 @@ tag_separator = ':'
 class Dockerfile():
     ''' This class is used as a wrapper to store dockerfile information
     retrieved from the parser.'''
+
     def __init__(self):
         self.structure = None
         self.envs = None
@@ -284,3 +285,43 @@ def get_base_image_tag(base_instructions):
     if len(image_tag_list) == 1:
         image_tag_list.append('')
     return tuple(image_tag_list)
+
+
+def find_git_info(line, dockerfile_path):
+    '''Given a line of ADD command and the path of dockerfile,
+    return the information(string format) on the git project name and sha
+    if the dockerfile is included in a git repository.
+    ADD command has a general format:
+    ADD [--chown=<user>:<group>] <src> <dst>
+    Currently we parse the <src>, but not use it.
+    '''
+    args = line.split(' ')
+    src_path = ''
+    # check if --chown exists
+    if args[1].startswith('--chown'):
+        # check if the line is valid
+        if len(args) < 4:
+            logger.error('Invalid ADD command line')
+        src_path = args[2]
+    else:
+        # the line has no --chown option
+        if len(args) < 3:
+            logger.error('Invalid ADD command line')
+        src_path = args[1]
+    # log the parsed src_path
+    logger.debug('Parsed src_path is %s', src_path)
+    # get the git project info
+    comment_line = common.check_git_src(dockerfile_path)
+    return comment_line
+
+
+def expand_add_command(dfobj):
+    dockerfile_path = dfobj.filepath
+    for i, command_dict in enumerate(dfobj.structure):
+        if command_dict['instruction'] == 'ADD':
+            comment_line = find_git_info(command_dict['content'],
+                                         dockerfile_path)
+            dfobj.structure[i]['content'] = command_dict['content'].strip('\n')\
+                + ' # ' + comment_line + '\n'
+            dfobj.structure[i]['value'] = command_dict['value']\
+                + ' # ' + comment_line
--- a/tests/test_analyze_docker_dockerfile.py
+++ b/tests/test_analyze_docker_dockerfile.py
@@ -15,11 +15,14 @@ class TestAnalyzeDockerDockerfile(unittest.TestCase):
         self.buildpackpinned = 'tests/dockerfiles/buildpack_deps_jessie_pinned'
         self.golang = 'tests/dockerfiles/golang_1.13_stretch'
         self.buildpackarg = 'tests/dockerfiles/buildpack_deps_jessie_arg'
+        self.pin_add = ('tests/dockerfiles/pin_add_command_test/'
+                        'pin_add_command_test_dockerfile')
 
     def tearDown(self):
         del self.buildpack
         del self.golang
         del self.buildpackarg
+        del self.pin_add
 
     def testDockerfileObject(self):
         dfobj = dockerfile.Dockerfile()
@@ -112,6 +115,30 @@ class TestAnalyzeDockerDockerfile(unittest.TestCase):
         struct = dfobj.structure[1]
         self.assertEqual(struct['value'], replace_value)
         self.assertEqual(struct['content'], replace_content)
+
+    # test find the git project name
+    def testFindGitInfo(self):
+        sample_line = 'ADD plain_file /tmp'
+        comment_line = dockerfile.find_git_info(sample_line, self.pin_add)
+        project_name = comment_line.split(',')[0]
+        self.assertIn(project_name, ['git project name: project',
+                                     'git project name: tern'])
+
+    # test adding the comments
+    def testExpandAddCommand(self):
+        dfobj = dockerfile.get_dockerfile_obj(self.pin_add)
+        self.assertFalse(dfobj.is_none())
+        dockerfile.expand_add_command(dfobj)
+        content = dfobj.structure[1]['content']
+        value = dfobj.structure[1]['value']
+        content_name = content.split(',')[0]
+        value_name = value.split(',')[0]
+        self.assertIn(content_name,
+                      ['ADD plain_file /tmp # git project name: project',
+                       'ADD plain_file /tmp # git project name: tern'])
+        self.assertIn(value_name,
+                      ['plain_file /tmp # git project name: project',
+                       'plain_file /tmp # git project name: tern'])
 
 
 if __name__ == '__main__':
--- a/tests/dockerfiles/pin_add_command_test/pin_add_command_test_dockerfile
+++ b/tests/dockerfiles/pin_add_command_test/pin_add_command_test_dockerfile
@@ -1,0 +1,2 @@
+FROM ubuntu
+ADD plain_file /tmp
--- a/tests/dockerfiles/pin_add_command_test/plain_file
+++ b/tests/dockerfiles/pin_add_command_test/plain_file
@@ -1,0 +1,1 @@
+This is a test file.


ForgetMe17:
I have run some local tests. i.e.
```
FROM ubuntu
ADD plain_file /tmp
ADD test_git_repo /tmp
ADD test_git_repo.tar /tmp
```
i got the output 
```
FROM ubuntu
ADD plain_file /tmp # Not a folder or tar file
ADD test_git_repo /tmp # git project name: (not found) sha info: 55cbf5cbc8b201a99ac63f70f6f6f0ff1d1ae630
ADD test_git_repo.tar /tmp # git project name: (not found) sha info: 55cbf5cbc8b201a99ac63f70f6f6f0ff1d1ae630
```
The dockerfile and git repo are in the sample folder.
I am now working on  adding some tests for these functions.

tern/analyze/docker/dockerfile.py
```diff
@@ -265,3 +265,38 @@ def get_base_image_tag(base_instructions):
...
+    ADD command has a general format:
+    ADD [--chown=<user>:<group>] <src> <dst>
+    '''
+    args = line.split(' ')
+    src_path = ''
+    # check if --chown exists
+    if args[1].startswith('--chown'):
+        # check if the line is valid
+        if len(args) < 4:
+            raise ValueError('Invalid ADD command line')
```
rnjudge: Instead of raising a `ValueError` can you log the error? If we raise the error it will cause Tern to quit and ideally we would continue with the rest of the potential lines in the Dockerfile. Something like:

```
logger.error("Invalid ADD command line: %s", line)
```

tern/analyze/docker/dockerfile.py
```diff
@@ -265,3 +265,38 @@ def get_base_image_tag(base_instructions):
...
+        src_path = args[2]
+    else:
+        # the line has no --chown option
+        if len(args) < 3:
+            raise ValueError('Invalid ADD command line')
```
rnjudge: Same comment as above. Log the error instead of raising it.

tern/analyze/docker/dockerfile.py
```diff
@@ -265,3 +265,38 @@ def get_base_image_tag(base_instructions):
...
+        src_path = args[1]
+    # get the absolute src_path
+    comment_line = general.check_git_src(dockerfile_path, src_path)
+    return comment_line
+
+
+def expand_add_command(dfobj):
+    dockerfile_path = dfobj.filepath
```
rnjudge: I'm not sure if `dfobj.filepath` will necessarily be an absolute path. @nishakm is it possible that this path could be relative? From what I can tell, `dfobj.filepath` gets assigned to whatever the user specifies as the Dockerfile, which may not be absolute.

ForgetMe17: I use open() and tarfile.open() to read the src file of ADD command. So once we make sure the relative path is in the sys.path, we can also get the src file.

tern/utils/general.py
```diff
@@ -116,6 +117,102 @@ def get_git_rev_or_version():
...
+                name_info = '(not found)'
+            comment_line = 'git project name: ' + name_info +\
+                           ' sha info: ' + sha_info
+        else:
+            comment_line = 'Not a git repository'
+    return comment_line
+
+
+def check_git_tar(path_to_tar):
+    '''Given the path to the tar file, return the name and sha of repository.
```
rnjudge: small nit, but might want to add in this description that if the tarfile is **not** a git repository, this function will also tell you that.

tern/utils/general.py
```diff
@@ -116,6 +117,102 @@ def get_git_rev_or_version():
...
+
+def check_git_src(dockerfile_path, src_path):
+    '''Given the src_path and the dockerfile path, return the git
+    repository name and sha information in the format of string.
+    This function check the type of src_path and use check_git_tar and
+    check_git_folder to handle the src_path'''
+    # get the folder path
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # get the absolute path of the target folder or tar file
+    src_path = os.path.join(dockerfile_folder_path, src_path)
```
rnjudge: What if the potential `git` directory, specified by `src_path`, is not in the same directory as the Dockerfile folder path? Or what if  the Dockerfile is in a few subfolders below? @nishakm not sure how hard we want to look for the source that is being copied in. It's also common to see the source being copied in as just `.` which could be multiple tars or directories. We should probably account for this in some way even if we don't filter through all the directories for potential git information.

rnjudge:
> I have run some local tests. i.e.
> 
> ```
> FROM ubuntu
> ADD plain_file /tmp
> ADD test_git_repo /tmp
> ADD test_git_repo.tar /tmp
> ```
> 
> i got the output
> 
> ```
> FROM ubuntu
> ADD plain_file /tmp # Not a folder or tar file
> ADD test_git_repo /tmp # git project name: (not found) sha info: 55cbf5cbc8b201a99ac63f70f6f6f0ff1d1ae630
> ADD test_git_repo.tar /tmp # git project name: (not found) sha info: 55cbf5cbc8b201a99ac63f70f6f6f0ff1d1ae630
> ```
> 
> The dockerfile and git repo are in the same folder.
> I am now working on adding some tests for these functions.

@ForgetMe17 what is the test_git_repo that you have in your test dockerfile? Could you also test with a git repository where the project name is available?

@nishakm Is the sha info useful if we don't know the name of the git project? Also, dumb question but how is this possible? This seems confusing to me but I am also no expert at what .git metadata provides.

ForgetMe17:
I created a local git repository, not cloning from github. So the repo does not have a remote url where I get the name of project.

tern/utils/general.py
ForgetMe17: In general the src file path is the relative path to the Dockerfile path. I tested that it works with the git repo in the subfolders. If the dockerfile is in a few subfolders below, the given dockerfile_path is changed. What we do here is looking around the folder containing dockerfile and its subfolders. So we need to locate the dockerfile in this situation. I am wondering if I have missed some situations.

rnjudge: >> In general the src file path is the relative path to the Dockerfile path

This is true in most cases, but we need to account for when it is not true. People do weird things with Dockerfiles :) 

>> If the dockerfile is in a few subfolders below, the given dockerfile_path is changed.

If the Dockerfile is a few subfolders below the git folder then wouldn't `src_path = os.path.join(dockerfile_folder_path, src_path)` give you `dir/dockerfile_folder_below/gitdirectory` even though the git directory could be a directory above in `dir`?

rnjudge:
@nishakm I am thinking we may also want to move `find_git_info`, `check_git_folder`, `check_git_tar` and `check_git_src` to `tern/analyze/docker/common.py` instead of leaving them in `dockerfile.py` and just leave the `expand_add_command` in dockerfile.py. Thoughts?

ForgetMe17:
In the commit `87b35d6`, I add tests for find_git_info and expand_add_command, and move `check_git_folder`, `check_git_tar` and `check_git_src` to `tern/analyze/docker/common.py`.

I test the plain file, tar file and git folder. For git folder, since a git repo cannot be uploaded unless as a submodule. So in the online ci test, the result for the git folder is `Not a git repository`.

In the test, the path of dockerfile is relative and the src_path is also relative. For the situations where the git repo is a few folders above the dockerfile, i think in this case the src_path in the ADD command should be like `../../git_repo`. I donnot know if it is valid in the dockerfile(probably not i think), but this pr does not work with this situation since the `os.path.join` will append `../../git_repo` to the dockerfile_path. So maybe i will work on locating src_file in this situation.

tern/analyze/docker/dockerfile.py
nishakm: @rnjudge No it's not an absolute path and it doesn't need to be that. We just need to find the git repo and commit HEAD. I don't think we need to read what is in the tar file.

tern/analyze/common.py
```diff
@@ -452,3 +454,100 @@ def get_os_style(image_layer, binary):
...
+            else:
+                name_info = '(not found)'
+            comment_line = 'git project name: ' + name_info +\
+                           ' sha info: ' + sha_info
+        else:
+            comment_line = 'Not a git repository'
+    return comment_line
+
+
+def check_git_tar(path_to_tar):
```
nishakm: I don't think we need to look into the tarball that gets included into the docker image. It is already in there by the time we analyze it. I think this function can be deleted.

tests/test_analyze_docker_dockerfile.py
```diff
@@ -102,6 +105,43 @@ def testParseFromImage(self):
...
+                     'content': ('ADD test_git_repo /tmp # Not a git '
+                                    'repository\n'),
+                     'value': 'test_git_repo /tmp'},
+                    {'instruction': 'ADD',
+                     'startline': 3,
+                     'endline': 3,
+                     'content': ('ADD test_git_repo.tar /tmp # git project '
+                                 'name: test_repo.git sha info: 55cbf5cbc8b'
+                                 '201a99ac63f70f6f6f0ff1d1ae630\n'),
+                     'value': 'test_git_repo.tar /tmp'}]
```
nishakm: I don't think this functionality is testable, because the git sha of the tern repo will change every time a PR is merged. We'll just have to test the parts that adds the comment and the function that gets the git repository name (if those two things are in different functions).

tern/analyze/common.py
```diff
@@ -452,3 +454,100 @@ def get_os_style(image_layer, binary):
...
+            # find sha information
+            sha_info = ''
+            with open(dot_git_folder_path + 'HEAD', 'r') as head_file:
+                # .git/HEAD contains current head ref
+                ref_info = head_file.readline()
+                ref_relative_path = ref_info.split(' ', 1)[1].strip('\n')
+                sha_path = dot_git_folder_path + ref_relative_path
+                # get sha info from ref
+                with open(sha_path, 'r') as sha_file:
+                    sha_info = sha_file.readline().strip('\n')
```
rnjudge: You can get the sha info in one line using `git rev-parse HEAD`. Take a look at `get_git_rev_or_version()` function in  tern/tern/utils/general.py for an example on how to do this. You may also want to break this into a mini-function called something like `get_git_sha(path)` for clarity.

tern/analyze/common.py
```diff
@@ -452,3 +454,100 @@ def get_os_style(image_layer, binary):
...
                     os_list=image_layer.os_guess), 'info'))
+
+
+def check_git_folder(path_to_folder):
+    '''Given a path to git folder, return the name and sha of repository.
+    This function looks into .git folder.'''
+    comment_line = ''
+    if os.path.isdir(path_to_folder):
+        # find .git folder
+        dot_git_folder_path = os.path.join(path_to_folder, '.git/')
```
rnjudge: You can find the top level directory (that will container .git inside it) with `git rev-parse --show-toplevel` command.

tern/analyze/common.py
```diff
@@ -452,3 +454,100 @@ def get_os_style(image_layer, binary):
...
+                    sha_info = sha_file.readline().strip('\n')
+            # find project name
+            name_info = ''
+            conf = configparser.ConfigParser()
+            conf.read(dot_git_folder_path + 'config', encoding='utf-8')
+            # use remote url for git project name
+            if conf.has_section('remote "origin"'):
+                # url is the first item
+                project_url = conf.get('remote "origin"', 'url')
+                name_info = project_url.split('/')[-1]
```
rnjudge: This is tricky because while most remotes will be named `origin`, it is possible to rename `origin` to something else. We also want to get the complete repo name (i.e. github.com/tern.git)

Assuming you are in a git repository, `basename `git rev-parse --show-toplevel` will give you just the name of the actual project. Finding the url will be a little harder.

tern/analyze/common.py
ForgetMe17: Thanks for your suggestion!

ForgetMe17:
commit `648427f`, Using command line and modified tests
In `tern/analyze/common.py`:
1.Delete function check_git_tar and check_git_folder.
2.Add get_git_toplevel which uses git command to find the toplevel directory.
3.Add get_git_project_name and get_git_sha to get the git repository information which uses the path_to_toplevel given by the get_git_toplevel.

Modified tests in the `tests/test_analyze_docker_dockerfile.py`

In the test, i am using the dockerfile in `tests/dockerfiles/pin_add_command_test` and the ADD command is `ADD plain_file /tmp`. The `plain_file` is also in `tests/dockerfiles/pin_add_command_test`.  `plain_file` is not a git repo, but it is in our `tern` git repository, so the project name we get is `tern`, which is the toplevel folder name of the git repository. But in the circleci test, i find that the project name is `project`. I think it is becasue in the circleci test environment, the toplevel folder name is change to `project`. In the Pull Request Lint and Test of the Github, the toplevel folder name is 'tern'. So in the test, i used AssertIn to match the name between the `tern` and `project`.

And thanks for your suggestions which really helps me a lot!

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
+        if isinstance(output, bytes):
+            name_info = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.error("Cannot find git repo name, toplevel path is %s",
+                     path_to_toplevel)
+    return name_info
+
+
+def get_git_toplevel(src_path):
+    '''Give a path, return true if it is in a git repository and return the
```
rnjudge: Small nit: This description says the function returns true and also returns the absolute path to the top level directory when only the latter is true :)

ForgetMe17:
I squashed the commits into one commit.

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
+    is in a git repository. Empty string will be returned if not.'''
+    command = ['git', 'rev-parse', '--show-toplevel']
+    path_to_toplevel = ''
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=src_path)
+        if isinstance(output, bytes):
+            path_to_toplevel = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.error("Cannot find git repo, src path is %s", src_path)
```
nishakm: Can you use `logger.debug` instead?

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
+    '''Given a absolute path to a git repository, return the HEAD sha.'''
+    command = ['git', 'rev-parse', 'HEAD']
+    sha_info = '(not found)'
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=path_to_toplevel)
+        if isinstance(output, bytes):
+            sha_info = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.error("Cannot find git repo sha, toplevel path is %s",
```
nishakm: Can you use `logger.debug` instead?

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
+                     path_to_toplevel)
+    return sha_info
+
+
+def get_git_projcet_name(path_to_toplevel):
```
nishakm: Change to `get_git_project_name`

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
+    '''Given a absolute path to a git repository, return the project name.'''
+    command = ['basename', path_to_toplevel]
```
nishakm: You can use `os.path.basename` instead of invoking a shell command.

tern/analyze/common.py
```diff
@@ -452,3 +453,76 @@ def get_os_style(image_layer, binary):
...
                     os_list=image_layer.os_guess), 'info'))
+
+
+def check_git_src(dockerfile_path, src_path):
+    '''Given the src_path and the dockerfile path, return the git
+    repository name and sha information in the format of string.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # get the path of the target folder or file
+    src_path = os.path.join(dockerfile_folder_path, src_path)
```
rnjudge: This `src_path` is incorrect here. You will want to use `get_git_toplevel(dockerfile_path)` for the `src_path` here and then you don't need to grab it later in line 472.

tern/analyze/docker/dockerfile.py
```diff
@@ -265,3 +266,38 @@ def get_base_image_tag(base_instructions):
...
+    for i, command_dict in enumerate(dfobj.structure):
+        if command_dict['instruction'] == 'ADD':
+            comment_line = find_git_info(command_dict['content'],
+                                         dockerfile_path)
+            dfobj.structure[i]['content'] = command_dict['content'].strip('\n')\
+                + ' # ' + comment_line + '\n'
```
rnjudge: You are updating the `dfobj.structure[i]['content']` correctly here, but we also need to update `dfobj.structure[i]['value']` in the structure dictionary to match. Note that the `value` is the same as the associated `content` value minus the Dockerfile command and newline character. Example of the dictionary from a run command:

```{'instruction': 'RUN', 'startline': 12, 'endline': 12, 'content': 'RUN pip3 install --upgrade pip && pip3 install tern\n', 'value': 'pip3 install --upgrade pip && pip3 install tern'}```

File: tern/analyze/common.py (hunk 16 above)
Comment (ForgetMe17) [reply tern/analyze/common.py]: I am considering this situation:
- dockerfile_folder
    - Dockerfile
    - git_project_folder

ADD command is like `ADD git_project_folder /tmp`. 

In this situation, args given to `check_git_src` equals: 
`src_path = git_project_folder`
`dockerfile_path = path_to_dockerfile_folder/dockerfile_folder/Dockerfile`.

So I use `os.path.join` to get the path `path_to_dockerfile_folder/dockerfile_folder/git_project_folder`.
If `src_path = get_git_toplevel(dockerfile_folder_path)`, we will get the toplevel directory containing this dockerfile instead of looking into the src given in ADD command first.

In another way, if ADD command is copying a folder which contains a git project(assume the name is A) into the image, and the dockerfile itself is in a git repo(assume the name is B), should we record the A or B?
- B
    - subfolder
        - Dockerfile
        - A(src of the ADD command)

Comment (rnjudge) [reply tern/analyze/common.py]: > I am considering this situation:
> 
>     * dockerfile_folder
>       
>       * Dockerfile
>       * git_project_folder
> 

Your code works for this situation, but this is a corner case. See below. 

> In another way, if ADD command is copying a folder which contains a git project(assume the name is A) into the image, and the dockerfile itself is in a git repo(assume the name is B), should we record the A or B?
> 
>     * B
>       
>       * subfolder
>         
>         * Dockerfile
>         * A(src of the ADD command)

We want to record the git information from the repo that is being copied, so we would want to record A. However, I think we should focus this PR for now on the following use case and we can add more later.

The most common use case we will see for this is:
* git_project_folder
    * directroy1
    * directory2/Dockerfile

ADD git_project_folder /tmp

which means that with your code now:
`dockerfile_folder_path = ~/git_project_folder/directory2`
`src_path = git_project_folder`
and thus, `src_path = ~/git_project_folder/directory2/git_project_folder`. If you check to see if this directory exists, it will return false and the comment will say "Unable to find the src file or filter".

Because the above is the most common case, I think we should just use `get_top_dir` of the dockerfile location and worry about the other corner cases later. We have a release next week and ideally, this functionality will be included. What do you think?

I also want to apologize, because I think I may have confused you a bit considering all of the situations. I am sorry for that and want you to know that you are doing an excellent job of contributing and making meaningful changes. Your efforts do not go unnoticed!

Comment (ForgetMe17) [reply tern/analyze/common.py]: Thinks for your explanation! Maybe I should figure out the use case clearly before coding. I am enjoying contributing and thanks for encouragement! I will work on it now!

File: tern/analyze/common.py [hunk 18]
```diff
@@ -452,3 +453,61 @@ def get_os_style(image_layer, binary):
...
+        - dir2/dockerfile
+    So we only use dockerfile_path to find the git repo info.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # locate the top level directory
+    path_to_toplevel = get_git_toplevel(dockerfile_folder_path)
+    # get the path of the target folder or file
+    logger.debug('looking into path: %s for git repo.', path_to_toplevel)
+    # add this line to avoid linting check
+    logger.debug('src_path of ADD command: %s.', src_path)
```
ForgetMe17: I find the `src_path` of ADD command is not used when looking for the toplevel folder. So I add this line to avoid linting check. `src_path` may be useful when considering other corner situations.

tern/analyze/common.py
```diff
@@ -452,3 +453,61 @@ def get_os_style(image_layer, binary):
...
+    repository name and sha information in the format of string.
+    Currently we only consider the following situation:
+    - target_git_project
+        - dir1
+        - dir2/dockerfile
+    So we only use dockerfile_path to find the git repo info.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # locate the top level directory
+    path_to_toplevel = get_git_toplevel(dockerfile_folder_path)
```
rnjudge: What happens if we just do `get_git_toplevel(dockerfile_path)`? Shouldn't we get the same thing? If so, this would eliminate having to create a `dockerfile_folder_path` variable that is only used once.

ForgetMe17: I set the `cwd(current wording diretory)` to be `dockerfile_folder_path` in `subprocess.check_output` when looking for toplevel dir. `dockerfile_path` points to a file, not a diretory which cannot be set as `cwd`.

tern/analyze/common.py
```diff
@@ -452,3 +453,61 @@ def get_os_style(image_layer, binary):
                     package_manager=binary,
                     package_format=image_layer.pkg_format,
                     os_list=image_layer.os_guess), 'info'))
+
+
+def check_git_src(dockerfile_path, src_path):
```
nishakm: Seems to me you don't need `src_path` here. You can get the path to the git repository from the directory where the Dockerfile is.

tern/analyze/common.py
```diff
@@ -452,3 +453,61 @@ def get_os_style(image_layer, binary):
...
+        - dir1
+        - dir2/dockerfile
+    So we only use dockerfile_path to find the git repo info.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    # locate the top level directory
+    path_to_toplevel = get_git_toplevel(dockerfile_folder_path)
+    # get the path of the target folder or file
+    logger.debug('looking into path: %s for git repo.', path_to_toplevel)
+    # add this line to avoid linting check
```
nishakm: The reason why the linter was complaining is because you don't use `src_path` anywhere.

tern/analyze/common.py
nishakm: So if you remove `src_path` these two lines can be removed as well.

tern/analyze/common.py
```diff
@@ -452,3 +453,61 @@ def get_os_style(image_layer, binary):
...
+    logger.debug('src_path of ADD command: %s.', src_path)
+    comment_line = ''
+    if os.path.exists(path_to_toplevel):
```
nishakm: Rather than check if the path exists you can check if `path_to_toplevel` is an empty string just by saying `if path_to_toplevel:`

tern/analyze/common.py
ForgetMe17: Do we still need to parse `ADD` command to get the `src_path`? since we don't need `src_path` in this function.
I am thinking if `find_git_info` in tern/analyze/docker/dockerfile.py is necessary, because it seems that we don't need to know what is being copied by `ADD` command, and what we use is the `dockerfile path` which is given by a `dfobj`.

nishakm: So it looks to me like `find_git_info` does the ADD command parsing, and you want to check if what got added is a git repository in itself. In this case, you will need functionality to:
1. See if src_path exists (if os.path.exists(src_path))
2. If it does, see if it is a git repository (get_git_toplevel(src_path))
This is only useful if you can produce a full repository url for the git project, otherwise you just get whatever was in the ADD command.
So I think in this case, you do not have to worry about parsing the ADD command to get the `src_path`. We just want to see if the Dockerfile lives in a github repo.

ForgetMe17: Maybe i can keep the current `find_git_info` which implements parsing ADD command, and not use the parsed `src_path`?

nishakm: Yup! In fact, you don't need to parse it out right now. You can work on that part for #585.

ForgetMe17:
Thanks for your suggestions!

tern/analyze/common.py
```diff
@@ -452,3 +453,59 @@ def get_os_style(image_layer, binary):
...
+        if isinstance(output, bytes):
+            sha_info = output.decode('utf-8').split('\n').pop(0)
+    except subprocess.CalledProcessError:
+        logger.debug("Cannot find git repo sha, toplevel path is %s",
+                     path_to_toplevel)
+    return sha_info
+
+
+def get_git_toplevel(path):
+    '''Give a path, return the absolute path to the top level directory if it
```
rnjudge: small nit that doesn't need to be fixed to merge this but should be noted in the future: "Given" not Give and should note that this has to be a path to a directory, not a file.

PR #582, opened 2020-03-16
Integrate file level data from scancode

### LINKED ISSUE #480: Map scancode's data into tern's data model
**Description**
#474 creates the plugin but it will just print the output to console and so it cannot be used in any of the formats. Fill out Tern's data properties based on scancode's data.

**To Do**
This will involve just populating the image object's properties with scancode's data.

**Background**
This issue depends on #476 

**Super Issues**
#284

nishakm:
This resolves #480 

The first commit modifies loading and saving to the cache
for file level data.
The second commit modifies the executor for scancode to
leverage file level data caching and collection for an image.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -11,6 +11,7 @@ import logging
 import os
 
 from tern.classes.package import Package
+from tern.classes.file_data import FileData
 from tern.classes.notice import Notice
 from tern.classes.command import Command
 from tern.command_lib import command_lib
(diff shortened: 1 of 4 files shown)


nishakm:
@PrajwalM2212 the first commit has changes to caching. Please take a look.

PrajwalM2212:
@nishakm Looks good. It is nice to have the logic for both packages and files at `common.py`. 
 It is this type of code organisation that makes this project easy to work with.

rnjudge:
Looks goot to me, OK to merge?

nishakm:
@rnjudge I can merge this

Issue #524, opened 2020-01-17
Include base OS value from /etc/os-release as part of Tern's data model


rnjudge:
**Describe the proposal**
Currently, Tern reports the base OS it finds from /etc/os-release in the output report as a notice object. Instead of reporting the OS as a notice object, propose to add the base OS value to Tern's data model so that it can be extracted directly from a JSON output report. 

I can think of a few ways to do this. There could be a `base_os` attribute in the image_layer class that would get set to the same value for each layer since they are all part of the same container with the same base OS. Or perhaps we only want to set the `base_os` value in the base image layer since doing so for every layer is a bit redundant.

nishakm:
I think we can use the property `os_guess` in the `image_layer` class for now. We can set it for just the first layer.

ForgetMe17:
In `tern/analyze/docker/analyze.py`, funtion `analyze_first_layer`(Line 95 and 108) does not use the return value of `common.get_os_style`. Should we set the `os_guess` as this return value to fix the issue?

rnjudge:
@ForgetMe17 We are going to let a new contributor for the project take this on. However, if you would like to mentor him or be a resource that he can ask questions to (should they arise) please go ahead!

As far as your proposal for setting the `os_guess` value, you are very close. However, `common.get_os_style` does not actually return a value. I think we will want to use `common.get_os_release` to set the `os_guess` value. If `get_os_release` does not return a value, then we will want to use the `os_guess` value(s) from `tern/command_lib/base.yml`.

ForgetMe17:
Yes! I am very happy to help a new contributor, and I think that I will also learn a lot from it!

aditi137:
Hi @rnjudge I am Aditi and was assigned this issue. I am unable to pass the commit sign-off check for creating a PR. Can you please help me with it? Thanks!

ForgetMe17:
Hi, @aditi137 . You can follow commit-message-format part in [CONTRIBUTING.md](https://github.com/vmware/tern/blob/master/CONTRIBUTING.md#commit-message-format). 
The footer of your commit message should like `Signed-off-by: Some Dev <somedev@example.com>`

aditi137:
Hey @ForgetMe17 I did have the footer in all my commit messages. Still not sure of the issue.

rnjudge:
> Hey @ForgetMe17 I did have the footer in all my commit messages. Still not sure of the issue.

[...]

PR #590, opened 2020-03-18
Include base OS value from /etc/os-release as part of Tern's data model

aditi137:
Obtain the `os_guess` value (as used in `common.get_os_style()`) from the path `tern/command_lib/base.yml` when `/etc/os-release` and `/usr/lib/os-release` do not exist.
Resolves: #524
Signed-off-by: Aditi Dutta <aditi011@e.ntu.edu.sg>

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -100,14 +100,10 @@ def get_os_release():
                             constants.etc_release_path)
     lib_path = os.path.join(rootfs.get_working_dir(), constants.mergedir,
                             constants.lib_release_path)
-    tern_path = command_lib.base_file
     if not os.path.exists(etc_path):
         if not os.path.exists(lib_path):
-            if not tern_path:
-                return ''
-            etc_path = tern_path
-        else:
-            etc_path = lib_path
+            return ''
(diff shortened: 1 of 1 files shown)


ForgetMe17:
`Commit 2db1d91 does not have a body.
The subject of commit 2db1d91 should be 50 characters or less.`
You can find the log in the commit_check-Details

nishakm:
@aditi137 Thanks for working on this! Git doesn't understand commit messages with no newlines. You will have to format your commit messages like this:
```
title

body

Resolves #<issue number>

Signed-off-by: Aditi Dutta <aditi011@e.ntu.edu.sg>
```

rnjudge:
I believe the dco issue is coming from the final commit message:

```
 Obtain os_guess value from tern/command_lib/base.yml Resolves: #524 S…

…igned-off-by: Aditi Dutta <aditi011@e.ntu.edu.sg>
```

My suggestion would be to squash these three commits to only one and focus on fixing just the one commit message like Nisha suggested.

aditi137:
Thank you so much @nishakm and @rnjudge. I'll try to follow your method.

Issue #539, opened 2020-02-18
Increase Test Coverage


rnjudge:
**Problem Statement**
Currently, the test coverage for Tern's code is 63%. See below, which is a full coverage run on Tern's code base.

```
$ coverage run -m unittest discover -s tests
$ coverage report --fail-under 85

Name                                      Stmts   Miss  Cover
-------------------------------------------------------------
tern/__init__.py                              0      0   100%
tern/__main__.py                             88     58    34%
tern/analyze/__init__.py                      0      0   100%
tern/analyze/common.py                      247    213    14%
tern/analyze/docker/__init__.py               0      0   100%
tern/analyze/docker/analyze.py               99     79    20%
tern/analyze/docker/container.py            105     57    46%
tern/analyze/docker/dockerfile.py           108     57    47%
tern/analyze/docker/helpers.py              103     79    23%
tern/analyze/docker/run.py                   93     74    20%
tern/analyze/passthrough.py                  44     31    30%
tern/classes/__init__.py                      0      0   100%
tern/classes/command.py                      74      7    91%
tern/classes/docker_image.py                105     16    85%
tern/classes/image.py                        75     13    83%
tern/classes/image_layer.py                 103      5    95%
tern/classes/notice.py                       32      0   100%
tern/classes/notice_origin.py                44      0   100%
tern/classes/origins.py                      31      0   100%
tern/classes/package.py                      71      1    99%
tern/classes/template.py                      9      0   100%
tern/command_lib/__init__.py                  0      0   100%
tern/command_lib/command_lib.py             158    122    23%
tern/report/__init__.py                       0      0   100%
tern/report/content.py                       46     37    20%
tern/report/errors.py                        33      0   100%
tern/report/formats.py                       47      0   100%
tern/report/report.py                        84     55    35%
tern/utils/__init__.py                        0      0   100%
tern/utils/cache.py                          38     24    37%
tern/utils/constants.py                      19      0   100%
tern/utils/general.py                        86     20    77%
tern/utils/rootfs.py                        164     74    55%
tests/test_analyze_docker_dockerfile.py      48      1    98%
tests/test_class_command.py                  84      1    99%
tests/test_class_docker_image.py             48      5    90%
tests/test_class_image.py                    62      1    98%
tests/test_class_image_layer.py              74      1    99%
tests/test_class_notice.py                   38      1    97%
tests/test_class_notice_origin.py            48      1    98%
tests/test_class_origins.py                  60      1    98%
[...]

PR #598, opened 2020-03-22
Add test to validate cache behavior on layers

zoek1:
Add support to three specific elements, Packages, Files and Notices and
the integration of them to ensure the correct behavior of populating
and saving given elements.

The use of mock objects in the cache test cases was implement to focus
on cache behavior and ensure one single reponsabiliby principle.

See: #539

Signed-off-by: Miguel Angel G <miguel@gordian.dev>

### DIFF
--- a/tests/test_analyze_common.py
+++ b/tests/test_analyze_common.py
@@ -4,9 +4,13 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import unittest
+from unittest.mock import Mock
+
 from tern.analyze import common
 from test_fixtures import TestImage
 from tern.classes.command import Command
+from tern.classes.file_data import FileData
+from tern.utils import cache
 
 
 class TestAnalyzeCommon(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


zoek1:
Any feedback on this? :smiley_cat:

tests/test_analyze_common.py
```diff
@@ -3,10 +3,16 @@
 # Copyright (c) 2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+from __future__ import absolute_import
```
nishakm: We only support Python 3.5 and higher. Do we need this for something else?

nishakm:
@zoek1 Thanks for the PR! It mostly looks good. I asked a question in the code and a small change in the commit message below:
```
The use of mock objects in the cache test cases was implement to focus
on cache behavior and ensure one single reponsabiliby principle. --> can you elaborate on this in your commit message?
```

tests/test_analyze_common.py
zoek1: No, sorry. I miss delete such line. I'll remove it and update the commit message.

nishakm:
@zoek1 Thanks for working on this! One small nit to fix before we merge this:
```
one single reponsabiliby principle. --> please change to responsibility
```
Thanks so much!

tests/test_analyze_common.py
zoek1: Done! :smile:

zoek1:
I updated the commit message and removed the `__future__` import. Also new changes in master broke the tests but now are fixed. I'll continue adding more tests in another PR related to `commons.py`. :smiley_cat:

nishakm:
@zoek1 Thanks for updating your commit message. This PR looks good. I will merge it after we cut a release (should be this week).

Issue #584, opened 2020-03-16
Include file level licenses in the default report


nishakm:
**Description**
Include licenses found by scancode (if any) into the default report format

**Background**
This depends on https://github.com/vmware/tern/pull/582 to be merged

**Super Issues**
#583

mukultaneja:
@nishakm, I would like to take this issue!

nishakm:
@mukultaneja It's a high priority bug though. So we need a PR pretty quickly. If you think you can dedicate time to it, then sure.

mukultaneja:
sure @nishakm, I will do it.

PR #599, opened 2020-03-22
Include file level licenses in the default report

mukultaneja:
This PR does following,

1. It Includes licenses found by scancode (if any) into
the default report format.

2. It adds `layer_file_licenses_list` message string to be
included into default report.

Resolved: #584

Signed-off-by: mukultaneja <mukultaneja91@gmail.com>

### DIFF
--- a/tern/formats/default/generator.py
+++ b/tern/formats/default/generator.py
@@ -29,11 +29,24 @@ def print_full_report(image):
         if layer.import_image:
             notes = notes + print_full_report(layer.import_image)
         else:
-            for layer_origin in layer.origins.origins:
-                notes = notes + content.print_notices(layer_origin,
-                                                      '\t', '\t\t')
+            if len(layer.origins.origins) == 0:
+                notes += '\tLayer: {0}:\n'.format(layer.fs_hash[:10])
+            else:
+                for layer_origin in layer.origins.origins:
+                    notes = notes + content.print_notices(layer_origin,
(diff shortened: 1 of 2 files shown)


tern/formats/default/generator.py
```diff
@@ -34,19 +34,35 @@ def print_full_report(image):
...
                     layer_pkg_list.append(pkg)
                 if package.pkg_license not in layer_license_list and \
                         package.pkg_license:
                     layer_license_list.append(package.pkg_license)
-            # Collect packages + licenses in the layer
-            notes = notes + formats.layer_packages_list.format(
-                list=", ".join(layer_pkg_list) if layer_pkg_list else 'None')
-            notes = notes + formats.layer_licenses_list.format(list=", ".join(
-                layer_license_list) if layer_license_list else 'None')
-            notes = notes + formats.package_demarkation
```
nishakm: Actually, we do want to report that no packages and licenses were found:
```
------------------------------------------------

        Packages found in Layer:  None
        Licenses found in Layer:  None
------------------------------------------------
```
The problem is that the layer sha is not reported.
This patch seems to report it:
```
        Layer: d2e6d9a0a0: 
```
But nothing about package and license. We want both:
```
        Layer: d2e6d9a0a0: 
        Packages found in Layer:  None
        Licenses found in Layer:  None
```

nishakm:
@mukultaneja This looks good. I'll merge it.

PR #602, opened 2020-03-23
analyze: docker: Check cache before finding shell

nishakm:
For images that have their files analysed but don't have a shell
or package manager, the analysis will error out with the appropriate
message that a package manager or shell was not found. However, the
file level analysis still existed in the cache, so it should not
error out in this case. This commit fixes the error by checking
if loading from the cache was successful first. Consequently, it
also checks to see if a shell was returned in the analysis of the
first layer so that analysis of subsequent layers can proceed only
if there is no cached data and there is a shell.

### DIFF
--- a/tern/analyze/docker/analyze.py
+++ b/tern/analyze/docker/analyze.py
@@ -84,17 +84,21 @@ def abort_analysis():
 
 
 def analyze_first_layer(image_obj, master_list, redo):
-    # find the binary and shell by mounting the base layer
-    target = rootfs.mount_base_layer(image_obj.layers[0].tar_file)
-    binary = common.get_base_bin()
-    shell = get_shell(image_obj, binary)
-    # set up a notice origin for the first layer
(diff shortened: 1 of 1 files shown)


tern/analyze/docker/analyze.py
```diff
@@ -84,17 +84,21 @@ def abort_analysis():
...
-    if binary:
-        if not common.load_from_cache(image_obj.layers[0], redo):
-            # Determine pacakge/os style from binary in the image layer
+    # set shell to empty string in case one is found
+    shell = ''
+    # try to load from cache
+    loaded = common.load_from_cache(image_obj.layers[0], redo)
+    # if there is no cache, find the binary and shell by mounting the
+    # base layer
+    if not loaded:
```
rnjudge: Will this cache have the binary information as well?

rnjudge: I guess the cache will have the `/etc/os-release` information so we don't necessarily need the binary.

Issue #585, opened 2020-03-16
Find full GitHub project link in Dockerfile lock functionality


rnjudge:
**Describe the Feature**
There is a pending PR for issue #509 that will append the git project name and sha in a comment for Dockerfile lock functionality. This is a good start but, ideally, we would be able to append the full https://<repolink><reponame> URL to the comment. In many cases the repository will be hosted on GitHub under the 'origin' label (i.e. when you list `git remote -v`) but this is not always the case. The user may have re-named the label from origin to something else or the code may not even be hosted on GitHub. There may also be multiple repos listed.

**Use Cases**
This will enhance the information supplied to the user of Tern's Dockerfile Lock functionality.

**Implementation Changes**
Run `git remote -v` at the top level of the git directory that is being `ADD`ed in the Dockerfile. Significant parsing will need to take place to the output to determine the repo URL. Most changes will take place in `tern/analyze/docker/dockerfile.py` in the `find_git_info()` function although you may also want to create a new function to determine the git project URL.

ForgetMe17:
Maybe i can work on this.

rnjudge:
@ForgetMe17 yes, let's get your other PR merged first and then you can build upon it for this issue :) Thanks!

ForgetMe17:
I think I can work on this issue.

rnjudge:
A PR for this issue was merged to master after the 2.0.0 release. Closing this.

PR #606, opened 2020-03-25
Find Git Project URL

ForgetMe17:
Adds git project url to the comments.

In tern/analyze/common.py
Adds function find_git_url which parses the output of 'git remote
-v' to get all the fecth labels containing a http or https url and
returns them as string.

Add test for this function.

Works towards #585.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -563,9 +563,54 @@ def get_git_sha(path_to_toplevel):
     return sha_info
 
 
+def get_git_url(dockerfile_path):
+    '''Given a dockerfile_path, return url of git project which contains
+    the dockerfile in a form of list.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
+    command = ['git', 'remote', '-v']
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=dockerfile_folder_path)
+        if isinstance(output, bytes):
+            lines = output.decode('utf-8').split('\n')
+            # pop the last line which is an empty line
+            lines.pop()
+            url_list = set()
+            for line in lines:
+                extract_url = extract_git_url_from_line(line)
+                if extract_url:
+                    url_list.add(extract_url)
+    except subprocess.CalledProcessError:
+        logger.debug("Cannot find git repo url, path is %s",
+                     dockerfile_folder_path)
+    return url_list
+
+
+def extract_git_url_from_line(line):
+    '''Given a line of git remote -v output, parse the url
+    line structure is '<label>\t<url> (fetch)'''
+    split_line = line.split(' ')
+    extract_url = ''
+    # use fetch url
+    if split_line[1] == '(fetch)':
+        split_line = split_line[0].split('\t')
+        # use https or git@ type
+        full_url = split_line[1]
+        if full_url.startswith('https://'):
+            extract_url = full_url.lstrip('https://')
+        elif full_url.startswith('http://'):
+            extract_url = full_url.lstrip('http://')
+        elif full_url.startswith('git'):
+            extract_url = full_url.replace('git@github.com:', 'github.com/', 1)
+        extract_url = extract_url.rstrip('.git')
+    return extract_url
+
+
 def get_git_toplevel(path):
-    '''Give a path, return the absolute path to the top level directory if it
-    is in a git repository. Empty string will be returned if not.'''
+    '''Given a path, return the absolute path to the top level directory if it
+    is in a git repository. Empty string will be returned if not.
+    Path should be a path to a directory not to a file.'''
     command = ['git', 'rev-parse', '--show-toplevel']
     path_to_toplevel = ''
     try:
--- a/tern/analyze/docker/dockerfile.py
+++ b/tern/analyze/docker/dockerfile.py
@@ -254,6 +254,11 @@ def find_git_info(line, dockerfile_path):
     logger.debug('Parsed src_path is %s', src_path)
     # get the git project info
     comment_line = common.check_git_src(dockerfile_path)
+    # get the git project link
+    url_list = common.get_git_url(dockerfile_path)
+    if url_list:
+        comment_url = ', '.join(url_list)
+        comment_line += ', project url(s): ' + comment_url
     return comment_line
 
 
--- a/tests/test_analyze_common.py
+++ b/tests/test_analyze_common.py
@@ -5,6 +5,7 @@
 
 import unittest
 from unittest.mock import Mock
+import re
 
 from tern.analyze import common
 from test_fixtures import TestImage
@@ -21,12 +22,14 @@ class TestAnalyzeCommon(unittest.TestCase):
         self.image = TestImage('5678efgh')
         self.file = FileData('README.txt', '/home/test')
         cache.cache = {}
+        self.test_dockerfile = 'tests/dockerfiles/buildpack_deps_jessie_curl'
 
     def tearDown(self):
         del self.image
         del self.command1
         del self.command2
         cache.cache = {}
+        del self.test_dockerfile
 
     def testGetShellCommands(self):
         command = common.get_shell_commands("yum install nfs-utils")
@@ -247,6 +250,19 @@ class TestAnalyzeCommon(unittest.TestCase):
         self.command2.set_remove()
         self.assertEqual(
             len(common.get_installed_package_names(self.command2)), 0)
+
+    def testGetGitURL(self):
+        ''' we check the url to be the following form:
+        github.com/<username>/tern'''
+        url_list = common.get_git_url(self.test_dockerfile)
+        check_num = len(url_list)
+        pass_num = 0
+        git_username_reg = r'([a-zA-Z\d_-]{0,38})'
+        pattern = re.compile(r'github.com/'+git_username_reg+r'/tern')
+        for url in url_list:
+            if pattern.match(url):
+                pass_num += 1
+        self.assertEqual(pass_num, check_num)
 
 
 if __name__ == '__main__':


tern/analyze/common.py
```diff
@@ -555,9 +556,42 @@ def get_git_sha(path_to_toplevel):
...
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=path_to_toplevel)
+        if isinstance(output, bytes):
+            lines = output.decode('utf-8').split('\n')
+            url_dict = {}
+            for line in lines:
+                # line structure is '<label>/t<url> (fetch)'
+                split_line = line.split(' ')
+                if len(split_line) < 2:
```
nishakm: Can you explain why you chose this check?

tern/analyze/common.py
```diff
@@ -555,9 +556,42 @@ def get_git_sha(path_to_toplevel):
...
+                    continue
+                # use fetch url
+                if split_line[1] == '(fetch)':
+                    split_line = split_line[0].split('\t')
+                    # use http or https type
```
nishakm: This could also be of the form `git@github.com/`. Maybe also check for `startswith('git')`?

tern/analyze/common.py
```diff
@@ -555,9 +556,42 @@ def get_git_sha(path_to_toplevel):
...
+                    if split_line[1].startswith('http'):
+                        url_dict[split_line[0]] = split_line[1]
```
nishakm: You may have to do more manipulation if the remote happened to be of the form `git@`

tests/test_analyze_docker_dockerfile.py
```diff
@@ -121,8 +121,13 @@ def testFindGitInfo(self):
         sample_line = 'ADD plain_file /tmp'
         comment_line = dockerfile.find_git_info(sample_line, self.pin_add)
         project_name = comment_line.split(',')[0]
+        url_info = comment_line.split(',')[2].strip(' ')
         self.assertIn(project_name, ['git project name: project',
                                      'git project name: tern'])
+        self.assertIn(url_info,
+                      ['url info: origin:https://github.com/vmware/tern '
+                       'upstream:https://github.com/vmware/tern.git',
+                       'url info: (not found)'])
```
nishakm: I may have missed something obvious here but why would url be `(not found)`?

tern/analyze/common.py
ForgetMe17: The last line of output is a empty line, so i use this to skip the last line.

tern/analyze/common.py
ForgetMe17: Yes, I can add this to the output comment.

tests/test_analyze_docker_dockerfile.py
ForgetMe17: In the circleci test, the project folder seems not to have the git project url. In github `Pull Request Lint and Test / PRs (pull_request)` test, it has the git project url. Since we need to pass both check, i use this form of test, like the test for git project name which is in the form of assertIn.

nishakm: OK, I got the CircleCI issue. I think there should be a better way to test this. Let me get back to you on this one.

tern/analyze/common.py
nishakm: You can remove the last line by doing `lines = output.decode('utf-8').split('\n').pop()`

tern/analyze/common.py
```diff
@@ -555,9 +556,42 @@ def get_git_sha(path_to_toplevel):
...
+    command = ['git', 'remote', '-v']
+    url_info = '(not found)'
+    try:
+        output = subprocess.check_output(  # nosec
+            command, stderr=subprocess.DEVNULL, cwd=path_to_toplevel)
+        if isinstance(output, bytes):
+            lines = output.decode('utf-8').split('\n')
+            url_dict = {}
+            for line in lines:
+                # line structure is '<label>/t<url> (fetch)'
```
nishakm: Did you mean `<label>\t<url>`?

tern/analyze/common.py
nishakm: Also, I don't think you need to keep the remote name. You can just keep the url.

tern/analyze/common.py
```diff
@@ -555,9 +556,42 @@ def get_git_sha(path_to_toplevel):
...
+                if split_line[1] == '(fetch)':
+                    split_line = split_line[0].split('\t')
+                    # use http or https type
+                    if split_line[1].startswith('http'):
+                        url_dict[split_line[0]] = split_line[1]
+            if url_dict:
+                # list all remotes
+                url_info = ''
+                for k, v in url_dict.items():
+                    url_info += k + ':' + v + ' '
```
nishakm: Here you can make a list of urls from the `git remote -v` result and make sure to convert it into the form `https://<domain>/<org>/<repo>`

tests/test_analyze_docker_dockerfile.py
nishakm: I tried running this test locally on my clone of tern and I get `url info: (not found)` which shouldn't happen. It should be something like `url info: https://github.com/nishakm/tern, https://github.com/vmware/tern`.

rnjudge:
Hi @ForgetMe17 -- Thanks for your PR. The purpose of issue #585 is to enhance the Dockerfile lock capability, which means that somehow we need to update the dockerfile object that we print out when the user calls the `tern lock` command. While your PR creates a function to find the git URL, it doesn't actually append it as a comment to the dockerfile object `ADD` line. Because you have already done the hard part of finding the URL, updating the dfobj is just a matter of calling your `get_git_url` function in the already existing `expand_add_command` or in the `find_git_info` function (your choice), both in `Dockerfile.py` and appending the found URL to the comment line.

tests/test_analyze_docker_dockerfile.py
nishakm: I ran a test on running `git remote -v` on CircleCI and GHA. It looks like the difference is just the addition of the upstream project in the GHA environment, which is expected because all the tests are run in the same environment there.
CircleCI:
```
#!/bin/bash -eo pipefail
python ci/test_commit_message.py

origin	git@github.com:vmware/tern.git (fetch)
origin	git@github.com:vmware/tern.git (push)

Commit message checks pass
```
GHA:
```

Run python ci/test_commit_message.py
origin	https://github.com/vmware/tern (fetch)
origin	https://github.com/vmware/tern (push)
upstream	https://github.com/vmware/tern.git (fetch)
upstream	https://github.com/vmware/tern.git (push)

Commit message checks pass
```
Either way, you should get one url here: `https://github.com/vmware/tern`

Another issue is that if you are working on a fork of the main project, then `git remote -v` will return something like `https://github.com/<username>/tern.git`. In this case the assert will still have to pass. I wonder if you could instead check if the beginning string has `https://github.com` and the end says `tern`.

tern/analyze/common.py
ForgetMe17: Yes! The form of one line in the output is like `<label>\t<url> (fetch/push)`

ForgetMe17:
Thanks for your reviews!

tests/test_analyze_docker_dockerfile.py
ForgetMe17: Shoud we include ssh?like
`upstream	ssh://git@github.com/vmware/tern.git (fetch)`

nishakm: I don't think we should. In fact, I think not including the transport protocol (`http/https/git`) would be better. So `get_git_url` should produce `['github.com/vmware/tern']` for the CircleCI and GHA case and `['github.com/vmware/tern', 'github.com/<username>/tern']` for the local development case.

tests/test_analyze_common.py
```diff
@@ -248,6 +251,19 @@ def testGetInstalledPackageNamesWithRemoveFlag(self):
         self.assertEqual(
             len(common.get_installed_package_names(self.command2)), 0)
 
+    def testGetGitURL(self):
```
ForgetMe17, line 254: In the latest commit, i change the test from test_analyze_dockerfile.py to test_analyze_common.py, since testing the url list is more convient than testing the comments and also needs less modifications.

tern/analyze/common.py
```diff
@@ -563,9 +563,54 @@ def get_git_sha(path_to_toplevel):
     return sha_info
 
 
+def get_git_url(dockerfile_path):
+    '''Given a dockerfile_path, return url of git project which contains
+    the dockerfile in a form of list.'''
+    # get the path of the folder containing the dockerfile
+    dockerfile_folder_path = os.path.dirname(dockerfile_path)
```
ForgetMe17, line 570: I use the dockerfile_folder_path as the cwd which assums that the dockerfile is under a git project.

tern/analyze/common.py
```diff
@@ -563,9 +563,54 @@ def get_git_sha(path_to_toplevel):
...
+            command, stderr=subprocess.DEVNULL, cwd=dockerfile_folder_path)
+        if isinstance(output, bytes):
+            lines = output.decode('utf-8').split('\n')
+            # pop the last line which is an empty line
+            lines.pop()
+            url_list = []
+            for line in lines:
+                extract_url = extract_git_url_from_line(line)
+                if extract_url and extract_url not in url_list:
+                    url_list.append(extract_url)
```
nishakm, line 583: If `url_list = set()` then you don't have to do the `extract_url not in url_list` check and just do `url_list.add(extract_url)`.

tests/test_analyze_common.py
nishakm: Can you post a test result from a Dockerfile here then?

ForgetMe17: I ran a local test on `tests/dockerfiles/pin_add_command_test/pin_add_command_test_dockerfile`
Here is the test code
`def testCreatLockedFile(self):`
     `dfobj = dockerfile.get_dockerfile_obj(self.testfile)`
     `dffile = dockerfile.create_locked_dockerfile(dfobj)`
     `dockerfile.write_locked_dockerfile(dffile, self.testfile + '_lock')`
I have some problems with uploading the locked file, so i just paste the content below(2 lines).
`FROM ubuntu@sha256:04d48df82c938587820d7b6006f5071dbbffceb7ca01d2814f81857c631d44df`
`ADD plain_file /tmp # git project name: tern, HEAD sha: 0f84e2099e085fd085a6e6348b8fa411011ff57f, project url: github.com/ForgetMe17/tern github.com/vmware/tern 
`

tern/analyze/docker/dockerfile.py
```diff
@@ -254,6 +254,13 @@ def find_git_info(line, dockerfile_path):
     logger.debug('Parsed src_path is %s', src_path)
     # get the git project info
     comment_line = common.check_git_src(dockerfile_path)
+    # get the git project link
+    url_list = common.get_git_url(dockerfile_path)
+    if url_list:
+        comment_url = ''
+        for url in url_list:
+            comment_url += url + ' '
+        comment_line += ', project url: ' + comment_url
```
rnjudge, line 263: small nit: change `project url` to `project url(s)` since there could be more than one.

tern/analyze/docker/dockerfile.py
```diff
@@ -254,6 +254,13 @@ def find_git_info(line, dockerfile_path):
     logger.debug('Parsed src_path is %s', src_path)
     # get the git project info
     comment_line = common.check_git_src(dockerfile_path)
+    # get the git project link
+    url_list = common.get_git_url(dockerfile_path)
+    if url_list:
+        comment_url = ''
+        for url in url_list:
+            comment_url += url + ' '
```
rnjudge, line 262: @nishakm  small nit -- do we want to separate the URLs with nothing but a space? I would think if there is a list of URLs we would want to separate by comma, but maybe that is getting too detailed :)

rnjudge:
> @rnjudge Can you look this over? It looks good to me.

I had two small nits -- one question for Nisha and one change request. Other than that looks good. Thank you very much @ForgetMe17!

tern/analyze/docker/dockerfile.py
nishakm: @rnjudge Good catch! I didn't notice. @ForgetMe17 can you make this change?

ForgetMe17: Yes!

ForgetMe17: The output now is (2 lines)
`FROM ubuntu@sha256:04d48df82c938587820d7b6006f5071dbbffceb7ca01d2814f81857c631d44df`
`ADD plain_file /tmp # git project name: tern, HEAD sha: 7009ee03938a4f18b9e884f101a4d1bb999a36f3, project url(s): github.com/vmware/tern, github.com/ForgetMe17/tern`

rnjudge:
@nishakm My small requested changes have been made, so this looks good to me for merge.

nishakm:
Merging. Thanks, @ForgetMe17, for sticking with this PR!

PR #632, opened 2020-04-09
Add docs to create a dev env. when using vagrant

rnjudge:
If Tern users follow the steps in the README to get started with
vagrant, they will end up using the bootstrap.sh file, which installs
the latest PyPI version of Tern. This is oftentimes all users need to
get started. If, however, they would like to pick up any new changes or
features thave have been merged since the release, they will need to
install Tern from the tip of master instead.

This commit updates the documentation in the README and CONTRIBUTING to
so that users can use vagrant and install Tern in a way that works for
their needs, either at the latest release version or in development mode
at the tip of master. It also updates the GitHub links in the vagrant
sections to refer to Tern's new org, tern-tools/tern.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,22 +103,27 @@ You may have already cloned the project and started working on it. If you're rea
 7. Highly recommended: If you want to set up your project for long-term open source contribution, I highly suggest following [this setup](https://github.com/nishakm/puns).
 
 ### Setting up a development environment on Mac and Windows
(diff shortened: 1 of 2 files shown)


vagrant/bootstrap.sh
```diff
@@ -24,4 +24,5 @@ sudo apt-get install -y docker.io
 sudo usermod -a -G docker vagrant
 
 # Install tern
-pip3 install tern
+git clone https://github.com/vmware/tern.git
+pip3 install .
```
nishakm, line 28: Any chance of making a "development bootstrap" and a "try it out bootstrap"?

Issue #499, opened 2019-11-22
Add functionality to extract Debian license data from copyright text


rnjudge:
**Describe the Feature**
Debian packages do not contain license metadata. Instead, the license information is contained in the copyright text. This results in an empty license list when you run Tern with the default output against a Debian image/BaseOS.

**Use Cases**
This would be helpful for the default output of Tern, primarily when running against Debian base images.

**Implementation Changes**
Add a function that can parse the copyright text for the licenses in a Debian image and report it as the "License" field for a package.

nishakm:
I was introduced to this project and I think it would be a good option to either incorporate or use as a dependency: https://github.com/nexB/debut

mukultaneja:
@nishakm, I would like to work on this issue.

nishakm:
@rnjudge OK to reassign to @mukultaneja?

rnjudge:
Sure!

rnjudge:
This issue was resolved with PR https://github.com/tern-tools/tern/pull/694. Closing.

Issue #565, opened 2020-02-29
"/bin/bash: cat: command not found" running example debian:buster scan


williamsmj:
When I run `tern -l report -o output.txt -i debian:buster` as suggested in the README, I get the output below. Is this the expected output? If so, I found the repeated error messages surprising, and it might be worth clarifying in the README that this is indeed what you should see.

```
This report was generated by the Tern Project
Version: 1.0.1

Docker image: debian:buster:
	Layer: 5a87144dc3:
		info: Found 'Debian GNU/Linux 10 (buster)' in /etc/os-release.
		info: Layer created by commands: /bin/sh -c #(nop) ADD file:e05e45c33042db4ec7f71a5952d65ee8cb3786dcd76fa7a990f48a2def1344e2 in / 
		info: Retrieved by invoking listing in command_lib/base.yml
names:
	in container:
	dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'

versions:
	in container:
	pkgs=`dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'`
	for p in $pkgs; do dpkg -l $p | awk 'NR>5 {print $3}'; done

copyrights:
	in container:
	pkgs=`dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'`
	for p in $pkgs; do cat /usr/share/doc/$p/copyright; echo LICF; done




		error: /bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
/bin/bash: cat: command not found
[snip ~100 lines of this]
/bin/bash: cat: command not found

	Invoking commands from command_lib/base.yml:
		warning: No listing for key licenses. Consider adding this listing to command_lib/base.yml.
No listing for key srcs. Consider adding this listing to command_lib/base.yml.

[...]

PR #649, opened 2020-04-22
Extract Debian license data from copyright text

mukultaneja:
This PR does following,

1. It proposed a license extraction logic based
on `pkg_format`.
2. It proposed a change in `base.yml` file to extract
copyright file location instead of copyright text.
3. It adds `pkg_licenses` property in `Package`
class to record multiple liceses in a package.
4. It updates default report generator to
print multiple licenses of a packge.
5. It updates default generator to address
'to many brances' build error.

Work towards: #499

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ dockerfile-parse~=0.0
 requests~=2.23
 stevedore>=1.32
 pbr>=5.4
+debut>=0.9
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -23,6 +23,8 @@ from tern.utils import cache
 from tern.utils import constants
 from tern.utils import general
 from tern.utils import rootfs
+from debut import debcon
+from debut import copyright as debut_copyright
 
 # global logger
 logger = logging.getLogger(constants.logger_name)
@@ -217,7 +219,8 @@ def convert_to_pkg_dicts(pkg_dict):
                'version': 'versions',
                'pkg_license': 'licenses',
                'copyright': 'copyrights',
-               'proj_url': 'proj_urls'}
+               'proj_url': 'proj_urls',
+               'pkg_licenses': 'pkg_licenses'}
     pkg_list = []
     len_names = len(pkg_dict['names'])
     # make a list of keys that correspond with package property names
@@ -235,6 +238,62 @@ def convert_to_pkg_dicts(pkg_dict):
             a_pkg.update({key: value[index]})
         pkg_list.append(a_pkg)
     return pkg_list
+
+
+def get_licenses_from_deb_copyright(deb_copyright):
+    '''
+    Given the debian copyright texts,
+    1. parse each copyright text
+    2. filter out all the available licenses
+    3. returns a list of unique licenses found inside
+    the copyright text
+    '''
+    collected_paragraphs = list()
+    pkg_licenses = set()
+    for paragraph in iter(debcon.get_paragraphs_data(deb_copyright)):
+        if 'license' in paragraph:
+            cp = debut_copyright.CopyrightLicenseParagraph.from_dict(paragraph)
+            collected_paragraphs.append(cp)
+
+    deb_pkg_data = debut_copyright.DebianCopyright(
+        collected_paragraphs).to_dict()
+    for paragraph in deb_pkg_data.get("paragraphs"):
+        pkg_license = paragraph.get("license")
+        if not pkg_license.startswith("*"):
+            pkg_license = pkg_license.split("\n")[0]
+            if pkg_license:
+                pkg_licenses.add(pkg_license)
+
+    return list(pkg_licenses)
+
+
+def get_deb_package_licenses(deb_copyrights):
+    '''
+    Given the all debian copyrights,
+    returns a list of unique licenses in
+    each copyright
+    '''
+    deb_licenses, deb_licenses_dict = list(), dict()
+    for deb_copyright in deb_copyrights:
+        pkg_licenses = get_licenses_from_deb_copyright(deb_copyright)
+        n = len(pkg_licenses)
+        # using an extra dict to validate if we have
+        # seen the license before. If `yes` then discard it
+        # else keep it. This would help us to get all
+        # unique licenses across all the packages
+        for index in range(n):
+            pkg_license = pkg_licenses[index]
+            if deb_licenses_dict.get(pkg_license, None):
+                pkg_licenses[index] = None
+            else:
+                deb_licenses_dict[pkg_license] = 1
+        pkg_licenses = list(
+            filter(lambda pkg_license: pkg_license is not None, pkg_licenses))
+        deb_licenses.append(pkg_licenses)
+    # clear the dict content
+    deb_licenses_dict.clear()
+
+    return deb_licenses
 
 
 def add_base_packages(image_layer, binary, shell):
@@ -261,6 +320,11 @@ def add_base_packages(image_layer, binary, shell):
             origin_layer, Notice(snippet_msg, 'info'))
         # get all the packages in the base layer
         pkg_dict, invoke_msg, warnings = collate_list_metadata(shell, listing)
+
+        if listing.get("pkg_format") == "deb":
+            pkg_dict["pkg_licenses"] = get_deb_package_licenses(
+                pkg_dict["copyrights"])
+
         if invoke_msg:
             image_layer.origins.add_notice_to_origins(
                 origin_layer, Notice(invoke_msg, 'error'))
--- a/requirements.in
+++ b/requirements.in
@@ -12,3 +12,4 @@ requests
 stevedore
 pbr
 dockerfile-parse
+debut
--- a/tern/classes/package.py
+++ b/tern/classes/package.py
@@ -39,6 +39,7 @@ class Package:
         self.__checksum = ''
         self.__origins = Origins()
         self.__files = []
+        self.__pkg_licenses = []
 
     @property
     def name(self):
@@ -60,9 +61,17 @@ class Package:
     def pkg_license(self):
         return self.__pkg_license
 
+    @property
+    def pkg_licenses(self):
+        return self.__pkg_licenses
+
     @pkg_license.setter
     def pkg_license(self, pkg_license):
         self.__pkg_license = pkg_license
+
+    @pkg_licenses.setter
+    def pkg_licenses(self, pkg_licenses):
+        self.__pkg_licenses = pkg_licenses
 
     @property
     def copyright(self):
--- a/tern/formats/default/generator.py
+++ b/tern/formats/default/generator.py
@@ -67,27 +67,65 @@ def get_layer_info_list(layer):
         pkg = package.name + "-" + package.version
         if pkg not in layer_pkg_list and pkg:
             layer_pkg_list.append(pkg)
-        if package.pkg_license not in layer_license_list and \
-                package.pkg_license:
-            layer_license_list.append(package.pkg_license)
+
+        package_licenses = get_package_licenses(package)
+        for package_license in package_licenses:
+            if package_license not in layer_license_list:
+                layer_license_list.append(package_license)
+
     return layer_pkg_list, layer_license_list, file_level_licenses
+
+
+def get_package_licenses(package):
+    '''Given a package collect complete list of package licenses'''
+    pkg_licenses = set()
+    if package.pkg_license:
+        pkg_licenses.add(package.pkg_license)
+
+    if package.pkg_licenses:
+        for pkg_license in package.pkg_licenses:
+            if pkg_license:
+                pkg_licenses.add(pkg_license)
+
+    return list(pkg_licenses)
+
+
+def get_layer_packages_licenses(layer):
+    '''Given a image layer collect complete list of package licenses'''
+    pkg_licenses = set()
+    for package in layer.packages:
+        package_licenses = get_package_licenses(package)
+        for package_license in package_licenses:
+            pkg_licenses.add(package_license)
+
+    return list(pkg_licenses)
+
+
+def get_layer_files_licenses(layer):
+    '''Given a image layer collect complete list of file licenses'''
+    file_level_licenses = set()
+    for f in layer.files:
+        for license_expression in f.license_expressions:
+            if license_expression:
+                file_level_licenses.add(license_expression)
+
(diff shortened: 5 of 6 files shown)


tern/command_lib/base.yml
```diff
@@ -94,7 +94,7 @@ dpkg:
       1:
         container:
           - "pkgs=`dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'`"
-          - "for p in $pkgs; do cat /usr/share/doc/$p/copyright; echo LICF; done"
+          - "for p in $pkgs; do /bin/cat /usr/share/doc/$p/copyright; echo LICF; done"
```
nishakm, line 97: Is this change more related to resolving #565 ?

rnjudge: If so, please move to its own PR :)

mukultaneja: yes, it will address issue #565 as well. Here, I have submitted this change because I am using `debut` external utility to parse Debian copyright data which accepts copyright file location instead of copyright data.

rnjudge: Not sure what @nishakm thinks but I personally think that if it's addressing another bug it should be it's own PR or at least a separate commit in this PR. Your commit message says "It proposed a change in base.yml file to extract copyright file location instead of copyright text." but you also say that this is addressing but #565. The change in this PR only seems to be changing/specifying the location of the cat utility (which addresses #565) -- it's not clear to me how this changes the information that is retrieved as your commit message suggests.

mukultaneja: @rnjudge I can understand your point is but the two changes which I proposed through this PR are related to each other. If I don't make a change in `base.yml`, I can't parse copyright text of any Debian package using `debut` utility. It is a pure coincidence that another issue is already being notified which requires the same change of `base.yml`. Although, I am ok to dedicate this `base.yml` change to issue #565. Anyway, the change in `base.yml` is the base of parsing copyright text.

rnjudge: If you want to keep the commits together that's fine but can you update your commit message to more accurately describe the changes you're making? The change to base.yml doesn't `extract copyright file location instead of copyright text` itself -- it may enable debut to do this -- but the change in base.yml as far as Tern code base is concerned is only specifying the location of the cat utility. Please also mention that this simultaneously works towards issue #565 somewhere in your commit message. Thank you!

mukultaneja: Sure!

rnjudge: Thank you! Really appreciate this PR!

tern/analyze/common.py
```diff
@@ -249,6 +250,24 @@ def add_base_packages(image_layer, binary, shell):
...
         # get all the packages in the base layer
         pkg_dict, invoke_msg, warnings = collate_list_metadata(shell, listing)
+
+        if listing.get("pkg_format") == "deb":
+            pkg_dict["licenses"] = list()
+            pd = os.path.join(
+                rootfs.mount_dir, constants.temp_folder, constants.mergedir)
+            for copyright in pkg_dict["copyrights"]:
+                cp = '{0}{1}'.format(pd, copyright.strip())
+                d = DebianCopyright.from_file(cp)
```
nishakm, line 260: I wonder if DebianCopyright has a parser that would read data from `pkg_dict["copyright"]` instead. If reading it from file is more accurate, we may want to create a separate python script for it.

mukultaneja: I think it is a good idea to abstract this function call using a custom module/script. It may also help in a situation where we may want to add more filtration for license text in the future.

nishakm: That is the rub. Typically, a package has a declared license (see rpm, pip, etc). In Debian's case, their licenses are listed in the copyright text of the package. So this is really post processing of the licenses in a Debian specific copyright file. This is why I asked if `debut` can parse any individual copyright text and get a list of licenses for it. If it can, you can just write a function to do it and call it here.

mukultaneja: @nishakm how to handle `Unable to import 'debut.copyright'` lint error here?

nishakm: you would have to install tern first in your virtual environment `pip install -e.[dev]`.

mukultaneja: I am sorry but I was referring circleci linting errors.

nishakm: You will have to add `debut` to `requirements.txt` and `requirements.in`.

tern/command_lib/base.yml
mukultaneja: @rnjudge, Done! I have moved the `base.yml` to a separate pull request.

tern/analyze/common.py
mukultaneja: @nishakm, I have updated the code with requested changes. Please take a look and share your feedback.

requirements.txt
```diff
@@ -12,3 +12,4 @@ dockerfile-parse~=0.0
 requests~=2.23
 stevedore>=1.32
 pbr>=5.4
+debut
```
nishakm, line 15: Can you give me the version of debut you were using at the point you created the PR and a list of dependencies?

tern/analyze/common.py
```diff
@@ -222,6 +223,59 @@ def convert_to_pkg_dicts(pkg_dict):
...
+    '''
+    Given the debian copyright file location,
+    1. parse the copyright file
+    2. filter out all the available licenses
+    3. returns a list of unique licenses found inside
+    the copyright text
+    '''
+    mounted_path = os.path.join(
+        rootfs.mount_dir, constants.temp_folder, constants.mergedir)
+    deb_copyright = '{0}{1}'.format(mounted_path, deb_copyright)
```
nishakm, line 236: This is interesting: what gets returned by `collate_list_metadata` should be a dictionary with a key called `copyrights` and these should be copyright text not copyright file locations. So what `deb_copyright` here is a string that looks like `.tern/temp/mergedir<some text>`. But `DebianCopyright.from_file` requires a file location only.

tern/analyze/common.py
```diff
@@ -222,6 +223,59 @@ def convert_to_pkg_dicts(pkg_dict):
...
+    deb_pkg_data = DebianCopyright.from_file(deb_copyright).to_dict()
```
nishakm, line 237: From the above inference, I don't think we can use this. And it looks like debut doesn't have a direct way of instantiating this class given just a debian copyright text. However, there also seems to be `debcon.get_paragraphs_data` [here](https://github.com/nexB/debut/blob/972761071d70a2665a434c9df83fdaf371c338ee/src/debut/debcon.py#L502) which we can possibly use directly. It might be worthwhile to look into this first.

mukultaneja: @nishakm the link which you pointed out to `get_paragraph_data_from_file` was my earlier proposal in the notebook but you suggested going in this way.

tern/analyze/common.py
mukultaneja: @nishakm if we see this code in the combination of updated `base.yml` with returning copyright file location then it would make complete sense. I did this because as you pointed out that `debut` does not have a way as of now to use the copyright text directly.

requirements.txt
mukultaneja: @nishakm, here along with `debut` version do I need to add its dependencies also in `requirements.txt`?

tern/analyze/common.py
nishakm: Yes, `get_paragraph_data_from_file` reads a file and then calls `get_paragraphs_data` on the read text. This is the functionality I would rather use, instead of referring to the file path. This helps with our separation of how individual OSs choose to record this metadata. I haven't tested out `get_paragraphs_data` but you could try it out and see if it works.

requirements.txt
nishakm: Nope. We just need to figure out how much the dependencies overlap with the packages we are already using. If there is no overlap then we can add `debut>=major.minor` where major.minor is the version you have. If there is overlap, we will have to add it as `debut~=major.minor`.

tern/analyze/common.py
nishakm: I would rather not have that change as I would like the enabling on `base.yml` to be as uniform across all package managers as possible.

requirements.txt
mukultaneja: @nishakm I have updated the `debut` version in `requirements.txt`. Please take a look once

requirements.txt
```diff
@@ -12,3 +12,4 @@ dockerfile-parse~=0.0
 requests~=2.23
 stevedore>=1.32
 pbr>=5.4
+debut==0.9.4
```
nishakm, line 15: Are any of debut's dependencies overlapping with tern's? If no then please change this to `debut>=0.9` and if yes then please change it to `debut~=0.9`. Thanks!

mukultaneja: Done!

tern/analyze/common.py
mukultaneja: Done! Please review the update file.

tern/analyze/common.py
```diff
@@ -222,6 +228,68 @@ def convert_to_pkg_dicts(pkg_dict):
...
+    '''
+    collected_paragraphs = list()
+    pkg_licenses = set()
+    for paragraph in iter(debcon.get_paragraphs_data(deb_copyright)):
+        if 'format' in paragraph or 'format-specification' in paragraph:
+            cp = CopyrightHeaderParagraph.from_dict(paragraph)
+        elif 'files' in paragraph:
+            cp = CopyrightFilesParagraph.from_dict(paragraph)
+        elif 'license' in paragraph:
+            cp = CopyrightLicenseParagraph.from_dict(paragraph)
```
nishakm, line 247: I wonder if we can just use this.

tern/analyze/common.py
```diff
@@ -222,6 +228,68 @@ def convert_to_pkg_dicts(pkg_dict):
...
+        else:
+            cp = CatchAllParagraph.from_dict(paragraph)
+        collected_paragraphs.append(cp)
+
+    deb_pkg_data = DebianCopyright(collected_paragraphs).to_dict()
+    for paragraph in deb_pkg_data.get("paragraphs"):
+        pkg_license = paragraph.get("license") \
+            if 'license' in paragraph else ''
```
nishakm, line 255: Because we seem to be doing the same check here, regardless of which paragraphs we get.

tern/analyze/common.py
```diff
@@ -249,6 +317,11 @@ def add_base_packages(image_layer, binary, shell):
             shell = constants.shell
         # get all the packages in the base layer
         pkg_dict, invoke_msg, warnings = collate_list_metadata(shell, listing)
+
+        if listing.get("pkg_format") == "deb":
+            pkg_dict["licenses"] = get_deb_package_licenses(
```
nishakm, line 322: Here's where our logic fails: we're assuming that each package has 1 declared license, but for Debian, one package can have many licenses. We would like to associate the licenses to the package it came from. How do we do that if we can only store 1 license string in the `Package` class?

nishakm: I would suggest to create a new property in `Package` called "licenses" to store multiple licenses if they can be found.

requirements.txt
```diff
@@ -12,3 +12,4 @@ dockerfile-parse~=0.0
 requests~=2.23
 stevedore>=1.32
 pbr>=5.4
+debut>=0.9.4
```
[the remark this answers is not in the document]
mukultaneja: Done!

tern/analyze/common.py
```diff
@@ -23,6 +23,12 @@
 from tern.utils import constants
 from tern.utils import general
 from tern.utils import rootfs
+from debut import debcon
+from debut.copyright import DebianCopyright
+from debut.copyright import CatchAllParagraph
+from debut.copyright import CopyrightHeaderParagraph
+from debut.copyright import CopyrightFilesParagraph
+from debut.copyright import CopyrightLicenseParagraph
```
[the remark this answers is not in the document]
mukultaneja: Done! I am using `from debut import copyright as debut_copyright` because of using built-in copyright build error.

tern/analyze/common.py
```diff
@@ -222,6 +228,68 @@ def convert_to_pkg_dicts(pkg_dict):
     return pkg_list
 
 
+def get_deb_package_license(deb_copyright):
```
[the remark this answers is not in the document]
mukultaneja: I have updated the function name with `get_licenses_from_deb_copyright`

tern/analyze/common.py
mukultaneja: Done! I have removed every other parsing function call except `CopyrightLicenseParagraph`.

tern/analyze/common.py
mukultaneja: Done!

tern/analyze/common.py
mukultaneja: > Here's where our logic fails: we're assuming that each package has 1 declared license, but for Debian, one package can have many licenses. We would like to associate the licenses to the package it came from. How do we do that if we can only store 1 license string in the `Package` class?

I can get your point here that is why the previous commits in this PR collect all the licenses of a package and assign `pkg_license` property a string of all licenses collected. So say if a package `dpkg` got two licenses `['GPL-V2, LGPL V3+']` then the package's  `pkg_license` property would be having a string value `GPL-V2, LGPL V3+` which means a package would be associated with its respective licenses.

mukultaneja: > I would suggest to create a new property in `Package` called "licenses" to store multiple licenses if they can be found.

I have incorporated this change in the latest commit. Now I am having `pkg_licenses` property in the `Package` class and storing package licenses inside it. This change is also required to update the default report generator to print `pkg_licenses` if we have any so made that change also in the latest commit.

nishakm:
@mukultaneja There might be some minor cleaning up needed. Would you mind if I submit this commit for you with the clean-up changes on top of it?

mukultaneja:
> @mukultaneja There might be some minor cleaning up needed. Would you mind if I submit this commit for you with the clean-up changes on top of it?

@nishakm, thanks for reviewing the changes. If it's easy to define the cleanup changes I would like to take them and finish the PR but if they are hard to define then I guess you can better take the decision :)

rnjudge:
Thanks for all your work on this @mukultaneja. Closing this PR as https://github.com/tern-tools/tern/pull/694 replaces it.

PR #654, opened 2020-04-23
Fix prospector too-many-branches error for plugins

nishakm:
Two files related to formats and extensions failed the R0912 pylint
check i.e. too-many-branches. This commit refactors the two files
such that the error no longer exists. The functionality does not
change but some of the logic was moved to different functions
which get called in the same file.

### DIFF
--- a/tern/formats/spdx/spdxtagvalue/layer_helpers.py
+++ b/tern/formats/spdx/spdxtagvalue/layer_helpers.py
@@ -88,6 +88,45 @@ def get_file_license_expressions(layer_obj):
     return list(license_expressions)
 
 
+def get_package_license_info_block(layer_obj):
+    '''Given a layer object, return the SPDX document block with licenses
+    from the files in the layer. Return an empty string if the files are
+    not analyzed'''
+    block = ''
+    if layer_obj.files_analyzed:
+        license_expressions = get_file_license_expressions(layer_obj)
+        if license_expressions:
+            for le in license_expressions:
+                block += 'PackageLicenseInfoFromFiles: {}\n'.format(le)
+        else:
+            block = 'PackageLicenseInfoFromFiles: NONE\n'
+    return block
+
+
+def get_layer_file_data_block(layer_obj, template):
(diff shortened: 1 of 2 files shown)


tern/formats/spdx/spdxtagvalue/layer_helpers.py
```diff
@@ -101,7 +140,6 @@ def get_layer_block(layer_obj, template, image_loc=''):
     would be downloaded along with the image.'''
     block = ''
     mapping = layer_obj.to_dict(template)
-    layer_checksum = get_layer_checksum(layer_obj)
```
rnjudge, line 104: This is not a blocker to merge, just curious on rationale: Why delete this line and variable assignment here if we are replacing it instead with two function calls to `get_layer_checksum` and still creating the layer_checksum variable inside `get_layer_file_data_block`?

rnjudge:
I had one question, otherwise looks good.

tern/formats/spdx/spdxtagvalue/layer_helpers.py
nishakm: The layer's checksum is used to differentiate the same file existing in multiple layers (particularly in the case of BerkleyDB files which get copied over to the next layer if any one of the linked files are modified). So for `get_layer_file_data_block` we need the layer's checksum in a couple of places to generate the SPDXID of the file. Meanwhile, we still need the layer's checksum for describing it at the layer level, hence just calling `get_layer_checksum` in line 164 below.

rnjudge: The `layer_checksum` we get in `get_layer_file_data_block` will be the same as `get_layer_checksum(layer_obj)` passed in to `block += 'PackageChecksum: {}\n'.format(get_layer_checksum(layer_obj))` right?

nishakm: Yup, same one. It just needs to be used in the `get_file_block` so rather than call it again, I stored the value in a temporary variable.

PR #657, opened 2020-04-23
Run prospector on all files for CI tests

rnjudge:
This commit changes the CI tests to run prospector on the entire tern
directory when a PR is opened instead of just the files that were
changed in the PR.

Before this commit, we were only running prospector on the files that
were changed in an attempt to avoid known linting failures being
reported. The known linting errors have been fixed, however, and we can
now run prospector on the full directory, which returns more accurate
results.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
       - setup
       - run: pip install prospector>=1.2 GitPython~=2.1
       - run: pip install .
-      - run: c=`python ci/evaluate_docs.py`; if [ -z $c ]; then echo "No .py files to lint"; else echo $c | xargs prospector; fi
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - setup
       - run: pip install prospector>=1.2 GitPython~=2.1
       - run: pip install .
-      - run: c=`python ci/evaluate_docs.py`; if [ -z $c ]; then echo "No .py files to lint"; else echo $c | xargs prospector; fi
+      - run: prospector .
```
nishakm, line 34: I think you can just run `prospector` without specifying a directory. It's how I've been running it anyway.

PR #669, opened 2020-04-27
Revert "Update base.yml to echo debian copyright file path"

nishakm:
`echo` actually does not work in this case as it only prints
out the path to the copyright file, not the copyright text.

This reverts commit 44ae4472eb07de47291afbae6381e68fa02a52c5.


mukultaneja:
sure, I will update the PR with what you suggested.

nishakm:
@mukultaneja This will revert your PR. Thanks again for working on this.

PR #676, opened 2020-04-29
analyze: Reduce branches in analyze_first_layer

nishakm:
Moved the lines that try to retrieve base layer packages to its
own function to reduce the number of branches in the analyze_first_layer
function.


nishakm:
Prospector passes: https://github.com/nishakm/tern/pull/new/analyze-fix-prospector-error

Issue #682, opened 2020-05-06
Split commands from shell script in dockerfile


ForgetMe17:
**Description**
As the first step to parse Dockerfile RUN instruction commands, we need to split commands from a dockerfile RUN commands. A full shell script contains these elements,
-  `variable assignment` : eg. `dpkgArch="$(dpkg --print-architecture)";`
- `command line`: eg. `wget -O go.tgz "$url";`
- `branch`: eg. `case .... esac;`, `if ... fi;`
- `function`:  eg. `foo(){.....}`
In this task, we need to parse a shell script into above parts and store them, possible in a List. For branches, we may need another task to parse its branches into `variable assignment` and `command line`.
 
**To Do**
Given a full shell script from RUN instruction of dockerfile, return a data structure that stores the parsing result. Most modifications will be made in `general.py`. 
- We can add a function called `parse_shell_script()` in `general.py` as a main function. It receives a string of RUN instruction, which is a shell script. It returns a command dict containing all the possible command in the given shell script.
- Once the above command dict is given, we can modify the `parse_command()` in `general.py` to parse the command dict to find possible installed software. This could be implemented by the following task issues.

Something need to be discussed:
- Data structure can be List. 

1. Maybe we can define a class for the parsed elements. It has a property shows what is this element, like `variable assignment` or `command line` etc. We can implement further parse like parsing the `branch` as the class function. 

2. Or maybe we can use dict as the elements, it has a key called `statement` whose value can be `variable assignment` or `command line` etc. And we add new functions in `general.py`.
- How to handle `branch`.
I am thinking adding a extra key pr property for the `command line` indicating if it is in a branch.
At first we can take the default branch as the command for parse. In the further plan, we can add more analyze on this like considering varibles, or if this branch only infects the version we install but not the software name we install.
- Should we consider `function`?

**Background**
On looking at the type of parsing needed for full shell scripts embedded in the run command, we may need to develop a shell script parser to catch all places where software could have been installed.

**Super Issues**
#521

nishakm:
An example of RUN commands in a Dockerfile that is difficult to parse:
https://github.com/docker-library/python/blob/06d7952f07eae237666f2fd0ccf6189a5d8ee83c/3.9-rc/buster/Dockerfile

ForgetMe17:
Here is a shell script containing `loop`, and the current output is 
[buster-slim.txt](https://github.com/tern-tools/tern/files/4694306/buster-slim.txt).
[...]

PR #688, opened 2020-05-14
Split commands from shell script in dockerfile

ForgetMe17:
This pr adds 2 functions in utils/general.py:
1. split_shell_script()
This functions split a shell script(in string format) into 3 types
of statements: variable, command and branch(if and case).
A dictionary is used to store the infomation of the statement.
It has following keys:
- type: statement type(variable, command, if or case)
- content: statement sentence(eg. 'apt-get update')
- name: (only for variable) variable name
- value: (only for variable) variable value
(more keys will be added when we move on to parsing the branches)
returns a list of statements

2. parse_shell_script()
This function parse a shell script to get a command list.
First it calls split_shell_script() to get a statement list,
For command, it calls parse_command() to parse the command
and extract command name, options and words.

Works towards #682.

### DIFF
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -8,6 +8,7 @@ import random
 import re
 import tarfile
 import subprocess  # nosec
+import shlex
 from contextlib import contextmanager
 from pathlib import Path
 from pbr.version import VersionInfo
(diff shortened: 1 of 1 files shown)


ForgetMe17:
Here are my local test output on 
[https://github.com/docker-library/python/blob/06d7952f07eae237666f2fd0ccf6189a5d8ee83c/3.9-rc/buster/Dockerfile](url).

[test_statments.txt](https://github.com/tern-tools/tern/files/4628075/test_statments.txt)
in the test_statements.txt: 
statements part are return value `statements` in split_shell_script()
commands part are return value `commands` in parse_shell_script()

ForgetMe17:
And this is the output on 
[https://github.com/docker-library/golang/blob/69b2283d45ae89c84383d41a9b1aab4d84e5cbae/1.14/buster/Dockerfile](https://github.com/docker-library/golang/blob/69b2283d45ae89c84383d41a9b1aab4d84e5cbae/1.14/buster/Dockerfile)
This dockerfile contains a if-statement and a case-statement.

Currently we are not handling branches.
[test_statments-2.txt](https://github.com/tern-tools/tern/files/4628139/test_statments-2.txt)

nishakm:
@ForgetMe17 I wonder if we can make use of your shell script parser to parse all RUN commands. What do you think about arranging your statement dictionary like this:
```
# assuming each element in the list is a statement
[{"variable": {"name": "dpgkarch", "value": "$(...)"}, "content": "..."},
 {"command": {"name": "apt-get",
                       "options": ["--no-install-recommends", "-y"],
                       "words": [...]}, "content": "...."},
 {"branch": {"type": "if", "statements": [ ... ] # statements within the if ... fi block},...]
```
This way, you capture all the elements you are interested in within one dictionary which you can use throughout the code.

The logic looks good to me. Good job!

ForgetMe17:
Thanks for your advice! I will work on rearrange my statements dictionary and move on to implement a function to parse `if` and `case` statements. I will push another commit in this pr.

ForgetMe17:
In the new commit i remove 1 function in previous commit and add 3 new functions.

remove 1. parse_shell_script()
no longer need this since in this commit, i move the parsing
command function into split_shell_script().

add 1. parse_shell_variables_and_command()
given a sentence, classify the variable and command type,
and then parse it.

add 2. parse_shell_if_branch()
extract all commands in the if sentence, return a branch dictionary.

add 3. parse_shell_case_branch()
extract all commands in the case sentence, return a branch dictionary.

my local test result are:
[1-14.txt for (https://github.com/docker-library/golang/blob/69b2283d45ae89c84383d41a9b1aab4d84e5cbae/1.14/buster/Dockerfile)](https://github.com/tern-tools/tern/files/4651098/1-14.txt)
[3-9-rc.txt for (https://github.com/docker-library/python/blob/06d7952f07eae237666f2fd0ccf6189a5d8ee83c/3.9-rc/buster/Dockerfile)](https://github.com/tern-tools/tern/files/4651103/3-9-rc.txt)

tern/utils/general.py
```diff
@@ -182,3 +183,175 @@ def parse_image_string(image_string):
...
+        - statements: branch variable and command statements list
+    - content: statement sentence(eg. 'apt-get update')
+
+    sentence is the text we are parsing.(string)
+    statement is the parsed result for sentence.(dictionary)
+    return a list of statements
+    """
+    # add ';' in case that the script is not ended with ';'
+    script += ';'
+    if skip_branch:
```
ForgetMe17, line 212: I use skip_branch here since i need to reuse split_shell_script() to parse the sentences in a branch. When we are dealing with sentences in a branch, they are already been splited by shlex, so we do not need to split again.

tern/utils/general.py
```diff
@@ -182,3 +183,175 @@ def parse_image_string(image_string):
...
+    # priority: keywords > remove_token > separator
+    # if come across a keyword(if or case), ignore remove_token and separator
+    # until meet the ending keyword(fi or esac)
+    # we will parse the keywords sentence in other functions
+    for tk in lexer:
+        # 1. check if in a branch
+        if sentence and sentence[0] in keywords.keys():
+            sentence.append(tk)
+            if tk.startswith(keywords[sentence[0]]):
+                if skip_branch:
```
ForgetMe17, line 233: Here we will skip branches in branch sentences, which means we will ignore nested branches.

tern/utils/general.py
nishakm: I think you can avoid this by creating another function that deals with the output of `shlex.split(script, posix=False)`.

tern/utils/general.py
```diff
@@ -182,3 +183,175 @@ def parse_image_string(image_string):
...
+def split_shell_script(script, skip_branch=False):
+    """Given a shell script, split it into statement of
+    1. variable
+    2. command
+    3. branch: if, case
+    Use dictionary to store infomations on a statement, it has following keys:
+    - variable(only for variable statement):
+        - name: variable name
+        - value: variable value
+    - variable(only for variable statement):
```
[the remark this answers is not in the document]
ForgetMe17: Yes, here is a typo.

tern/utils/general.py
```diff
@@ -182,3 +183,175 @@ def parse_image_string(image_string):
...
+        - name: command name(eg. apt-get)
+        - options: command options(eg. ["--no-install-recommends", "-y"])
+        - words: command words
```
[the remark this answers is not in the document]
ForgetMe17: Not yet. I am thinking of this function in the following issues.

ForgetMe17:
Making another PR on this issue.
Closing.

PR #690, opened 2020-05-14
extensions: scancode: Increase timeout limit

nishakm:
Golang binaries take a large amount of time to scan by scancode.
The default timeout is at 120 seconds. Scancode reports an error
when it hits the timeout limit. In order to let scancode finish
scanning filesystems with large binaries, we increate the timeout
to 300 seconds which is 5 minutes.


rnjudge:
Seems reasonable to me :) LGTM.

PR #691, opened 2020-05-19
[WIP] UI for tern json reports

abhaykatheria:
This PR adds functionality to produce html page for json data,
and renders it in a tree view. 
Signed-off-by: abhaykatheria <abhay.katheria1998@gmail.com>

### DIFF
--- a/tern/formats/json/generator.py
+++ b/tern/formats/json/generator.py
@@ -10,7 +10,6 @@ JSON document generator
 import json
 from tern.formats import generator
 
-
 class JSON(generator.Generate):
     def generate(self, image_obj_list):
         '''Given a list of image objects, create a json object string'''
--- a/tern/formats/json/html.py
+++ b/tern/formats/json/html.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2020 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+
+from tern.utils import general
+
+def get_version():
+    '''Return the version string for the --version command line option'''
+    ver_type, commit_or_ver = general.get_git_rev_or_version()
+    message = ''
+    if ver_type == "package":
+        message = "Tern version {}".format(commit_or_ver)
+    else:
+        message = "Tern at commit {}".format(commit_or_ver)
+    return message
+
+def print_licenses_only(image_obj_list):
+    '''Print a complete list of licenses for all images'''
+    full_license_list = []
+    for image in image_obj_list:
+        for layer in image.layers:
+            for package in layer.packages:
+                if (package.pkg_license and
+                        package.pkg_license not in full_license_list):
+                    full_license_list.append(package.pkg_license)
+
+            for f in layer.files:
+                for license_expression in f.license_expressions:
+                    if license_expression not in full_license_list:
+                        full_license_list.append(license_expression)
+    return full_license_list
+
+
+css ='''
+<style>
+ul, #myUL {
+  list-style-type: circle;
+}
+li {
+    padding-top: 5px;padding-bottom: 5px;
+}
+#myUL {
+  margin: 0;
+  padding: 0;
+}
+.caret {
+  cursor: pointer;
+  user-select: none;
+  text-decoration: underline;
+  font-family: 'Oswald', sans-serif;
+}
+.caret::before {
+  content: " \\25B6";
+  color: ;
+  display: inline-block;
+  margin-right: 6px;
+}
+.caret-down::before {
+  transform: rotate(90deg);
+}
+.nested {
+  display: none;
+}
+.active {
+  display: block;
+} 
+.header {
+  text-align: left;
+  background: #f1f1f1;
+}
+</style>\n
+'''
+
+js= ''' 
+<script>
+var toggler = document.getElementsByClassName("caret");
+var i;
+
+for (i = 0; i < toggler.length; i++) {
+  toggler[i].addEventListener("click", function() {
+    this.parentElement.querySelector(".nested").classList.toggle("active");
+    this.classList.toggle("caret-down");
+  });
+} 
+</script>\n
+'''
+head = """
+<!DOCTYPE html>
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link href="https://fonts.googleapis.com/css2?family=Inconsolata:wght@300&family=Oswald&display=swap" rel="stylesheet"> 
+<link href="https://fonts.googleapis.com/css2?family=Josefin+Sans:ital,wght@1,300&display=swap" rel="stylesheet"> 
+%s \n
+</head>
+<body>
+<div class="header">
+<img src="https://raw.githubusercontent.com/tern-tools/tern/master/docs/img/tern_logo.png" height="60px">\n
+<br>
+</div>\n
+<div style="font-family: 'Josefin Sans', sans-serif;">
+<p>
+%s
+<p>
+The following report was generated for "%s" image.
+</div>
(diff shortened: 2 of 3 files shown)


tern/formats/json/html.py
```diff
@@ -0,0 +1,375 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2017-2020 VMware, Inc. All Rights Reserved.
```
rnjudge, line 3: Small nit but this should just be `Copyright (c) 2020` since that is the year the file originated.

abhaykatheria: ok sure :)

tern/formats/json/html.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+html_code = '''<!DOCTYPE html>
+   <head>
+      <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css" integrity="sha384-Gn5384xqQ1aoWXA+058RXPxPg6fy4IWvTNh0E263XmFcJlSAwiGgFAW/dAiS6JXm" crossorigin="anonymous">
+      <meta charset="utf-8" />
+      <title>Test page</title>
+      <script type="text/javascript" >
+         /**
+         * JSONViewer - by Roman Makudera 2016 (c) MIT licence.
```
rnjudge, line 14: Where was all this code copy and pasted from?

abhaykatheria: This code was adopted from [here](https://www.cssscript.com/json-data-tree-view/) I modified it a little bit to suit our need.

rnjudge:
This is a good start, thanks @abhaykatheria! I attached a few screen shots of the html page that is generated for reference. Here are a few things I noticed.
Screen Shot 2020-05-21 at 10 31 44 AM
1) In some cases, the newline characters are represented as a black `n` at the end of the line.
2) The numbers in the `{}` brackets are confusing. I know this is the number of elements in the list but not sure this always makes sense or is clear.
3) There are a few headings that look like this: `0: {1}` or `1: {3}`. What is the `0` and `1`? This is not intuitive or clear. The headings should be descriptive
4) We likely want to include some basic information at the top of the html page (before the json data starts). Some examples of things that might be useful:
* Command that the user ran to generate the report.
* Version of Tern that was used to generate the report.
* Summary of licenses found?
5) We want to add a note to the user that the report.html was generated if they enable the json output so that they know it's also viewable as an html page.
6) We can use `report.html` as the default name, but if a user specifies an output file using the `-o` command line option, we probably want to the name of the html file to correspond to the name of the file that user specified. i.e If the user names the output file `output.txt` we should name the html file `output.txt.html`. 
7) The numbers like `0:{9}` for the list of packages should be replaced with the names of the packages. 
Screen Shot 2020-05-21 at 10 44 49 AM

Let me know if you have any questions :)

abhaykatheria:
Thanks for your feedback @rnjudge .
1. I have fixed the unwanted n issue.
2. I have made changes that will only show the length of the list for packages and files list.
3. Actually most of the data in JSON is in the form of a list. Unlike the dictionary which has key-value pairs I gave index as heading for each element of the list. Here's a screenshot to make it clear.
image
In this the history list has 2 dicts and both dont have a name key so i thought its better to just name  with their indices.
4. I have made changes to show the current version. I am working on displaying the command. About license summary do we need to list all the licenses at one place or we need to list them layer-wise.?
5. I will add a note in the documentation about it.
6. Nice idea I have implemented it.
7. I made changes for it.
image

rnjudge:
Can you make the `Tern at commit` string not a part of the header? i.e. put the text in the white part just above the list. This is where the licenses should go as well.

>   Actually most of the data in JSON is in the form of a list. Unlike the dictionary which has key-value pairs I gave index as heading for each element of the list. 

I'm also still thinking about what we want to do for the indicie headers. Listing the numerical value is not intuitive but may be our only option.

>     4. I have made changes to show the current version. I am working on displaying the command. About license summary do we need to list all the licenses at one place or we need to list them layer-wise.?

I think listing a summary of all the licensees from all the layers at the top (similar to how we do in the default report) is sufficient. Note that these should be unique license lists.

tern/formats/json/generator.py
```diff
@@ -10,7 +10,6 @@
 import json
 from tern.formats import generator
 
-
```
rnjudge, line 13: This newline needs to be preserved.

tern/formats/json/html.py
```diff
@@ -0,0 +1,175 @@
...
+
+def get_version():
+    '''Return the version string for the --version command line option'''
+    ver_type, commit_or_ver = general.get_git_rev_or_version()
+    message = ''
+    if ver_type == "package":
+        message = "Tern version {}".format(commit_or_ver)
+    else:
+        message = "Tern at commit {}".format(commit_or_ver)
+    return message
```
rnjudge, line 16: This function is unnecessary --`tern/report/content.py` has a function called `get_tool_version()` that will already do this work for you:

```
 15 def get_tool_version():                                                         
 16     '''Return a string describing the version and where it came from'''         
 17     ver_type, ver = get_git_rev_or_version()                                    
 18     if ver_type == 'commit':                                                    
 19         return formats.commit_version.format(commit_sha=ver)                    
 20     return formats.packaged_version.format(version=ver)    
```

So you can

tern/formats/json/html.py
```diff
@@ -0,0 +1,175 @@
...
+            for package in layer.packages:
+                if (package.pkg_license and
+                        package.pkg_license not in full_license_list):
+                    full_license_list.append(package.pkg_license)
+
+            for f in layer.files:
+                for license_expression in f.license_expressions:
+                    if license_expression not in full_license_list:
+                        full_license_list.append(license_expression)
+    return full_license_list
```
rnjudge, line 32: Given that this is basically a direct copy of the same function in `tern/formats/default/generator.py` I wonder if we want to move this functionality to somewhere more common so both the default report and the json html report can use it. @nishakm what do you think about a licenses summary at the top of the html page for json reports? Note: the summary would only be in the html page for the json report, not the actual file.json report itself. I personally think the summary is helpful.

tern/report/report.py
```diff
@@ -154,5 +156,9 @@ def report_out(args, *images):
         logger.error("%s not a recognized plugin.", args.report_format)
     elif args.output_file:
         write_report(report, args)
+        if args.report_format == "json":
```
rnjudge, line 159: We should add a notice to the user (via the logger) that says something like "Generating <filename>.html which can be viewed in the web browser of your choice".

tern/report/report.py
```diff
@@ -154,5 +156,9 @@ def report_out(args, *images):
...
+            create_html(json.loads(report),images,args,filename=args.output_file)
```
rnjudge, line 160: When the user does not specify an output file on the command line (i.e. `tern report -i ubuntu:latest -f json`), the file `report.html.html` is created.

tern/formats/json/html.py
nishakm: You can move this to `tern/report/content.py`. I would rename this function `get_licenses_only` to reflect that we are returning a list rather than a string.

nishakm:
Is the plan to generate json output and print a html file? I wonder if this could be its own format. The reason I propose this is that at some point, tern will run in a docker container as a service and will POST json objects to another service that may render the view. This could be an option for those who just want to run the tool on their desktop during development or in a build pipeline like Jenkins.

tern/report/report.py
rnjudge: Per our discussion today, let's make the html file generation it's own command line options, so you can ignore this suggestion.

PR #694, opened 2020-05-20
Get Debian package licenses from copyrights

nishakm:
This PR adds functionality to parse Debian style
copyright texts to get package level licenses.

Thanks to mukultaneja <mtaneja@vmware.com>
for their work.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -12,3 +12,4 @@ requests
 stevedore
 pbr
 dockerfile-parse
+debut
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ dockerfile-parse~=0.0
 requests~=2.23
 stevedore>=1.32
 pbr>=5.4
+debut>=0.9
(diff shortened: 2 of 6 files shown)


rnjudge:
I think this PR got some additional commits added in with it. Can you rebase with the tip of master?

nishakm:
@mukultaneja please review

rnjudge:
I tested these changes on `debian:buster` and got a list of licenses. Nice work @mukultaneja and @nishakm!! Side note: The license representations are rather messy compared to other package mangers output but that is not a reflection of this PR, rather the way Debian represents them. For example, `GPLv2+` and `GPL-2+` are both reported.

mukultaneja:
> I tested these changes on `debian:buster` and got a list of licenses. Nice work @mukultaneja and @nishakm!! Side note: The license representations are rather messy compared to other package mangers output but that is not a reflection of this PR, rather the way Debian represents them. For example, `GPLv2+` and `GPL-2+` are both reported.

@rnjudge we can find many like that :).

PR #700, opened 2020-05-21
classes: Set digest when instantiating DockerImage

nishakm:
We can make use of the available functions to get the Docker image
digest and set the values for checksum_type and checksum when
instantiating a DockerImage class. This commit will parse the repotag
string to get the repository name and tag. If a tag was used, then
the image digest will also be set. It is assumed that either a tag
or an image digest will be used. The tests have been modified
accordingly.


nishakm:
@rnjudge Let me check if there are any string changes in the reports before merging.

PR #701, opened 2020-05-22
Fix error with missing keys in cache

nishakm:
As the project develops, there will be more data that we will
add to the cache. However, these keys may not exist in users'
caches as they get new releases. This change gets default values
if the keys do not exist using "get".


rnjudge:
Looks good! Merging.

PR #707, opened 2020-05-28
Fix package pinning when package is also pkg mngr

### LINKED ISSUE #702: Dockerfile lock gets confused when installing pip package with an OS package manager
**Describe the bug**
In Tern's Dockerfile we apt-install the pip package like so:
```
RUN tdnf remove -y toybox && tdnf install -y tar findutils attr util-linux python3
python3-pip python3-setuptools git && pip3 install --upgrade pip && pip3 install tern
```

**To Reproduce**
1. Run `tern lock Dockerfile` at the tip of master.
2. See resulting `Dockerfile.lock` file.

**Error in terminal**
The resulting Dockerfile.lock incorrectly pins the python3-pip package like so: **python3-pip==20.1.1==20.1.1-3.7.5-3.ph3**
```
RUN tdnf remove -y toybox && tdnf install -y tar-1.30-4.ph3 findutils-4.6.0-5.ph3
 attr-2.4.48-1.ph3 util-linux-2.32.1-2.ph3 python3-3.7.5-3.ph3 python3-
pip==20.1.1==20.1.1-3.7.5-3.ph3 python3-setuptools-3.7.5-3.ph3 git-2.23.3-1.ph3 && pip3 
install --upgrade pip && pip3 install tern==2.0.
```

**Expected behavior**
The python3-pip package should be pinned properly.

rnjudge:
Since the addition of language package managers, the Dockerfile lock
functionality was getting confused when trying to pin packages that also
double as language package managers.

This commit does the following to resolve the issue:
 1) In analyze.py, instead of comparing each of the packages in the
    layer to see if they are installed in the RUN line, first gather
    the list of packages being installed in the dockerfile RUN line
    and then compare those against the packages installed in the layer.
    This prevents unnecessary comparisons and duplicate pinnings.

 2) Add a 'should_pin' method to dockerfile.py that checks whether the
    package in the RUN line should be pinned or not. Use this function
    in the existing 'expand_package' function in dockerfile.py to
    determine if we should expand/pin the package or not.

 3) Modify the way 'expand_package' in dockerfile.py actually pins the
    package and version in the dockerfile object. Instead of using the
    replace method, which was expanding non-exact matches of a package
    name, we go word by word in the RUN line to determine if a package
    should be pinned.

  4) Modify the 'package_in_dockerfile' function to return the list of
     packages being installed in a given RUN line instead of returning
     true or false. Also change the name of the function to more
     accurately reflect its purpose. The function is now called
     'get_install_packages'.
Resolves #702

### DIFF
--- a/tern/analyze/docker/analyze.py
+++ b/tern/analyze/docker/analyze.py
@@ -142,15 +142,19 @@ def analyze_subsequent_layers(image_obj, shell, master_list, redo, dfobj=None,  
                 if dfile_lock:
                     # collect list of RUN commands that could install pkgs
                     run_dict = d_file.get_run_layers(dfobj)
-                    for package in image_obj.layers[curr_layer].packages:
-                        # check that package is in current dfobj RUN line
-                        if d_file.package_in_dockerfile(
-                                run_dict[curr_layer - 1], package.name):
-                            d_file.expand_package(
-                                run_dict[curr_layer - 1], package.name,
-                                package.version,
-                                command_lib.check_pinning_separator(
-                                    pkg_listing))
+                    # use the run_dict to get list of packages being installed
+                    install_list = d_file.get_install_packages(
+                        run_dict[curr_layer - 1])
+                    for install_pkg in install_list:
+                        for layer_pkg in image_obj.layers[curr_layer].packages:
+                            if install_pkg == layer_pkg.name:
+                                # dockerfile package in layer, let's pin it
+                                d_file.expand_package(
+                                    run_dict[curr_layer - 1],
+                                    install_pkg,
+                                    layer_pkg.version,
+                                    command_lib.check_pinning_separator(
+                                        pkg_listing))
             if command_list:
                 rootfs.undo_mount()
                 rootfs.unmount_rootfs()
--- a/tern/analyze/docker/dockerfile.py
+++ b/tern/analyze/docker/dockerfile.py
@@ -15,6 +15,7 @@ from tern.utils import general
 from tern.utils import constants
 from tern.analyze.docker import container
 from tern.analyze import common
+from tern.command_lib import command_lib
 
 # global logger
 logger = logging.getLogger(constants.logger_name)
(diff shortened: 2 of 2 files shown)


tern/analyze/docker/analyze.py
```diff
@@ -142,15 +142,19 @@ def analyze_subsequent_layers(image_obj, shell, master_list, redo, dfobj=None,
...
-                                run_dict[curr_layer - 1], package.name,
-                                package.version,
-                                command_lib.check_pinning_separator(
-                                    pkg_listing))
+                    # use the run_dict to get list of packages being installed
+                    install_list = d_file.get_install_packages(
+                        run_dict[curr_layer - 1])
+                    for install_pkg in install_list:
+                        for layer_pkg in image_obj.layers[curr_layer].packages:
+                            if install_pkg == layer_pkg.name:
```
nishakm, line 150: how about `if install_pkg in [p.name for p in image_obj.layers[curr_layer].packages]`?

tern/analyze/docker/dockerfile.py
```diff
@@ -188,12 +189,38 @@ def expand_from_images(dfobj):
             counter = counter + 1
 
 
+def should_pin(run_line, package, index):
+    '''This check is necessary to make sure that commands that double as
+    packages (i.e. language package managers like pip, gem) are pinned properly
+    in a locked dockerfile. For example, in the RUN line "pip install pkg"
+    we would not pin 'pip' to its version, but if the line was
+    "apt install pip" then we would want to pin the version.
```
nishakm, line 197: The command parsing functions create `Command` objects with names which are different from the arguments (words in this case). Why not make use of this functionality?

tern/command_lib/command_lib.py
```diff
@@ -28,6 +28,7 @@
 common_file = pkg_resources.resource_filename('tern', 'command_lib/common.yml')
 
 # command library
+global command_lib
```
nishakm: I don't think this should be here?

nishakm: Oh looks like you removed it :)

tern/analyze/docker/analyze.py
rnjudge: install_pkg is an element in install_list -- so it has no value to compare to the loop in this suggestion. Also, if we do this then we are basically just doing what we did before, looking through the packages in the layer to see if they are in the dockerfile RUN line. We want to only look at the packages that are being installed in the RUN line first and then check if they are in the layer.

rnjudge: How do we get the `layer_pkg` version this way? using `p.version` in the `expand_function` is unrecognized.

rnjudge: Agree your suggestion is more compact, but we need the `layer_pkg` property data available inside the if statement.

nishakm: OK, got it.

tern/analyze/docker/dockerfile.py
nishakm: Per our discussion, the issue here is locating where on the full RUN command we need to insert the pinning separator. That isn't covered by the `Command` object. So this looks good.

Issue #678, opened 2020-04-30
Enable analysis for OCI images


nishakm:
**Describe the Feature**
OCI images have a different layout from the one created by `docker save`. So Tern cannot analyze these images. Enable analysis for OCI images.

**Use Cases**
Tools like podman and umoci create OCI images. The layout is understood by most registries. Docker images can be converted to OCI images using skopeo. Tern should be able to deal with these images.

**Implementation Changes**
1. Create a derivative class of Image called `OCI` which will implement `load_image` based on the new layout.
2. Figure out a way to indicate the layout type in the command line. This would mean a major revision change.

nishakm:
This is now fixed.

PR #712, opened 2020-06-01
Enable analysis for OCI images

mukultaneja:
This commit proposes changes to analyze
OCI images using tern.

Work towards: #678


mukultaneja:
@nishakm @rnjudge I have been working on issue #678 which is to analyze OCI images. I would like to propose an approach here. This PR as of now is in scratch form and mostly to take suggestions over my approach. Please take a look once and share your feedback. 
I am using the below command to analyze OCI debain image,
```
skopeo copy docker://debain:latest oci:///root/debian:latest
tern report -c oci:///root/debian:latest -o output.txt -f default
```

tern/__main__.py
```diff
@@ -104,6 +104,16 @@ def do_main(args):
             else:
                 run.execute_docker_image(args)
                 logger.debug('Report completed.')
+        elif args.oci_image:
+            # Check if the image is of oci://<image-location>:<image-tag>
+            if not check_oci_image_string(args.oci_image):
+                sys.stderr.write('Error running Tern\n'
+                                 'Please provide docker image '
```
rnjudge, line 111: is the `--oci-image` command line option designed to handle both a docker image string in `image:tag` and an `oci://<image-location>:<image-tag>` format? If so, we should modify the help string on the parser `add_argument` lines below as that help message only directs users to use `oci://<image-location>:<image-tag>`. If not, this string should be modified to direct users to only provide the oci image location/tag format.

tern/analyze/docker/run.py
```diff
@@ -69,6 +69,35 @@ def analyze(image_obj, args, dfile_lock=False, dfobj=None):
         analyze_docker_image(image_obj, args.redo, dfile_lock, dfobj)
 
 
+def execute_oci_image(args):
```
rnjudge, line 72: We should create a directory called `tern/analyze/oci` and put this is a `run.py` file there.

tern/utils/oci_image.py
```diff
@@ -0,0 +1,32 @@
+
```
rnjudge, line 1: I think this file should be under: `tern/analyze/oci/helpers.py`

tern/__main__.py
mukultaneja: I think we should have separate arguments to deal with both types of images and yes I need to change the string message to direct users to only provide the oci image location/tag format.

nishakm: Would it make sense to add a new command line option called `--type, -t` were you can enter `docker` or `oci`?

nishakm:
> Currently, I am targetting OCI images without umoci tool version. Please take a look once and share your feedback.
I'd say this is the expected way that folks would give an OCI image to tern, i.e., they've used `skopeo` to copy a docker image into the oci format into a local directory and provide that directory to tern.

> I am using the below command to analyze OCI debain image,
> 
> ```
> skopeo copy docker://debain:latest oci:///root/debian:latest
> tern report -c oci:///root/debian:latest -o output.txt -f default
> ```

I wonder if `podman` would be able to natively build an OCI type image. Maybe that's something I can poke at.

tern/__main__.py
mukultaneja: @nishakm I added `-t` option as you mentioned and I think it would make more sense having one option `-i` to define container image and `-t` is type of container image to call respective operations on to the image.

tern/utils/oci_image.py
mukultaneja: I updated the code to add modularity similar to the docker.

tern/analyze/oci/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
```
rnjudge, line 3: small nit: should be `2020` as the copyright year.

tern/__main__.py
```diff
@@ -69,6 +69,33 @@ def create_top_dir(working_dir=None):
...
+        # Check if the image is of image:tag
+        # or image@digest_type:digest format
+        if not check_image_string(args.image):
+            sys.stderr.write('Error running Tern\n'
+                             'Please provide docker image '
+                             'string in image:tag or '
+                             'image@digest_type:digest format\n')
+            sys.exit(1)
+        if general.check_tar(args.image):
+            logger.error("%s", errors.incorrect_raw_option)
```
nishakm, line 84: I think this option should be different from the `-i` option now. Or else this is going to get more complicated. @rnjudge What do you think?

rnjudge: Not sure I understand your question. The `-w` option is already separate from the `-i` option, right?

nishakm: Shouldn't our check here then be `if args.raw`?

rnjudge: I believe this check here is to make sure that the user didn't provide a raw option tarball when they are trying to provide a docker or OCI image.

tern/analyze/oci/__init__.py
mukultaneja: I updated the copyright year.

tern/__main__.py
mukultaneja: @nishakm, this check is part of the `execute_image` method and this method is only for docker or OCI image execution. The `args.raw_image` still I am keeping the part `do_main` method.

rnjudge: Confirmed that this is just a validation check to make sure the user didn't accidentally choose the `-i` option but provide a raw image tarball. Seems OK to me.

tern/__main__.py
```diff
@@ -153,12 +168,18 @@ def main():
...
+    parser_report.add_argument('-i', '--image',
+                               help="Image that exists locally -"
+                               "either can be a docker image with format"
                                " image:tag"
+                               " or an OCI image with format"
+                               " oci://<image-location>:<image-tag>"
                                " The option can be used to pull docker"
                                " images by digest as well -"
                                " <repo>@<digest-type>:<digest>")
+    parser_report.add_argument('-t', '--type',
```
rnjudge, line 180: I would like to propose that we call this option`client` instead of `type`. This makes more sense to me as we would like to support the podman client in the future which represents image blogs differently than docker or skopeo might, even if they all are stored on dockerhub technically as OCI images.

nishakm: Agree - this makes better sense than the generic "type" in the container image context.

nishakm:
> I am using the below command to analyze OCI debain image,
> 
> ```
> skopeo copy docker://debain:latest oci:///root/debian:latest
> tern report -c oci:///root/debian:latest -o output.txt -f default
> ```

From my experiments, it looks like the typical way that an oci image is represented is as a directory with this layout:
```
index.json
oci-layout
blobs/
        |_ <hash>
                       |_ artifact1
                       |_ artifact2
```

You can get this layout using the "docker-ish" way using podman.
```
podman pull docker://debian:latest
podman save --format oci-archive -o debian.tar debian:latest
tar xvf debian.tar -C debian-oci
```

Podman also has an `oci-dir` format for saving but this looks like a cross between the docker image layout and the oci image layout. I will ask about this further. I think it's a way to make podman compatible with images built with docker but not sure.

You can also get this layout using skopeo:
```
skopeo copy docker://debian:latest oci:debian:latest
```
This will create a folder called `debian` which will have the above OCI layout.

So we have a number of ways to get to the common directory with the OCI layout. So I think we should expect that users provide a path to the OCI directory as input for Tern and perform tests on the layout accordingly. We can inform users in the documentation how to generate the OCI directory. Note that we can't do this with Docker as that directory is only accessible via `docker save` which won't work unless you do a `docker pull`, so we can keep that machinery.

nishakm:
@mukultaneja This is a pretty big contribution. Would it be possible to split it up into several smaller commits that we can merge in little by little? How about starting off with the OCIImage image class and some tests for that?

mukultaneja:
> @mukultaneja This is a pretty big contribution. Would it be possible to split it up into several smaller commits that we can merge in little by little? How about starting off with the OCIImage image class and some tests for that?

do you want me to close this PR and send separate PRs with several commits? I was thinking if we could use this PR as a module and make modifications in it as we feel. Say the next commit would be adding tests for OCI images. Anyway so far changes, would allow us to generate a default report for an OCI image.

nishakm:
> > @mukultaneja This is a pretty big contribution. Would it be possible to split it up into several smaller commits that we can merge in little by little? How about starting off with the OCIImage image class and some tests for that?
> 
> do you want me to close this PR and send separate PRs with several commits? I was thinking if we could use this PR as a module and make modifications in it as we feel. Say the next commit would be adding tests for OCI images. Anyway so far changes, would allow us to generate a default report for an OCI image.

I would rather the commits be self describing and independent. You can either do one PR with many commits or several PRs each being independent. The idea is that the changes are small and logical so they are easy to follow.

nishakm:
Reg: `podman save --format oci-dir`: https://github.com/containers/libpod/issues/6544#issuecomment-641523650

PR #722, opened 2020-06-05
formats: spdxtagvalue: Fix 2.2 validation errors

nishakm:
This commit brings in two changes in order to produce valid SPDX 2.2
documents

The first change is to use the licenses at the file level instead of
license expressions. This is because the license expressions do not
necessarily match SPDX license identifiers. All references to licenses
within the document either need to be SPDX license identifiers or a
license reference string which would then be expanded at the end of
the document. This is the same way we manage the licenses reported
by the package managers i.e. each license string has a unique identifier
associated with it, which is used throughout the document and later
describing the actual license string collected. This method of listing
licenses is to avoid warnings with the SPDX validator tools.

In order to accomplish this, we add two generic helper functions in
formats.py which can be used by package_helpers.py and file_helpers.py
to make the license blocks of the document. We also add functions in
file_helpers.py and layer_helpers.py to collect unique licenses for
the FileData and ImageLayer objects. All this information is collected
in image_helpers.py when generating the license block for the whole
image object.

The second change is to use the SHA1 checksum of a file instead of
SHA256 as the SPDX 2.2 spec requires SHA1 to be the hashing algorithm
to calculate the package verification code.

### DIFF
--- a/tern/formats/spdx/formats.py
+++ b/tern/formats/spdx/formats.py
@@ -7,6 +7,7 @@
 SPDX document formatting
 """
 
+import hashlib
 import uuid
 
 # basic strings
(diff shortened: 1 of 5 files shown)


rnjudge:
@nishakm Just curious -- did you check if the output from this image validated?

nishakm:
> @nishakm Just curious -- did you check if the output from this image validated?

You mean if the SPDX documents that are generated are now valid? I used the SPDX java tools to validate the text files produced for a few images.

PR #732, opened 2020-06-16
Generate report.txt when no output file specified

### LINKED ISSUE #731: report.txt not generated when no output file is specified
**Describe the bug**
The README tells the user to run `tern report -i debian:jessie` and `Take a look at report.txt to see what packages are installed in the Docker image and how Tern got this information.` Tern does not create report.txt, however, and instead prints this report information to the console. Additionally, `tern report --help` states:

```
 -o FILE, --output-file FILE
                        Write the report to a file. If no file is given the
                        default file in utils/constants.py will be used
```

but the default file (report.txt) is not used.  This is because of the following code in `report.py` which only writes the report to a file if an output file is specified, otherwise it prints it to the console. If we want to follow the current documentation guidelines, the `report_out` function should call write_report regardless if the `args.output_file` value exists or not.

```
151 def report_out(args, *images):                                                  
152     report = generate_report(args, *images)                                     
153     if not report:                                                              
154         logger.error("%s not a recognized plugin.", args.report_format)         
155     elif args.output_file:                                                      
156         write_report(report, args)                                              
157     else:                                                                       
158         print(report)  
```

If we don't want Tern to create a default report file (report.txt) when no output file is specified, we need to update our documentation accordingly. @nishakm do you have much of a preference either way? Personally, I would like to see a default report file generated instead of printed to the console. 

**To Reproduce**
1. run `tern report -i debian:jessie`
2. report.txt is not generated.

rnjudge:
Tern and its documentation tells users that if no output file is
specified, Tern will generate and populate a `report.txt` file. Due to how
the logic was structured in `report.py`, however, Tern was instead
printing the report output to the console and failing to generate a
`report.txt` file. This commit fixes the logic in report.py and generates
a default `report.txt` when no output file is specified.

Resolves #731

### DIFF
--- a/tern/report/report.py
+++ b/tern/report/report.py
@@ -152,7 +152,5 @@ def report_out(args, *images):
     report = generate_report(args, *images)
     if not report:
         logger.error("%s not a recognized plugin.", args.report_format)
-    elif args.output_file:
-        write_report(report, args)
     else:
-        print(report)
+        write_report(report, args)


nishakm:
@rnjudge Looks good! OK to merge?

rnjudge:
> @rnjudge Looks good! OK to merge?

Please! Thanks @nishakm

PR #740, opened 2020-06-23
Add HTML report format

abhaykatheria:
This PR adds an HTML format to tern. Now users can
generate an html tree view for their tern reports.
using -f html command line option.

In report/content.py a new function get_license_only
was added to get the license summary from an image list.
This function is now used by default and html format
generator functions.

### DIFF
--- a/tern/utils/constants.py
+++ b/tern/utils/constants.py
@@ -47,5 +47,7 @@ workdir = 'workdir'
 mergedir = 'mergedir'
 # report file
 report_file = 'report.txt'
+# html report file
+html_file = 'report.html'
 # locked dockerfile
 locked_dockerfile = 'Dockerfile.lock'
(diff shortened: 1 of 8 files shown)


tern/utils/constants.py
```diff
@@ -47,5 +47,7 @@
 mergedir = 'mergedir'
 # report file
 report_file = 'report.txt'
+# html report file
+html_file = 'report.html'
```
nishakm, line 51: This made me realize that the default output should be to console, but it looks like we do need a default file for certain output files. In this case, I wonder if this will work if we revert https://github.com/tern-tools/tern/pull/732?

rnjudge: No, it will not work if we revert #732. The way that Tern was implementing printing to an output file was either print to file specified by user OR print to console. We would need a new fix.

rnjudge: Some discussion in https://github.com/tern-tools/tern/issues/731 and proposal to merge this PR as-is since it adheres to current documentation around reporting.

nishakm: Sounds good to me

PR #742, opened 2020-06-24
Use container image layer index

### LINKED ISSUE #608: Change logging and reporting of layer checksums to the layer index
**Problem Statement**
Tern's logging and report checksum information is useful in identifying layers in the container image. However, the checksums do not actually match the checksums in the image manifest (they are tern's own checksums). These make it difficult to debug which layer brought in what packages and which layer metadata was not found

**Describe the Proposal**
Change the logging and reporting of the layers to reference the layer checksum from the container image manifest.

1. Create a method in the image_layer.py class to return the layer tarball's hash. This is the name of of the file in layer.tar_file.
2. Find all the places where layer.fs_hash[:10] is being used and use the new method instead.

rnjudge:
Tern's logging and reporting checksum information is useful in
identifying layers in a container image. However, the current checksums
do not match the checksums in the image manifest. Instead of using the
Tern-calculated filesystem hash, use the layer index. To do
this, introduce a new property in the image_layer class called
layer_index. Set the layer_index for each layer in the load_image()
function in docker_image.py. Lastly, replace the calls to
layer.fs_hash[:10] with the new layer_index property.

Resolves #608

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -288,7 +288,7 @@ def add_base_packages(image_layer, binary, shell):
         1. get the listing from the base.yml
         2. Invoke any commands against the base layer
         3. Make a list of packages and add them to the layer'''
-    origin_layer = 'Layer: ' + image_layer.fs_hash[:10]
+    origin_layer = 'Layer {}'.format(image_layer.layer_index)
     if image_layer.created_by:
         image_layer.origins.add_notice_to_origins(origin_layer, Notice(
             formats.layer_created_by.format(created_by=image_layer.created_by),
@@ -493,7 +493,7 @@ def add_snippet_packages(image_layer, command, pkg_listing, shell):
         4. For each unique package name, find the metadata and add to the
         layer'''
     # set up a notice origin for the layer
-    origin_layer = 'Layer: ' + image_layer.fs_hash[:10]
+    origin_layer = 'Layer {}'.format(image_layer.layer_index)
     # find packages for the command
     cmd_msg = formats.invoke_for_snippets + '\n' + \
         content.print_package_invoke(command.name)
@@ -554,7 +554,7 @@ def get_os_style(image_layer, binary):
     might be based off of given the pkg_format + package manager. If the binary
     provided does not exist in base.yml, add a warning notice'''
     origin_command_lib = formats.invoking_base_commands
-    origin_layer = 'Layer: ' + image_layer.fs_hash[:10]
+    origin_layer = 'Layer {}'.format(image_layer.layer_index)
     pkg_format = command_lib.check_pkg_format(binary)
     os_guess = command_lib.check_os_guess(binary)
     os_release = get_os_release(image_layer)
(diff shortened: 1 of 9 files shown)


tern/classes/image_layer.py
```diff
@@ -295,6 +296,15 @@ def gen_fs_hash(self):
...
 
+    def get_layer_hash(self):
+        '''Get the layer hash from the name of the directory that holds
+        layer.tar. This comes from the layer.tar_file property if the image
+        class was created with a tar_file.'''
+        if self.__tar_file:
+            # extract from <fs path>/tarball_hash/layer.tar
+            return self.__tar_file[:-64]
+        return ''
+
```
nishakm, line 307: I may have thought this was more straightforward than it looks. "layer.tar" is a Docker thing. OCI images have a different way of storing the layers i.e. their digest is in the config "layers" and they correspond to the hash of the tarball or whatever artifact. Could you move this function into the DockerImage class, and set the layer checksum property with the result of this and then replace `layer.fs_hash[:10]` with `layer.checksum`?

rnjudge: Had some discussion with Nisha about this. Instead of trying to pull the layer hash from the variety of container image client manifests (that all represent the hash in different ways/places) we will refer to the layers by order. We feel that this is more widely understood and intuitive.

tern/analyze/common.py
```diff
@@ -69,7 +69,7 @@ def load_packages_from_cache(layer):
     raw_pkg_list = cache.get_packages(layer.fs_hash)
     if raw_pkg_list:
         logger.debug(
-            'Loading packages from cache: layer \"%s\"', layer.fs_hash[:10])
+            'Loading packages from cache: layer \"%s\"', layer.layer_index)
```
nishakm, line 72: I would prefer not to have this changed because the cache is keyed based on the layer's `fs_hash`

tern/analyze/common.py
```diff
@@ -90,7 +90,7 @@ def load_files_from_cache(layer):
     raw_file_list = cache.get_files(layer.fs_hash)
     if raw_file_list:
         logger.debug(
-            'Loading files from cache: layer \"%s\"', layer.fs_hash[:10])
+            'Loading files from cache: layer \"%s\"', layer.layer_index)
```
nishakm, line 93: Same for this change - the cache is keyed based on the layer's `fs_hash`

tern/extensions/cve_bin_tool/executor.py
```diff
@@ -29,7 +29,7 @@ def execute(self, image_obj, redo=False):
         command = 'cve-bin-tool -x -u now'
         for layer in image_obj.layers:
             # execute the command for each layer
-            logger.debug("Analyzing layer %s", layer.fs_hash[:10])
+            logger.debug("Analyzing layer %s", layer.layer_index)
```
nishakm, line 32: This log message is fine because the layer is not cached.

tern/analyze/common.py
rnjudge: Fixed!

Issue #743, opened 2020-06-24
Find installed software from shell script


ForgetMe17:
**Description**
In pr #717 , we parse a shell script into statements. We need to fiter the install command and extact what will be installed in the command.

**To Do**
1. Filter the install command.
2. Find what is being installed by the install command.

**Background**
Previous work #682.

**Super Issues**
#521

ForgetMe17:
For a case mentioned in #744 .
```
apt-get install --no-install-recommends --no-install-suggests -y gnupg1 ca-certificates # INSTALL ca-certificates
...
...
apt-get remove --purge --auto-remove -y ca-certificates # REMOVE ca-certificates
```
`ca-certificates` should NOT be in the installed package list.
Besides, we may need to add `remove` to the remove subcommands of `apt-get`:
```
apt-get:
  install:
    - 'install' # subcommand to install
  remove:
    - 'purge' # subcommand to remove a package
  ignore: # list of subcommands that don't add or remove packages
    - 'update'
  packages: 'dpkg' # refer to base.yml's method of collection
```

ForgetMe17:
Closing with PR #756 merged.

PR #744, opened 2020-06-24
Find installed software from shell script

ForgetMe17:
This commit add a function find_installed_pkg_from_shell_script()
in tern/analyze/common.py which will filter the install commands
from the statements and find the installed software name.

Work towards #743.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -684,3 +684,33 @@ def get_git_toplevel(path):
     except (subprocess.CalledProcessError, FileNotFoundError):
         logger.debug("Cannot find git repo toplevel, path is %s", path)
     return path_to_toplevel
+
+
+def find_installed_pkg_from_shell_script(shell_script):
+    '''Given a shell script, first split it into statements, and then
+    filter the install commands to get the installed software.
+    Return a list of installing command objects and report on
+    ignored commands and unrecognized commands.'''
+    statements = general.split_shell_script(shell_script)
+    command_list = []
+    # traverse the statements, pick out the loop and commands.
+    for stat in statements:
+        if 'command' in stat:
+            command_list.append(Command(stat['command']))
+        elif 'loop' in stat:
+            loop_stat = stat['loop']['loop_statements']
+            for st in loop_stat:
+                if 'command' in st:
+                    command_list.append(Command(st['command']))
+    # filter commands
+    report = ''
+    if command_list:
+        for command in command_list:
+            command_lib.set_command_attrs(command)
+        ignore_msgs, filter1 = remove_ignored_commands(command_list)
+        unrec_msgs, filter2 = remove_unrecognized_commands(filter1)
+        if ignore_msgs:
+            report = report + formats.ignored + ignore_msgs
+        if unrec_msgs:
+            report = report + formats.unrecognized + unrec_msgs
+    return consolidate_commands(filter2), report
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -8,6 +8,7 @@ import random
 import re
 import regex
 import tarfile
+import shlex
 import subprocess  # nosec
 from contextlib import contextmanager
 from pathlib import Path
@@ -264,8 +265,8 @@ def parse_shell_variables_and_command(concatenated_command):
                                      'value': match_res.group(4)}
         statement['content'] = concatenated_command
     else:
-        # use clean_command() to clean tab and line indentations
-        statement['command'] = clean_command(concatenated_command)
+        # use shlex to clean tab, line indentations and long whitespaces.
+        statement['command'] = ' '.join(shlex.split(concatenated_command))
     return statement
 
 


ForgetMe17:
Here are the output for test dockerfiles.
[buildpack_deps_buster.txt](https://github.com/tern-tools/tern/files/4825338/buildpack_deps_buster.txt)
[buildpack_deps_buster_scm.txt](https://github.com/tern-tools/tern/files/4825339/buildpack_deps_buster_scm.txt)
[debian_buster_slim.txt](https://github.com/tern-tools/tern/files/4825340/debian_buster_slim.txt)

It works well with the install command `apt-get`, this command is in the `snippets.yml`.

tern/analyze/common.py
```diff
@@ -684,3 +684,31 @@ def get_git_toplevel(path):
...
+    for stat in statements:
+        if 'command' in stat:
+            command_list.append(Command(stat['command']))
+        elif 'loop' in stat:
+            loop_stat = stat['loop']['loop_statements']
+            for st in loop_stat:
+                if 'command' in st:
+                    command_list.append(Command(st['command']))
+    # filter commands
+    report = ''
```
ForgetMe17, line 704: I just paste the codes of `filter_install_commands` in `tern/analyze/common.py`. We can make modifications on the report here to give more specifice reports on the installed software.

ForgetMe17:
How to deal with commands `tar wget` needs to be discussed. We can record the file got by `wget` as `downloaded` if possible.

tern/analyze/common.py
nishakm: This looks good!

ForgetMe17: Actually the report is good enough to tell us the ignored commands and unrecognized commands. For example:
```
Ignored Commands:apt-get update

Unrecognized Commands:rm -rf /var/lib/apt/lists/*
```
Maybe there is no need to modify the report here.

tern/analyze/common.py
```diff
@@ -684,3 +684,33 @@ def get_git_toplevel(path):
...
+    if command_list:
+        for command in command_list:
+            command_lib.set_command_attrs(command)
+        ignore_msgs, filter1 = remove_ignored_commands(command_list)
+        unrec_msgs, filter2 = remove_unrecognized_commands(filter1)
+        if ignore_msgs:
+            report = report + formats.ignored + ignore_msgs
+        if unrec_msgs:
+            report = report + formats.unrecognized + unrec_msgs
+    return consolidate_commands(filter2), report
```
ForgetMe17, line 716: The first return value is a list of installing `Command objects`. We can get the installed software names by looking `words` property of `Command` class.

ForgetMe17:
I have a question here.
If we have a shell script like:
```
apt-get install --no-install-recommends --no-install-suggests -y gnupg1 ca-certificates
...
...
apt-get remove --purge --auto-remove -y ca-certificates
```
should `ca-certificates` be in our installed software list?

ForgetMe17:
In the new commit, i use shlex to clean commands because it can deal with long whitespaces.
If a long whitespaces is in the command like
`apt-get install -y curl            wget`
the `words` we can from parse_command() will be `['curl', ' ', ' ', ' ',  'wget']`.

rnjudge:
> I have a question here.
> If we have a shell script like:
> 
> ```
> apt-get install --no-install-recommends --no-install-suggests -y gnupg1 ca-certificates # INSTALL ca-certificates
> ...
> ...
> apt-get remove --purge --auto-remove -y ca-certificates # REMOVE ca-certificates
> ```
> 
> should `ca-certificates` be in our installed software list?

It think it depends if these commands are executed in the same RUN command. If they are executed in the same `RUN` command, then we wouldn't count ca-certificates in the installed software list as it gets removed in the same layer. If they are executed in different `RUN` commands then we would count ca-certificates as installed in the layer where it was installed but not in the layer where it was removed.

ForgetMe17:
Thanks for advice! I find that consolidate_commands() works for this purpose.

nishakm:
> I have a question here.
> If we have a shell script like:
> 
> ```
> apt-get install --no-install-recommends --no-install-suggests -y gnupg1 ca-certificates # INSTALL ca-certificates
> ...
> ...
> apt-get remove --purge --auto-remove -y ca-certificates # REMOVE ca-certificates
> ```
> 
> should `ca-certificates` be in our installed software list?

It shouldn't. `filter_installed_commands` should be able to remove the `remove` commands in `snippets.yml`. I think it's not removing it because `remove` is not listed as one of the remove subcommands:
```
apt-get:
  install:
    - 'install' # subcommand to install
  remove:
    - 'purge' # subcommand to remove a package
  ignore: # list of subcommands that don't add or remove packages
    - 'update'
  packages: 'dpkg' # refer to base.yml's method of collection
```

ForgetMe17:
Yes maybe we need to add `remove` in the remove subcommands.

ForgetMe17:
I have filed a new PR to update functions to implement finding installed software from shell script, and commented the question above to the issue #743 .
Closing.

PR #751, opened 2020-06-29
Remove raise immediately statements

### LINKED ISSUE #201: Revisit try-except-raise error dismisses in the code
Pylint try-except-raise error dismissals occur in the code at places where the error needs to bubble up to the top level executable which will log the error and either exit gracefully or soldier on. Investigate the "right" way to resolve these errors rather than dismissing them.

nishakm:
This commit removes raise only statements in try-except branches
found by pylint. Although the purpose was to re-raise the same
exception, it makes sense to be explicit about it. The change
basically replaces the empty raises with an explicit raising of
the same exception that the try branch brakes on.

Fixes #201

### DIFF
--- a/tern/analyze/docker/container.py
+++ b/tern/analyze/docker/container.py
@@ -179,8 +179,8 @@ def extract_image_metadata(image_tag_string):
             os.remove(placeholder)
         if not os.listdir(temp_path):
             raise IOError('Unable to untar Docker image')
-    except docker.errors.APIError:  # pylint: disable=try-except-raise
-        raise
+    except docker.errors.APIError as e:
+        raise docker.errors.APIError(e)
 
 
 def close_client():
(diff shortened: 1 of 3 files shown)


tern/analyze/docker/container.py
```diff
@@ -179,8 +179,8 @@ def extract_image_metadata(image_tag_string):
             os.remove(placeholder)
         if not os.listdir(temp_path):
             raise IOError('Unable to untar Docker image')
-    except docker.errors.APIError:  # pylint: disable=try-except-raise
-        raise
+    except docker.errors.APIError as e:
+        raise docker.errors.APIError(e)
```
rnjudge, line 183: `load_base_image()` and `load_full_image()` in `report.py` both call `load_image()` which calls `extract_image_metadata()` where we raise this error. Do we want to catch the `APIError` error specifically in `load_base_image()` and `load_full_image`?

rnjudge:
Looks good, just had one question about catching the APIError.

tern/analyze/docker/container.py
nishakm: It's not related to this fix, but I can submit a PR to add that change.

rnjudge: Thanks! I think it makes sense to catch it along with the others. In the meantime, I'll merge this.

Issue #695, opened 2020-05-20
Collect go package metadata for reports


rnjudge:
**Describe the Feature**
Currently, Tern does not lists the metadata of packages and modules installed using the golang package manager.

**Use Cases**
When images install go packages, it would be useful for Tern to report on the metadata of those packages.

**Implementation Changes**
Tern already supports adding snippets to the command library. It is possible that we might be able to add something to `snippets.yml` for package collection similar to how we collect other package information, but it will likely be more involved. This is something @abhaykatheria is working on for his GSoC project this summer.

nishakm:
Going to close this issue for now as Go modules seem to be the standardized way the community is heading.

PR #754, opened 2020-07-01
Enable golang package listing

abhaykatheria:
This commit adds commands in snippets.yml and base.yml to
to capture golang module name, version, proj_url and copyright
metadata in the output report. This commit does not currently
collect license information for golang modules as more advanced
parsing is required but functionality for this will come in
the future.

Work towards #695

Signed-off-by: abhaykatheria <abhay.katheria1998@gmail.com>

### DIFF
--- a/tern/command_lib/base.yml
+++ b/tern/command_lib/base.yml
@@ -379,3 +379,35 @@ npm:
           - "pkgs=`npm list -g | sed 's/\\(^[^a-z]*\\)\\([_.a-z0-9-]*\\)\\(@.*\\)/\\2/g' | grep -v '/usr*' | uniq`"
           - "for p in $pkgs; do npm view $p repository.url; done"
     delimiter: "\n"
+# golang----------------------------------------------------------------------
+go:
+  pkg_format: ''
+  os_guess:
+    - 'None'
+  path:
+    - 'usr/local/go/bin'
+  names:
+    invoke:
+      1:  
+        container:
+          - "go list -m all | tail -n +2 | cut -d ' ' -f 1" 
+    delimiter: "\n"
+  versions:
+    invoke:
+      1:
+        container:
+          - "go list -m all | tail -n +2 | cut -d ' ' -f 2"
+    delimiter: "\n"
+  copyrights:
+    invoke:
+      1:
+        container:
+          - "dirs=`go list -m -f={{{{.Dir}}}} all | tail -n+2`"
+          - "for d in $dirs; do find $d -type f -iname 'License' -o -iname 'license.txt' -o -iname 'license.md' | xargs cat; echo LICF; done"
+    delimiter: "LICF"
+  proj_urls:
+    invoke:
+      1:
+        container:
+          - "go list -m all | tail -n +2 | cut -d ' ' -f 1"
+    delimiter: "\n"
--- a/tern/command_lib/snippets.yml
+++ b/tern/command_lib/snippets.yml
@@ -178,3 +178,13 @@ microdnf:
     - 'check-update'
     - 'clean'
   packages: 'rpm'
+
+go:
+  install:
+    - 'build'
+    - 'mod'
+  remove:
+    - 'remove'
+  ignore:
+    - 'clean'
+  packages: 'go'


nishakm:
@abhaykatheria I tested the scripts against other listings and they seem to run fine.

It looks like your scripts may need some tweaking:
```
$ python tern/tools/verify_invoke.py --image debos:buildstage --keys base go names
[sudo] password for nisha:
Command failed. /bin/sh: 1: Syntax error: Unterminated quoted string <<--- check your listing command

Error executing command in chroot
Traceback (most recent call last):
  File "/home/nisha/terndev/tern/tern/utils/rootfs.py", line 236, in run_chroot_command
    shell, '-c', command_string)
  File "/home/nisha/terndev/tern/tern/utils/rootfs.py", line 77, in root_command
    1, cmd=full_cmd, output=None, stderr=error.decode())
subprocess.CalledProcessError: Command '['sudo', 'unshare', '-pf', '--mount-proc=/home/nisha/.tern/temp/mergedir/proc', 'chroot', '/home/nisha/.tern/temp/mergedir', '/bin/sh', '-c', "go list -m all | tail -n +2 | cut -d ' ' -f 1' "]' returned non-zero exit status 1. <-- you may have an extra quote here.
```

abhaykatheria:
```
# Dockerfile References: https://docs.docker.com/engine/reference/builder/

# Start from the latest golang base image
FROM golang:1.14-alpine

# Set the Current Working Directory inside the container
WORKDIR /app

# Copy go mod and sum files
COPY go.mod go.sum ./

# Download all dependencies. Dependencies will be cached if the go.mod and go.sum files are not changed
RUN go mod download

# Copy the source from the current directory to the Working Directory inside the container
COPY . .

# Build the Go app
RUN CGO_ENABLED=0 GOOS=linux go build -v -a -installsuffix cgo -o swag cmd/swag/main.go
```
I am testing on this docker file.
The original docker file can be found [here](https://github.com/swaggo/swag)

tern/command_lib/base.yml
```diff
@@ -379,3 +379,41 @@ npm:
...
+  pkg_format: ''
+  os_guess:
+    - 'None'
+  path:
+    - 'usr/local/go/bin'
+  pkg_separators:
+    - '@'
+    - '@^'
+    - '@~'
+  pinning_separator: ':'
```
rnjudge, line 393: Reminder that we don't need these for go

tern/command_lib/base.yml
```diff
@@ -379,3 +379,41 @@ npm:
...
+    invoke:
+      1:
+        container:
+          - "/usr/local/go/bin/go list -m all | tail -n +2 | cut -d ' ' -f 2"
+    delimiter: "\n"
+  copyrights:
+    invoke:
+      1:
+        container:
+          - "export GOPATH=/go"
```
rnjudge, line 410: While the default `$GOPATH` environment variable for the docker-library golang image is `/go`, it can be changed. I propose setting this environment variable based on the value of the `GOPATH` value in the config and if it's not available there, using `/go` as the default. I will open a PR for this shortly.

rnjudge: Opened https://github.com/tern-tools/tern/pull/780

tern/command_lib/base.yml
```diff
@@ -379,3 +379,41 @@ npm:
...
+  pkg_separators:
+    - '@'
+    - '@^'
+    - '@~'
+  pinning_separator: ':'
+  names:
+    invoke:
+      1:  
+        container:
+          - "/usr/local/go/bin/go list -m all | tail -n +2 | cut -d ' ' -f 1"
```
rnjudge, line 398: Since https://github.com/tern-tools/tern/pull/780 will now set the PATH when it exists, we no longer need to call the go command using `/usr/local/go/bin/go`. We can just use `go` now.

tern/command_lib/base.yml
```diff
@@ -379,3 +379,41 @@ npm:
...
+        container:
+          - "/usr/local/go/bin/go list -m all | tail -n +2 | cut -d ' ' -f 2"
+    delimiter: "\n"
+  copyrights:
+    invoke:
+      1:
+        container:
+          - "export GOPATH=/go"
+          - "dirs=`/usr/local/go/bin/go list -m -f={{{{.Dir}}}} all | tail -n+2`"
+          - "for d in $dirs;do find $d -type f -iname 'License' -o -iname 'license.txt' -o -iname 'license.md' | xargs cat; echo LICF ; done;"
```
rnjudge, line 412: small nit: semi-colons should hug the character before and there should be a space after the semi-colon before the proceeding character. This line needs a few fixes -- 1) `for d in $dirs; do` and 2) `echo LICF; done`

tern/command_lib/base.yml
```diff
@@ -379,3 +379,35 @@ npm:
...
+      1:
+        container:
+          - "dirs=`go list -m -f={{{{.Dir}}}} all | tail -n+2`"
+          - "for d in $dirs; do find $d -type f -iname 'License' -o -iname 'license.txt' -o -iname 'license.md' | xargs cat; echo LICF; done"
+    delimiter: "LICF"
+  proj_urls:
+    invoke:
+      1:
+        container:
+          - "/usr/local/go/bin/go list -m all | tail -n +2 | cut -d ' ' -f 1"
```
rnjudge, line 412: We can just use `go` here :)

PR #756, opened 2020-07-07
Update functions to use the shell script parser

ForgetMe17:
This PR updates functions `split_command()`, `get_shell_commands()` and `clean_command()` to implement a shell script parser for RUN commands.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -32,11 +32,18 @@ logger = logging.getLogger(constants.logger_name)
 
 def get_shell_commands(shell_command_line):
     '''Given a shell command line, get a list of Command objects'''
-    comm_list = general.split_command(shell_command_line)
-    cleaned_list = []
-    for comm in comm_list:
-        cleaned_list.append(Command(general.clean_command(comm)))
-    return cleaned_list
+    statements = general.split_command(shell_command_line)
+    command_list = []
+    # traverse the statements, pick out the loop and commands.
+    for stat in statements:
+        if 'command' in stat:
+            command_list.append(Command(stat['command']))
+        elif 'loop' in stat:
+            loop_stat = stat['loop']['loop_statements']
+            for st in loop_stat:
+                if 'command' in st:
+                    command_list.append(Command(st['command']))
+    return command_list
 
 
 def load_from_cache(layer, redo=False):
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -8,6 +8,7 @@ import random
 import re
 import regex
 import tarfile
+import shlex
 import subprocess  # nosec
 from contextlib import contextmanager
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


ForgetMe17:
In the code check, it failed on Prospector_Linting which says:
```
tern/analyze/common.py
  Line: 196
    pep8: E741 / ambiguous variable name 'l' (col 9)
    pep8: E741 / ambiguous variable name 'l' (col 9)
```
it seems that using `l` for the variable name will casue this problem.

tern/utils/general.py
```diff
@@ -45,15 +46,123 @@ def initialize_names():
...
-    return re.sub(cleaning, '', command).strip()
+    '''Given a command string (only contains one command, does not contain &&
+    or ;), clean out all whitespaces, tabs and line indentations'''
+    return ' '.join(shlex.split(command))
 
 
-def split_command(command):
-    '''Given a string of concatenated commands, return a list of commands'''
-    return re.split(concat, command)
+def split_command(shell_script):
```
ForgetMe17, line 54: i move the function split_shell_script() here.

tern/utils/general.py
```diff
@@ -45,15 +46,123 @@ def initialize_names():
 
 
 def clean_command(command):
-    '''Given a command string, clean out all whitespaces, tabs and line
-    indentations
-    Leave && alone'''
-    return re.sub(cleaning, '', command).strip()
+    '''Given a command string (only contains one command, does not contain &&
+    or ;), clean out all whitespaces, tabs and line indentations'''
+    return ' '.join(shlex.split(command))
```
ForgetMe17, line 51: Use `shlex` to clean commands instead of `re.sub`. `shlex` can clean long whitespaces.

tern/analyze/common.py
```diff
@@ -32,11 +32,18 @@
 
 def get_shell_commands(shell_command_line):
     '''Given a shell command line, get a list of Command objects'''
-    comm_list = general.split_command(shell_command_line)
-    cleaned_list = []
-    for comm in comm_list:
-        cleaned_list.append(Command(general.clean_command(comm)))
-    return cleaned_list
+    statements = general.split_command(shell_command_line)
```
ForgetMe17, line 35: Here we use the shell script parser to get statements, and then pick out `commands` and `commands in a loop`. From each `command` statement, we append it to the `command_list`.

tern/utils/general.py
nishakm: > In the code check, it failed on Prospector_Linting which says:
> 
> ```
> tern/analyze/common.py
>   Line: 196
>     pep8: E741 / ambiguous variable name 'l' (col 9)
>     pep8: E741 / ambiguous variable name 'l' (col 9)
> ```
> 
> it seems that using `l` for the variable name will casue this problem.

Looks like prospector version 1.3 checks for [letters that look like numbers](https://www.flake8rules.com/rules/E741.html). I will submit a fix and rerun the build.

nishakm:
@ForgetMe17 I tested your changes and they look good! So I will try to fix the build.

nishakm:
@ForgetMe17 can you rebase your PR? I think that will fix your build.

ForgetMe17:
> @ForgetMe17 can you rebase your PR? I think that will fix your build.

Yes!

PR #764, opened 2020-07-11
Add report for branch statement

### LINKED ISSUE #763: Report RUN command statements that are non-deterministic like if and case statements
**Describe the Feature**
Currently, the report notices just says that if, case and for statements are unrecognizable. It would be nice to add notices for non-deterministic branching statements like if and case statements as you could only know the status of the branch at build time.

**Implementation Changes**
The new shell script parser produces a dictionary that will identify if and case statements. We could just look at this object and extract the if and case statements to create notices for them.

ForgetMe17:
Previously, we just ignore the branch statement during parsing, this
commit add report for branch statement. The report will have a new
section called 'Non-deterministic branching statement' if there is a
branch in the RUN command.

Modified get_shell_commands() in tern\analyze\common.py.

Resolves #763.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -31,9 +31,11 @@ logger = logging.getLogger(constants.logger_name)
 
 
 def get_shell_commands(shell_command_line):
-    '''Given a shell command line, get a list of Command objects'''
+    '''Given a shell command line, get a list of Command objects and report on
+    branch statements'''
     statements = general.split_command(shell_command_line)
     command_list = []
+    branch_report = ''
     # traverse the statements, pick out the loop and commands.
     for stat in statements:
         if 'command' in stat:
@@ -43,7 +45,13 @@ def get_shell_commands(shell_command_line):
             for st in loop_stat:
                 if 'command' in st:
                     command_list.append(Command(st['command']))
-    return command_list
+        elif 'branch' in stat:
+            branch_report = branch_report + '\n'.join(stat['content']) + '\n\n'
+    if branch_report:
+        # add prefix
+        branch_report = '\nNon-deterministic branching statement: \n' + \
+                        branch_report
+    return command_list, branch_report
 
 
 def load_from_cache(layer, redo=False):
@@ -478,7 +486,7 @@ def filter_install_commands(shell_command_line):
         3. Return installed command objects, and messages for ignored commands
         and unrecognized commands'''
     report = ''
-    command_list = get_shell_commands(shell_command_line)
+    command_list, branch_report = get_shell_commands(shell_command_line)
     for command in command_list:
         command_lib.set_command_attrs(command)
     ignore_msgs, filter1 = remove_ignored_commands(command_list)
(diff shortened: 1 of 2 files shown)


tern/analyze/common.py
```diff
@@ -43,7 +45,13 @@ def get_shell_commands(shell_command_line):
             for st in loop_stat:
                 if 'command' in st:
                     command_list.append(Command(st['command']))
-    return command_list
+        elif 'branch' in stat:
+            branch_report = branch_report + '\n'.join(stat['content']) + '\n\n'
```
ForgetMe17, line 49: For branch statements, we just record its content in the report. Like:
```
Unrecognized Commands:set -eux
wget -O go.tgz $url
echo ${goRelSha256} *go.tgz | sha256sum -c -
tar -C /usr/local -xzf go.tgz
rm go.tgz
go version

Non-deterministic branching statement: 
case "${dpkgArch##*-}" in 		amd64) goRelArch='linux-amd64'
goRelSha256='08df79b46b0adf498ea9f320a0f23d6ec59e9003660b4c9c1ce8e5e2c6f823ca'
armhf) goRelArch='linux-armv6l'
goRelSha256='b5e682176d7ad3944404619a39b585453a740a2f82683e789f4279ec285b7ecd'
arm64) goRelArch='linux-arm64'
goRelSha256='cd813387f770c07819912f8ff4b9796a4e317dee92548b7226a19e60ac79eb27'
i386) goRelArch='linux-386'
goRelSha256='cdcdab6c8d1f2dcea3bbec793352ef84db167a2eb6c60ff69e5cf94dca575f9a'
ppc64el) goRelArch='linux-ppc64le'
goRelSha256='b896b5eba616d27fd3bb8218de6bef557cb62221e42f73c84ae4b89cdb602dec'
s390x) goRelArch='linux-s390x'
goRelSha256='22e67470fe872c893face196f02323a11ffe89999260c136b9c50f06619e0243'
*) goRelArch='src'
goRelSha256='6d643e46ad565058c7a39dac01144172ef9bd476521f42148be59249e4b74389'
echo >&2
echo >&2 "warning: current architecture ($dpkgArch) does not have a corresponding Go binary release; will be building from source"
echo >&2
esac

if [ "$goRelArch" = 'src' ]
then 		echo >&2
echo >&2 'error: UNIMPLEMENTED'
echo >&2 'TODO install golang-any from jessie-backports for GOROOT_BOOTSTRAP (and uninstall after build)'
echo >&2
exit 1
fi

```

ForgetMe17:
Here are my local test output. 
[buildpack_deps_buster.txt](https://github.com/tern-tools/tern/files/4906091/buildpack_deps_buster.txt)
[buildpack_deps_buster_scm.txt](https://github.com/tern-tools/tern/files/4906092/buildpack_deps_buster_scm.txt)
[debian_buster_slim.txt](https://github.com/tern-tools/tern/files/4906093/debian_buster_slim.txt)

tests/test_analyze_common.py
```diff
@@ -32,7 +32,7 @@ def tearDown(self):
         del self.test_dockerfile
 
     def testGetShellCommands(self):
-        command = common.get_shell_commands("yum install nfs-utils")
+        command, _ = common.get_shell_commands("yum install nfs-utils")
```
nishakm, line 35: If you can add a test here for a branching command such as `if [ -z $var ]; then do something; fi`

ForgetMe17:
I have made small nits on the test.

PR #771, opened 2020-07-17
Enable Tern to run in a container for non-Linux host

### LINKED ISSUE #679: Do not use overlayfs to debug container layers
**Describe the Feature**
Since tern uses the overlayfs driver to analyze packages layer by layer, it cannot run within a container image without an external mount to a Linux host. Not all hosts are Linux and Docker runs on MacOS and Windows. Hence using an alternative for "debugging" container image filesystems would allow the full operation to take place in a running container.

**Use Cases**
This is so developers using Windows and MacOS can run tern without needing to install Vagrant.

**Implementation Changes**
We have to experiment with other methods of combining layers other than using a CoW storage driver.

rnjudge:
Tern does not currently run within a container image without an external
mount to a linux host. This pull request updates the code as well as the
Dockerfile and docker_run.sh script to install fuse-overlayfs and use it to
mount the diff layers when the --non-linux command line option is
envoked. These changes will enable Tern to run in a container on a
non-Linux host instead of having to use vagrant.

Resolves #679 

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,19 +1,17 @@
-# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-FROM photon:3.0
+FROM debian:buster
 
-# install system dependencies
-# photon:3.0 comes with toybox which conflicts with some dependencies needed for tern to work, so uninstalling
-# toybox first
-RUN tdnf remove -y toybox && tdnf install -y tar findutils attr util-linux python3 python3-pip python3-setuptools git && pip3 install --upgrade pip
+# Install fuse-overlayfs and Tern dependencies
+RUN apt-get update && apt-get -y install wget gnupg2 tar findutils attr util-linux python3 python3-pip python3-setuptools git && pip3 install --upgrade pip && echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_10/Release.key -O Release.key && apt-key add - < Release.key && apt-get update -qq && apt-get -qq -y install buildah fuse-overlayfs && rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
-# install tern with latest changes
+# Install tern with latest changes
 COPY dist/tern-*.tar.gz .
 RUN pip install tern-*.tar.gz
 
-# make a mounting directory
-RUN mkdir hostmount
+# Adjust storage.conf to enable Fuse storage.
+RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
 
-ENTRYPOINT ["tern", "-q", "-b", "/hostmount"]
+ENTRYPOINT ["tern", "--driver", "fuse"]
 CMD ["-h"]
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -71,8 +71,8 @@ def create_top_dir(working_dir=None):
 
 def do_main(args):
     '''Execute according to subcommands'''
-    # set bind mount location if working in a container
-    rootfs.set_mount_dir(args.bind_mount, args.working_dir)
+    # Set mount/working dir according to user
+    rootfs.set_working_dir(args.working_dir)
     # create working directory
     create_top_dir(args.working_dir)
     if not args.quiet:
@@ -131,14 +131,17 @@ def main():
     parser.add_argument('-k', '--keep-wd', action='store_true',
                         help="Keep the working directory after execution."
                         " Useful when debugging container images")
-    parser.add_argument('-b', '--bind-mount', metavar='BIND_DIR',
-                        help="Absolute path to bind mount target. Needed"
-                        " when running from within a container.")
     parser.add_argument('-r', '--redo', action='store_true',
                         help="Repopulate the cache for found layers")
     parser.add_argument('-wd', '--working-dir', metavar='PATH',
                         help="Change default working directory to specified"
                         "absolute path.")
+    parser.add_argument('-d', '--driver', metavar="DRIVER_OPTION",
+                        help="Required when running Tern in a container."
+                        "Using 'fuse' will enable the fuse-overlayfs driver "
+                        "to mount the diff layers of the container. If no "
+                        "input is provided, 'fuse' will be used as the "
+                        "default option.")
     # sys.version gives more information than we care to print
     py_ver = sys.version.replace('\n', '').split('[')[0]
     parser.add_argument('-v', '--version', action='version',
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,13 @@
-# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-FROM photon:3.0
+FROM debian:buster
 
-# install system dependencies
-# photon:3.0 comes with toybox which conflicts with some dependencies needed for tern to work, so uninstalling
-# toybox first
-RUN tdnf remove -y toybox && tdnf install -y tar findutils attr util-linux python3 python3-pip python3-setuptools git && pip3 install --upgrade pip && pip3 install tern
+# Install fuse-overlayfs and Tern dependencies
+RUN apt-get update && apt-get -y install wget gnupg2 tar findutils attr util-linux python3 python3-pip python3-setuptools git && pip3 install --upgrade pip && pip3 install tern && echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_10/Release.key -O Release.key && apt-key add - < Release.key && apt-get update -qq && apt-get -qq -y install buildah fuse-overlayfs && rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
-# make a mounting directory
-RUN mkdir hostmount
+# Adjust storage.conf to enable Fuse storage.
+RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
 
-ENTRYPOINT ["tern", "-q", "-b", "/hostmount"]
+ENTRYPOINT ["tern", "--driver", "fuse"]
 CMD ["-h"]
--- a/README.md
+++ b/README.md
@@ -108,25 +108,34 @@ Build the Docker image (called `ternd` here). You may need to use sudo:
 $ docker build -t ternd .
 ```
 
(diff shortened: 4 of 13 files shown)


rnjudge:
Also should note that I am full-heartedly open to changing the name of the command line argument from "non-linux" to something less clunky. Suggestions welcome :)

rnjudge:
I will re-base and re-run the tests once https://github.com/tern-tools/tern/pull/773 is merged. In the mean time, the review can still continue :)

nishakm:
In commit 1:
```
 Enable Terns run in a container for non-Linux host -> suggest Enable tern to run or just Run in a container for non-linux hosts

When that option is envoked -> s/envoked/invoked
```

rnjudge:
> In commit 1:
> 
> ```
>  Enable Terns run in a container for non-Linux host -> suggest Enable tern to run or just Run in a container for non-linux hosts
> 
> When that option is envoked -> s/envoked/invoked
> ```

Heh, thank you!

tern/__main__.py
```diff
@@ -72,7 +72,7 @@ def create_top_dir(working_dir=None):
 def do_main(args):
     '''Execute according to subcommands'''
     # set bind mount location if working in a container
-    rootfs.set_mount_dir(args.bind_mount, args.working_dir)
+    rootfs.set_mount_dir(args.non_linux, args.bind_mount, args.working_dir)
```
nishakm, line 75: I think we can consolidate these options:
1. args.working_dir and args.bind_mount mean essentially the same thing here - make the working directory different from the default. So I think the `args.bind_mount` option should be deprecated. Either way, we are changing the CLI. I am not sure if this kind of change requires a major release change. Thoughts?
2. I envision that in the future, we will need to support running tern in an unprivileged container. In that sense, `non_linux` may not be a good name for this option. How about `--driver` and give options `fuse, overlay` and make `overlay` the default? That way we can add the `rsync` or whatever else option there may be to run tern in an unprivileged way.

tern/__main__.py
```diff
@@ -139,6 +139,10 @@ def main():
     parser.add_argument('-wd', '--working-dir', metavar='PATH',
                         help="Change default working directory to specified"
                         "absolute path.")
+    parser.add_argument('-n', '--non-linux', action='store_true',
```
nishakm, line 142: So if we were to choose `--driver` this would not be a `store_true` action.

ci/Dockerfile
```diff
@@ -1,19 +1,17 @@
...
+# Install tern with latest changes
 COPY dist/tern-*.tar.gz .
 RUN pip install tern-*.tar.gz
 
-# make a mounting directory
-RUN mkdir hostmount
+# Adjust storage.conf to enable Fuse storage.
+RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
 
-ENTRYPOINT ["tern", "-q", "-b", "/hostmount"]
```
nishakm, line 18: If quiet more is not turned on here, maybe there should be some documentation showing how to turn off logging when running in container mode?

rnjudge: Agree. I can add that to do the README.

tern/__main__.py
rnjudge: 1) I would agree that deprecating the bind mount option and combining it into a `--driver` option should not constitute a major release. Especially because this CLI change is in the Dockerfile and I think most people build the Dockerfile provided in the repository, or copy the Dockerfile and slightly modify it, which won't affect what they have to interact with the CLI options.

2) I like the `--driver` suggestion. I will incorporate this.

rnjudge: Combining the `bind_mount` and `non_linux` option will require the `--driver` option inputs to be a little funky. Folks will have to use either `--driver fuse` (which will use the default location inside the container) or `--driver /location/to/bind/mount`. @nishakm Does this sound ok?

nishakm: OK so here's my reasoning:
We have potentially 3 modes of operation and hence 3 ways a user can invoke tern:
| Natively on Linux  | In a privileged container | In an unprivileged container  |
|-------------------------|----------------------------------|---------------------------------------|
| --driver overlayfs  | --driver fuse                     | --driver rsync (or whatever)   |
| deps: none           | deps: fuse-overlayfs        | deps: whatever is required    |
| no external mnt    | external mnt if driver is overlayfs | no external mnt         |

In this situation I think avoiding the added complexity of checking to see if the driver choice is overlayfs and you are running in a container is a bit redundant because if you were running on a linux host, you could just tern natively. But if you were running in a container, it's less effort to just install fuse-overlay and not use a bind mount at all. This is why I think we should get rid of that option altogether.

nishakm: We could still allow the `-wd` option if someone wanted to point the working directory to a place other than the user's home directory.

rnjudge: Ok, totally with you on getting rid of the `bind_mount` option. My question was more about having a --driver overlayfs option in addition to `--driver fuse`. If we are running in a container (regardless of Linux or non-linux host) and we just use the fuse-overlay to do the mounting, why does the user need to specify `--driver overlay` or `--driver fuse`? In either case, if we use fuse-overlayfs, we will be mounting inside the container.

nishakm: Oh I see what you're saying. We can default to the overlayfs mounting unless given the `--driver` option which can either be `fuse` or something else. I'm with you on this one. `overlayfs` need not be one of the options :+1:

rnjudge: Cool! I will note in the "help" for the `--driver` CLI option that right now the only supported option for `--driver` is `fuse` but once we support non-privlidged containers we can change that note to expand for rsync, or whatever we want to call it.

rnjudge:
> I had one question about turning on logging. Typically, when running the container as a service, the logging is not desirable in the output report.

The logging won't be in the output report, it would jut log to the terminal. If someone is running this with a CI/CD pipeline in I would assume the output could be ignored. This logging behavior is also consistent with Tern's default operation. Until the Beta release this also seemed like a good idea for users who might be using Tern in a container so they can more easily log bugs or issues that might arise while Tern runs without having to modify the Dockerfile to enable logging.

rnjudge:
I've tested this in a Mac environment and it works. @tevoinea would you be willing to pull and test these changes in a Windows environment using an image with multiple layers? In order to pull and test you will need to do the following:

```
$ git clone -b remove-overlay git@github.com:rnjudge/tern.git
$ cd tern
$ python setup.py sdist
$ cp ci/Dockerfile .
$ docker build -t ternd .
$ ./docker_run.sh ternd:latest "report -i <your_multi_layer_image>" > output.txt
```

Thanks in advance!

tevoinea:
Thanks for looking into this @rnjudge !

I had to modify the steps you listed above a little bit to work in my environment but the result should be the same. I did:

```powershell
git clone https://github.com/rnjudge/tern.git
cd tern
git checkout remove-overlay
python setup.py sdist
cp ci/Dockerfile .
docker build -t ternd .
docker run --privileged --device /dev/fuse -v /var/run/docker.sock:/var/run/docker.sock --rm ternd:latest report -i golang:latest > output.txt
```

The git commands shouldn't make a difference (I hadn't set up ssh git), and I had to manually run the script inside docker_run.sh to have it run in powershell.

As for the result, I got the following shell output:
```shell
docker run --privileged --device /dev/fuse -v /var/run/docker.sock:/var/run/docker.sock --rm ternd:latest report -i golang:latest > output.txt
2020-07-28 19:46:31,882 - DEBUG - __main__ - Starting...
2020-07-28 19:46:31,882 - DEBUG - run - Setting up...
2020-07-28 19:46:31,886 - DEBUG - container - Checking if image "golang:latest" is available on disk...
2020-07-28 19:46:31,890 - DEBUG - container - Image "golang:latest" found
2020-07-28 19:46:31,891 - DEBUG - container - Checking if image "golang:latest" is available on disk...
2020-07-28 19:46:31,894 - DEBUG - container - Image "golang:latest" found
2020-07-28 19:46:48,648 - DEBUG - rootfs - Running command: tar -tf /root/.tern/temp.tar
2020-07-28 19:46:48,652 - DEBUG - rootfs - Running command: tar -x --exclude=.wh.* -f /root/.tern/temp.tar -C /root/.tern/temp
2020-07-28 19:46:49,913 - DEBUG - rootfs - Running command: tar -tf /root/.tern/temp/5cf51fecf2f54460092bd8ac9dd41d6448b4c3e8bd8a930f1b39f2ca7ddf147c/layer.tar
2020-07-28 19:46:49,956 - DEBUG - rootfs - Running command: tar -x --exclude=.wh.* -f /root/.tern/temp/5cf51fecf2f54460092bd8ac9dd41d6448b4c3e8bd8a930f1b39f2ca7ddf147c/layer.tar -C /root/.tern/temp/5cf51fecf2f54460092bd8ac9dd41d6448b4c3e8bd8a930f1b39f2ca7ddf147c/contents
2020-07-28 19:46:50,485 - DEBUG - rootfs - Running command: /usr/local/lib/python3.7/dist-packages/tern/tools/fs_hash.sh /root/.tern/temp/5cf51fecf2f54460092bd8ac9dd41d6448b4c3e8bd8a930f1b39f2ca7ddf147c/contents
2020-07-28 19:46:50,487 - WARNING - report - Error in loading image: [Errno 13] Permission denied: '/usr/local/lib/python3.7/dist-packages/tern/tools/fs_hash.sh'
2020-07-28 19:46:50,487 - WARNING - run - Cannot retrieve full image metadata
2020-07-28 19:46:50,487 - DEBUG - run - Teardown...
2020-07-28 19:46:50,488 - DEBUG - rootfs - Running command: rm -rf /root/.tern/temp/mergedir
2020-07-28 19:46:50,490 - DEBUG - rootfs - Running command: rm -rf /root/.tern/temp/workdir
2020-07-28 19:46:50,745 - DEBUG - __main__ - Report completed.
2020-07-28 19:46:50,746 - DEBUG - __main__ - Finished
```
 and output.txt was empty `-a----         7/28/2020     12:46              0 output.txt`

rnjudge:
@tevoinea do you have root permissions in your Windows environment?

tevoinea:
@rnjudge I do, I also tried running it in an administrator elevated powershell as well and it's the same output.

rnjudge:
@tevoinea Thank you! Let me take a look and get back to you.

rnjudge:
The permissions error is unique to Windows and should most likely be addressed in another PR.

rnjudge:
The permissions issue is unique to Windows but we can work around this by running `chmod +x` the `fs_hash.sh` script.

@tevoinea Could you delete the tern directory that you previously checked out and run `git config --global core.autocrlf input` BEFORE re-cloning the changes (from the same branch) and re-building the tern image to test? I just did this on a Windows VM using VMware Fusion and was able to generate a tern output.

tevoinea:
@rnjudge It worked! This is awesome 😁

I noticed a small problem with the file output when using the `-f json` and `-o output.json` flags though. Here's the command I ran:

`docker run --privileged --device /dev/fuse -v /var/run/docker.sock:/var/run/docker.sock --rm ternd:latest report -f json -o D:\tern\output.json -i golang:latest`

It doesn't create the output.json file. However just the `-f json` flag with `> output.json` works perfectly.

rnjudge:
> @rnjudge It worked! This is awesome 😁
> 
> I noticed a small problem with the file output when using the `-f json` and `-o output.json` flags though. Here's the command I ran:
> 
> `docker run --privileged --device /dev/fuse -v /var/run/docker.sock:/var/run/docker.sock --rm ternd:latest report -f json -o D:\tern\output.json -i golang:latest`
> 
> It doesn't create the output.json file. However just the `-f json` flag with `> output.json` works perfectly.

@tevoinea Yes, that behavior is because Tern is running in a container. The `-o` flag works when you run natively in a Linux environment. When running in a container, the `-o` flag will put the output file in the running container just before it is removed and exits. In your case, the output file can be redirected using `>` to whatever location you would normally use the -o flag with (i.e. `> D:\tern\output.json`). Hope this works!

tevoinea:
@rnjudge Ah, I understand! That will work just fine 😊

PR #773, opened 2020-07-18
Bug fix with Dockerfile RUN parsing

ForgetMe17:
Fixed Tern exits with 'killed' when Dockerfile RUN statement is complex.
Changed logic in the function consolidate_commands() in
tern/analyze/common.py.

Fixes tern-tools#772.

### DIFF
--- a/tern/analyze/common.py
+++ b/tern/analyze/common.py
@@ -465,6 +465,7 @@ def consolidate_commands(command_list):
         new_list.append(command_list.pop(0))
 
     while command_list:
+        # match the first command with its following commands.
         first = command_list.pop(0)
         for _ in range(0, len(command_list)):
             second = command_list.pop(0)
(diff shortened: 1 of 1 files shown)


tern/analyze/common.py
```diff
@@ -473,8 +474,11 @@ def consolidate_commands(command_list):
                 new_list.append(second)
             else:
                 if not first.merge(second):
-                    command_list.append(first)
```
ForgetMe17, line 476: append(first) here may cause two or more same Command objects in the `command_list`.
It seems more reasonable to append `second` since we have popped `second` from the list, if we cannot merge `second` into `first`, we need to keep `second` in the list instead of `first` for next while loop to analyze it.

nishakm:
@rnjudge Did you test this change on top of your changes to see if it works?

rnjudge:
> @rnjudge Did you test this change on top of your changes to see if it works?

Yes! I get a report for the more complex Dockefile. Dockerfile lock works too with the more complex file. Changes look good. Nice work @ForgetMe17!

Issue #612, opened 2020-03-27
Perform analysis on multistage Dockerfiles


nishakm:
**Describe the Feature**
Tern can read a Dockerfile, build and image and analyze the image to provide a report and pin a Dockerfile. However, it cannot do these for multistage Dockerfiles accurately because the intermediate stages get thrown away after a build is complete. Why not enable analysis and pinning for multistage Dockerfiles?

**Use Cases**
Multistage Dockerfiles are the de-facto build mechanism for golang projects in particular. A Dockerfile will typically use the golang image to build and an alpine image to deploy the built golang binary.

**Implementation Changes**
A proposed implementation can look like this:
1. Split the Dockerfile by stage, making a single Dockerfile for each stage
2. Build and analyze each stage
3. For reporting, perhaps organize each stage as a different section and indicate that each is a build stage of the next. Pinning a multistage Dockerfile is straightforward.

abhaykatheria:
@nishakm  that sounds awesome i will start looking into it.

ForgetMe17:
Maybe i can work on this issue!

ForgetMe17:
Hi nisha, I am considering this way of spliting:

Multistage Dockerfile
```
FROM golang:1.7.3
WORKDIR /go/src/github.com/alexellis/href-counter/
RUN go get -d -v golang.org/x/net/html  
COPY app.go .
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .

FROM alpine:latest  
RUN apk --no-cache add ca-certificates
WORKDIR /root/
COPY --from=0 /go/src/github.com/alexellis/href-counter/app . 
CMD ["./app"]  
```
into

Stage 1
```
FROM golang:1.7.3
WORKDIR /go/src/github.com/alexellis/href-counter/
RUN go get -d -v golang.org/x/net/html  
COPY app.go .
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
```

Stage 2
```
FROM golang:1.7.3
WORKDIR /go/src/github.com/alexellis/href-counter/
RUN go get -d -v golang.org/x/net/html  
COPY app.go .
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .

FROM alpine:latest  
RUN apk --no-cache add ca-certificates
WORKDIR /root/
COPY --from=0 /go/src/github.com/alexellis/href-counter/app . # Copy file from previous stage.
[...]

PR #774, opened 2020-07-18
Build multistage dockerfile images

ForgetMe17:
This PR will implement functions on building multistage dockerfile images. Here are goals:
1. Get the dockerfile for build.
2. Get the splited dockerfile object for analyze.

To implement this:
1. Split multistage dockerfile by stage.
2. Combine stages in order to get dockerfile for build.
3. Adding necessary tests.


tern/analyze/docker/dockerfile.py
```diff
@@ -363,3 +364,41 @@ def split_multistage_dockerfile(dfobj):
         # dfobj_temp not empty, append to the list
         dfobj_list.append(copy.copy(dfobj_temp))
     return dfobj_list
+
+
+def get_multistage_image_dockerfiles(dfobj_multi):
```
ForgetMe17, line 369: In this new function, we will create dockerfiles for each stage to build. A list of the dockerfile names will be returned.

tern/analyze/docker/dockerfile.py
```diff
@@ -363,3 +364,41 @@ def split_multistage_dockerfile(dfobj):
...
+    file_path_list = []
+    structure = []
+    idx = 0
+    for command_dict in dfobj_multi.structure:
+        if command_dict['instruction'] == 'FROM':
+            if structure:
+                file_path = dfobj_multi.filepath + '_%d' % (idx)
+                # we make a new dir for the dockerfile of each stage.
+                if not os.path.isdir(file_path):
+                    os.mkdir(file_path)
```
ForgetMe17, line 381: Since docker.image.build receives a path of folder which contains Dockerfile, so i chose to create a folder for each stage dockerfile.

tern/analyze/docker/run.py
```diff
@@ -180,3 +180,26 @@ def execute_dockerfile(args):  # noqa C901,R0912
...
+def build_multistage(dockerfile_path):
+    """Given a multistage dockerfile, first get the dockerfiles for each
+    stage and then build the dockerfiles.
+    TO DO:
+    1.Add image analyze
+    2. Clean the temp dockerfiles"""
+    dfobj_multi = dockerfile.get_dockerfile_obj(dockerfile_path)
+    file_path_list = dockerfile.get_multistage_image_dockerfiles(dfobj_multi)
+    print(file_path_list)
+    for dfile in file_path_list:
```
ForgetMe17, line 194: Same logic with execute_dockerfile(), we are going to loop on this for each dockerfile.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +335,70 @@ def write_locked_dockerfile(dfile, destination=None):
         file_name = constants.locked_dockerfile
     with open(file_name, 'w') as f:
         f.write(dfile)
+
+
+def check_mutistage_dockerfile(dfobj):
+    """Given a dockerfile object, return if it is a multistage dockerfile."""
+    from_line = [1 for c in dfobj.structure if c['instruction'] == 'FROM']
+    return len(from_line) > 1
```
nishakm, line 343: spelling error: `check_multistage_dockerfile`

Straightforward logic is good for readability:
```
342     count = 0                                                                   
343     for c in dfobj.structure:                                                   
344         if c['instruction'] == 'FROM':                                          
345             count = count + 1                                                   
346     return count > 1  
```

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +335,70 @@ def write_locked_dockerfile(dfile, destination=None):
...
+
+
+def split_multistage_dockerfile(dfobj):
```
nishakm, line 346: I am getting 3 Dockerfile objects for a 2 stage dockerfile: https://github.com/anuvu/zot/blob/master/Dockerfile

tern/analyze/docker/dockerfile.py
nishakm: I wonder if you can make use of the `check_multistage_dockerfile` function to also return the start line and the end line of the first `FROM` line and the next `FROM` line and then use that to split the dockerfiles. I would think that for a 2 stage dockerfile, you will only need one new Dockerfile object not including the original.

tern/analyze/docker/dockerfile.py
ForgetMe17: In the new commit, i have removed this function. We are now spliting the dockerfile for building, and `split_multistage_dockerfile()` gets the dfobj for analyze, so i will fix this function while i am working on the analyze.

tern/analyze/docker/dockerfile.py
ForgetMe17: Good advice! @nishakm. I have made these changes in the new commit.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
...
+
+
+def check_multistage_dockerfile(dfobj):
+    """Given a dockerfile object, return if it is a multistage dockerfile,
+    and the index of FROM line in the dfobj structure."""
+    from_line = []
+    for idx, st in enumerate(dfobj.structure):
+        if st['instruction'] == 'FROM':
+            from_line.append(idx)
+    return len(from_line) > 1, from_line
```
rnjudge, line 346: See Nisha's notes on readability here.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
         file_name = constants.locked_dockerfile
     with open(file_name, 'w') as f:
         f.write(dfile)
+
+
+def check_multistage_dockerfile(dfobj):
+    """Given a dockerfile object, return if it is a multistage dockerfile,
+    and the index of FROM line in the dfobj structure."""
```
rnjudge, line 341: small nit: `and the indexes of the FROM lines` since there will be multiple from lines in a multistage dockerfile.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
         file_name = constants.locked_dockerfile
     with open(file_name, 'w') as f:
         f.write(dfile)
+
+
+def check_multistage_dockerfile(dfobj):
+    """Given a dockerfile object, return if it is a multistage dockerfile,
```
rnjudge, line 340: small nit: `return True or False if it is a multistage dockerfile`

rnjudge: Also, do you see using the True/False flag? The only place where `check_multistage_dockerfile` is used in this PR, you throw away the True/False return value. Perhaps we only return the indexes of the FROM lines in the dfobj structure for this function? If the length of the list that you return is > 1, then we know it is multistage.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
...
+    return len(from_line) > 1, from_line
+
+
+def get_multistage_image_dockerfiles(dfobj_multi):
+    """Given a multistage dockerfile object, return a list of structures
+     for building image."""
+    file_path_list = []
+    structure = []
+    file_idx = 0
+    _, from_line = check_multistage_dockerfile(dfobj_multi)
```
rnjudge, line 355: maybe call this variable `from_lines` (plural) since there will be multiple?

tern/analyze/docker/dockerfile.py
rnjudge, line 340: Do you see yourself using the True/False return value here? It seems somewhat redundant. The only other place `check_multistage_dockerfile` is called in the PR, you throw away the True/False value. Users of this function will know it is a multistage dockerfile if the length of the list returned is > 1.

tern/analyze/docker/dockerfile.py
rnjudge: Proposal to only return the `from_line` here.

tern/analyze/docker/dockerfile.py
ForgetMe17: No. I think this is redundant here. I will remove this.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
...
+
+
+def check_multistage_dockerfile(dfobj):
+    """Given a dockerfile object, return the index(es) of FROM line(s)
+    in the dfobj structure."""
+    from_lines = []
+    for idx, st in enumerate(dfobj.structure):
+        if st['instruction'] == 'FROM':
+            from_lines.append(idx)
+    return from_lines
```
ForgetMe17, line 346: I have removed True/False return value and use `from_lines` here since there are possibly more than one FROM line.

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,47 @@ def write_locked_dockerfile(dfile, destination=None):
...
+                if not os.path.isdir(file_path):
+                    os.mkdir(file_path)
+                file_path += '/Dockerfile'
+                file_idx += 1
+                write_dockerfile_by_structure(file_path, structure)
+                file_path_list.append(file_path)
+        structure.append(dfobj_multi.structure[idx])
+    if structure:
+        file_path_list.append(dfobj_multi.filepath)
+    return file_path_list
```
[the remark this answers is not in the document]
rnjudge: According to the proposal in https://github.com/tern-tools/tern/issues/612#issuecomment-657929436, this is working as intended.

ForgetMe17:
Good advice! @nishakm , it will be more easier to clean up in the following process.

ForgetMe17:
I paste the changes from the previous commit here.
```
--- a/tern/analyze/docker/dockerfile.py
+++ b/tern/analyze/docker/dockerfile.py
@@ -355,14 +355,18 @@ def get_multistage_image_dockerfiles(dfobj_multi):
     from_lines = check_multistage_dockerfile(dfobj_multi)
     # Pop the first FROM
     from_lines.pop(0)
+    # Get the temp folder path
+    temp_folder_path = os.path.dirname(dfobj_multi.filepath) + '/dftemp'
+    if not os.path.isdir(temp_folder_path):
+        os.mkdir(temp_folder_path)
     for idx in range(len(dfobj_multi.structure)):
         if idx in from_lines:
             if structure:
-                file_path = dfobj_multi.filepath + '_%d' % (file_idx)
+                df_folder_path = temp_folder_path + '/%d' % (file_idx)
                 # we make a new dir for the dockerfile of each stage.
-                if not os.path.isdir(file_path):
-                    os.mkdir(file_path)
-                file_path += '/Dockerfile'
+                if not os.path.isdir(df_folder_path):
+                    os.mkdir(df_folder_path)
+                file_path = df_folder_path + '/Dockerfile'
                 file_idx += 1
                 write_dockerfile_by_structure(file_path, structure)
                 file_path_list.append(file_path)
```

tern/analyze/docker/dockerfile.py
```diff
@@ -333,3 +334,51 @@ def write_locked_dockerfile(dfile, destination=None):
...
+    """Given a multistage dockerfile object, return a list of structures
+     for building image."""
+    file_path_list = []
+    structure = []
+    file_idx = 0
+    from_lines = check_multistage_dockerfile(dfobj_multi)
+    # Pop the first FROM
+    from_lines.pop(0)
+    # Get the temp folder path
+    temp_folder_path = os.path.dirname(dfobj_multi.filepath) + '/dftemp'
```
nishakm, line 359: Can you move the constant `dftemp` to `tern/utils/constants.py`?

ForgetMe17: Do you mean `temp_folder_path = os.path.dirname(dfobj_multi.filepath) + constans.dftemp_dir` 
and in `tern/utils/constants.py` , `dftemp_dir='/dftemp'`

nishakm: In `tern/utils/constants.py` enter:
```
# temporary directory for multistage Dockerfile analysis
multistage_dir = 'dftemp'
```
And then in this line
`temp_folder_path` = os.path.join(os.path.dirname(dfobj_multi.filepath), multistage_dir)`

ForgetMe17: Got it!

Issue #757, opened 2020-07-08
Enable Tern to pull image digests and images using the registry HTTP(s) APIs


prahaladdarkin:
**Describe the Feature**
Enable Tern to pull images and image digests directly using the registry HTTP(s) APIs. This capability will ensure that Tern does not require the presence of docker client libs and utilities to pull images from any registry.

**Use Cases**
* Run Tern within a docker container as a Kubernetes service in non-privileged mode without requiring docker on docker
* Enable light-weight and fast pulling of image from various OCI compatible registries for e.g. Harbor and GCR.

**Implementation Changes**
I currently do not have a complete set of changes that would need to be done. However, as a start I was evaluating whether the [Google Container Registry Python SDKs](https://github.com/google/containerregistry) can be used for this purpose.
Alternatively, we can create a standalone command line executable written in Go that uses the [go-containerregistry SDK](https://github.com/google/go-containerregistry) or related tools kie ```crane```

mukultaneja:
@nishakm @prahaladdarkin @rnjudge Can I take it?

nishakm:
I figured out how to query the API using python: https://gist.github.com/nishakm/98d2fceaf8647bf521e98451123e5152

We will need `requests` to do this. `requests` doesn't have a "Bearer" authentication class, but it looks like one can create a custom authentication class: https://requests.readthedocs.io/en/master/user/authentication/#new-forms-of-authentication

A python implementation of the API handshake needs the following:
1. Docker style image path resolution: All of Docker's library images have the path `library/<image>`, but one can just use <image>. For example, `debian` resolves to `library/debian`. Alternatively, we don't resolve the path and hint to users that they may need to add `library` to the path. This is not needed for other images. This also holds for the domain name. Docker inherently resolves `debian:buster` to `https://index.docker.io/v1/debian/manifests/buster` and we may either resolve this ourselves or ask the user to provide a domain like `docker.io/debian:buster`. We will also need to keep a list of `https` endpoints for the different registries.
2. Implement the "bearer" authentication to this API. Luckily, for OCI compliant registries, the paths are standardized:
a. `GET "https://{domain}/v2/{image_path}/manifests/{tag}`
b. In the response headers, you should get the authentication domain, type, and query parameters.
c. Get a token using `GET "https://{auth_domain}?query_params"`. `requests` makes this easy.
[...]

PR #775, opened 2020-07-20
Pull Docker images using the registry HTTP(s) APIs

mukultaneja:
This commit adds support to download docker
images using HTTP Endpoint

Work towards: #757

Signed-off-by: mukultaneja <mukultaneja91@gmail.com>

### DIFF
--- a/tern/analyze/docker/container.py
+++ b/tern/analyze/docker/container.py
@@ -15,12 +15,15 @@ import pwd
 import requests
 import sys
 import time
-
+import json
+from dateutil import parser
+from datetime import datetime
 from tern.utils.constants import container
 from tern.utils.constants import logger_name
 from tern.utils.constants import temp_tarfile
 from tern.utils import rootfs
 from tern.utils import general
+from tern.utils.constants import docker_endpoint, temp_folder
 
 
 # timestamp tag
@@ -31,6 +34,7 @@ logger = logging.getLogger(logger_name)
 
 # global docker client
 client = None
+auth_token = None
 
 
 def check_docker_setup():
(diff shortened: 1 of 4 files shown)


nishakm:
@mukultaneja Can you make your commit message more descriptive?
```
This commit adds support to download docker -> This should be in the body of the commit message

images using HTTP Endpoint -> download images using the Docker Registry API

Work towards: #757 -> Fixes #757
```
Some more information about what changed would be nice. Thanks for working on this!

mukultaneja:
> @mukultaneja Can you make your commit message more descriptive?
> 
> ```
> This commit adds support to download docker -> This should be in the body of the commit message
> 
> images using HTTP Endpoint -> download images using the Docker Registry API
> 
> Work towards: #757 -> Fixes #757
> ```
> 
> Some more information about what changed would be nice. Thanks for working on this!

sure @nishakm I will do that. It was just an initial commit sort of POC kind!

nishakm:
@mukultaneja It would be better if this functionality was in the `tern/analyze/docker/container.py` file for now. That is where the Docker engine is used. We want to replace the function to pull the image with this logic.

mukultaneja:
> @mukultaneja It would be better if this functionality was in the `tern/analyze/docker/container.py` file for now. That is where the Docker engine is used. We want to replace the function to pull the image with this logic.

I assume we want to ignore all the container-related methods such as `check_container` and even docker-image as well.  Please correct me if I am wrong here, Instead of using the container APIs, we want to use docker REST APIs to download manifests, config and layers on the runtime so I thought it could be better if we add inside `Image` class any way we are using all the details inside `load_image` method of the class.

nishakm:
> @nishakm I assume we want to ignore all the container-related methods such as `check_container` and even docker-image as well. Please correct me if I am wrong here, Instead of using the container APIs, we want to use docker REST APIs to download manifests, config and layers on the runtime so I thought it could be better if we add inside `Image` class, anyway we are using all the details inside `load_image` method of the class.

Yup, we will use the Docker registry's REST API to pull the image. However, we still need the Docker engine to build an image from a Dockerfile as well as use it for the Dockerfile lock feature. So we only want to replace the `pull_image` function.


> P.S Yeah could be a case in future if we want to incorporate new types of container Images then I guess what you suggesting it makes sense possibly. But for OCI and docker, I think we can add in the `Image` class.

The `Image` class is meant to be a skeleton image that shouldn't have any functionality specific to Docker or OCI. So I would rather if we create an OCI image that inherits from this base Image class. OCI has some backwards compatibility features to Docker images, so it is better to keep them separate.

PR #777, opened 2020-07-23
Pull Docker images using the registry HTTP(s) APIs

mukultaneja:
This commit adds support to download docker
images using the Docker Registry API. It updates
analyze/docker/container `pull_image` method to
download supporting files for an image.

Fixes: #757

### DIFF
--- a/tern/analyze/docker/container.py
+++ b/tern/analyze/docker/container.py
@@ -15,12 +15,15 @@ import pwd
 import requests
 import sys
 import time
-
+import json
+from dateutil import parser
+from datetime import datetime
 from tern.utils.constants import container
 from tern.utils.constants import logger_name
 from tern.utils.constants import temp_tarfile
 from tern.utils import rootfs
 from tern.utils import general
+from tern.utils.constants import docker_endpoint, temp_folder
 
 
 # timestamp tag
@@ -31,6 +34,7 @@ logger = logging.getLogger(logger_name)
 
 # global docker client
 client = None
+auth_token = None
 
 
 def check_docker_setup():
@@ -88,12 +92,33 @@ def pull_image(image_tag_string):
     '''Try to pull an image from Dockerhub'''
     logger.debug("Attempting to pull image \"%s\"", image_tag_string)
     try:
-        image = client.images.pull(image_tag_string)
-        logger.debug("Image \"%s\" downloaded", image_tag_string)
-        return image
-    except (docker.errors.ImageNotFound, docker.errors.NotFound):
-        logger.warning("No such image: \"%s\"", image_tag_string)
+        rootfs.create_working_dir()
+        logger.debug("Downloading Image \"%s\"", image_tag_string)
+        image = os.path.join(rootfs.mount_dir, temp_folder)
+        image_name, image_tag = image_tag_string.split(":")
+        manifest = pull_image_manifest(image_name, image_tag)
+        with open(os.path.join(image, "manifest.json"), "w") as f:
+            json.dump(manifest, f)
+
+        digest = manifest.get("config").get("digest")
+        config = pull_image_config(image_name, image_tag, digest)
+        config_name = digest.split(":")[1]
+        with open(os.path.join(image, config_name), "w") as f:
+            json.dump(config, f)
+
+        layers = manifest.get("layers")
+        for layer, content in pull_image_layers(image_name, image_tag, layers):
+            with open(os.path.join(image, layer), "wb") as f:
+                f.write(content)
+
+    except Exception as ex:
+        logger.warning(ex)
         return None
+
+    logger.debug("Image saved \"%s\"", image_tag_string)
+    logger.debug("Image \"%s\" downloaded", image_tag_string)
+
+    return image
 
 
 def get_image(image_tag_string):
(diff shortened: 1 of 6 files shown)


mukultaneja:
@nishakm, I have closed the earlier PR and tried to address all the comments in this PR. I was thinking on the point where we want to support the `Dockerfile lock` feature and If I am not mistaken we need to modify this [function](https://github.com/tern-tools/tern/blob/master/tern/analyze/docker/container.py#L161) as we don't need the `if` condition now but we need `else` part for the `Dockerfile`. Also, do we need to download the image (files) on every run?

tern/analyze/docker/container.py
```diff
@@ -88,11 +92,13 @@ def pull_image(image_tag_string):
     '''Try to pull an image from Dockerhub'''
     logger.debug("Attempting to pull image \"%s\"", image_tag_string)
     try:
-        image = client.images.pull(image_tag_string)
+        logger.debug("Downloading Image \"%s\"", image_tag_string)
+        image = save_image(image_tag_string)
```
nishakm, line 96: You can just replace this whole logic with what is in `save_image`.

tern/analyze/docker/container.py
```diff
@@ -192,3 +198,139 @@ def close_client():
...
+def generage_image_blobs_url(image_name, digest):
+    '''
+    Generate a docker image blobs url end point with
+    given image name and tag
+    '''
+    url = "{0}/{1}/blobs/{2}"
+    return url.format(docker_endpoint, image_name, digest)
+
+
+def is_token_expired():
```
nishakm, line 248: I think this bearer authentication logic can be placed in `tern/utils/general.py` as it can be used for endpoints other than Docker.

tern/analyze/docker/container.py
```diff
@@ -192,3 +198,139 @@ def close_client():
...
+    start = datetime.timestamp(parser.isoparse(token_issued_at))
+    end = datetime.timestamp(datetime.now())
+
+    if end - start > token_expires_in:
+        return True
+
+    return False
+
+
+def generate_auth_token(image_name, image_tag):
```
nishakm, line 262: `get_auth_token(endpoint, image_name, image_tag)` is the more generic signature. You will then be able to move this to `general.py`.

tern/analyze/docker/container.py
nishakm: `is_token_expired(auth_token)` so you don't need to use a global `auth_token` variable.

tern/analyze/docker/container.py
```diff
@@ -192,3 +198,139 @@ def close_client():
...
+    '''Given manifest data, saves the image layers inside mount dir'''
+    global auth_token
+    headers = dict()
+    headers['Accept'] = 'application/vnd.docker.image.rootfs.diff.tar.gzip'
+    for layer in layers:
+        digest = layer.get("digest")
+        url = generage_image_blobs_url(image_name, digest)
+        # to generate token if it is expired while downloading a layer
+        if is_token_expired():
+            generate_auth_token(image_name, image_tag)
```
nishakm, line 312: I believe you can get a refresh token so you don't have to keep checking if a token is expired.

tern/analyze/docker/container.py
```diff
@@ -192,3 +198,139 @@ def close_client():
...
+    response = requests.get(url)
+    authentication = response.headers.get("Www-Authenticate").split(",")
+    auth_url, params = None, dict()
+    for line in authentication:
+        if "Bearer realm" in line:
+            auth_url = json.loads(line.split("=")[1])
+        if "service" in line:
+            params["service"] = json.loads(line.split("=")[1])
+        if "scope" in line:
+            params["scope"] = json.loads(line.split("=")[1])
```
nishakm, line 278: to get a refresh token, add `params["offline_token"] = True`. Then you will get a response with a `refresh_token` key. Then you can reuse that token for all the other GET requests - for manifest and layers.

nishakm: Curious about why you chose to use `json.loads` instead of `strip('\"').

tern/analyze/docker/container.py
mukultaneja: @nishakm here different endpoint means container services apart from docker you meant? I think getting auth token could be different for other endpoints in that case.

tern/analyze/docker/container.py
mukultaneja: @nishakm I tried to get `offline_token` using below api docs, 
https://docs.docker.com/registry/spec/auth/oauth/
https://docs.docker.com/registry/spec/auth/token/
but afraid could not get it.

nishakm: Update: I don't think the OAuth2 offline token thing works. So, let's try the check for token expiration time for now.

nishakm: Here's how I implemented the `auth_url` and `params` collecting:
```
    authentication = resp.headers.get("Www-Authenticate").split(",")            
    auth_url = authentication[0].split('=').pop().strip('\"')                   
    params = {}                                                                 
    for item in authentication[1:]:                                             
        param_list = item.split('=')                                           
       params[param_list[0]] = param_list[1].strip('\"')
```

tern/analyze/docker/container.py
nishakm: Definitely, but you would still have implemented the Bearer Authentication method. Incidentally, `requests` has a way of adding new auth methods: https://requests.readthedocs.io/en/master/user/authentication/#new-forms-of-authentication

tern/analyze/docker/container.py
mukultaneja: @nishakm I have moved the auth request inside `general.py` with `url` and `params` arguments

tern/analyze/docker/container.py
mukultaneja: @nishakm Done!

mukultaneja:
@nishakm I have tried to resolve most of the comments suggested by you. I have moved the auth token generation call to `get_auth_token` in `general.py` but I could not remove the usage of `auth_token` as a `global` variable as it is being used as a state holder between `is_token_expired` and `generate_auth_token` functions. It is like the outcome of `generate_auth_token` in the previous function call would be an input to `is_token_expired` in the next function call.

tern/analyze/docker/container.py
```diff
@@ -192,3 +217,104 @@ def close_client():
...
+def generage_image_blobs_url(image_name, digest):
+    '''
+    Generate a docker image blobs url end point with
+    given image name and tag
+    '''
+    url = "{0}/{1}/blobs/{2}"
+    return url.format(docker_endpoint, image_name, digest)
+
+
+def is_token_expired(auth_token):
```
nishakm, line 240: I had implemented this by checking if the response is 401. If it is, then get a fresh token and try again. What do you think of this method?

mukultaneja: I think the current implementation is much cleaner compared to 401 response.

mukultaneja:
@nishakm please suggest how should I resolve CI / CD failures. They are coming up due to `docker_image` class functionality.

mukultaneja:
@nishakm when are we planning to include these changes?

nishakm:
> @nishakm when are we planning to include these changes?

I don't think this is ready to merge in yet, mostly because the generic part of the code is not separate from the docker specific part of the code. And I seem to have a hard time providing you feedback on what to do in this respect. Would you mind if I take your changes and work on top of them?

mukultaneja:
> > @nishakm when are we planning to include these changes?
> 
> I don't think this is ready to merge in yet, mostly because the generic part of the code is not separate from the docker specific part of the code. And I seem to have a hard time providing you feedback on what to do in this respect. Would you mind if I take your changes and work on top of them?

sure @nishakm, please go ahead and take the changes. I am sorry for a hard time.

nishakm:
@mukultaneja do you want to try working on this again, from the tip of the main branch? You can take some inspiration from here: https://github.com/vsoch/oci-python/blob/master/opencontainers/distribution/reggie/client.py. Meanwhile I will close this PR.

Issue #781, opened 2020-08-04
Completeness of a tern scan


yannjor:
**Describe the Feature**
This feature would give you a summary at the end of the scan telling you the completeness of the scan, i.e. how big of a percentage of the files in the container can be attributed to a detected package in the bill of materials, and a list of the files that could not be associated to any package.  

**Use Cases** 
This feature would give the user more assurance about how much they can "trust" the scan results.

**Implementation Changes** 
Tern would first need the functionality to extract file info for packages to know what files were installed by which package. The completeness result would also need to be presented in the final output in some way.

rnjudge:
- I like the idea of expressing a “completeness %” of a scan but I think there is some ambiguity and room for clarification in that definition. I would think of completeness as the number of files/packages found where an associated license was detected in the file/package. It sounds like you are defining the completeness as a percentage of the files that belong to a package where the license was reported in the SBoM? What about files that have file-level licenses different from that of the overall package to which they belong? If you are not performing file-level scanning, the completeness percentage would incorrectly reflect the percentage of files covered by the detected SBoM if it just assumed that the file had the same license as the package that installed it. It sounds like your proposal is only concerned with associating a file with the package that installed it and counting that as “complete” for finding the license without actually looking at the file-level license. Am I understanding that correctly? I’m curious how you see handling this scenario where a file-level license differs from the package license. I would have a similar question about files that do not explicitly state a file-level license as well.

- We don’t currently have file-level scanning (i.e. looking at the contents of a file to find the license statement inside the file) enabled natively in Tern. File scanning is something that can be accomplished in Tern using the Scancode plugin. When you say, “Our idea is to scan all files in the container and associate the installed packages with the discovered files” are you planning to look at the file-level licenses of each individual file with Scancode, or just scan for the existence of a file natively in Tern?

- I would start the implementation for your proposal by first adding functionality to extract file info for packages. There was a contributor [working towards this ](https://github.com/tern-tools/tern/pull/573)a few months but the PR was eventually closed before it was merged.

- I’m not sure exactly how you are envisioning representing the association between installed packages and the files that they install, but the SPDX tag:value spec has ways to do this nicely. Or, if you see yourself creating a new type of output report with a more visually appealing presentation, it may be a good candidate for its own [custom report format plugin](https://github.com/tern-tools/tern/blob/master/docs/creating-custom-templates.md).

CsatariGergely:
[...]

PR #573, opened 2020-03-06
Extract file info for packages

PrajwalM2212:
Signed-off-by: PrajwalM2212 <prajwalmmath@gmail.com>

PR #787, opened 2020-08-14
Extract file info for packages

yannjor:
## Motivation
See issue #781 

## Proposed changes
This is a continuation of @PrajwalM2212 's pr #573. I tried to address the latest comments and change requests made by @nishakm on that pull request. 

Signed-off-by: Yann Jorelle <yann.jorelle@nokia.com>


yannjor:
I was thinking of merging the file data obtained when running tern with scancode [here](https://github.com/tern-tools/tern/blob/master/tern/extensions/scancode/executor.py#L113) with the package file data. But then I noticed that when you run tern with scancode, it doesn't seem to run the scripts in the command library that get the installed packages. 

This is a bit confusing to me because in this case scancode isn't really an extension, as it doesn't extend the metadata obtained from running the command library scripts. In my opinion it would make more sense if when running tern with scancode, it would first run the command library scripts and then run scancode to fill in the file metadata.

Does this make any sense, or am I missing something :blush: ?

nishakm:
> This is a bit confusing to me because in this case scancode isn't really an extension, as it doesn't extend the metadata obtained from running the command library scripts. In my opinion it would make more sense if when running tern with scancode, it would first run the command library scripts and then run scancode to fill in the file metadata.
> 
> Does this make any sense, or am I missing something blush ?

Hi @yannjor, sorry for getting back to you on this so late. Tern extensions are just a way to run any analyzer that is not "container aware" on a container image and to collate the results in any of the supported formats. I designed it this way because a lot of file analyzers are written in other languages and support different environments. For example, scancode cannot be installed with python 3.7 or 3.8 right now, so it cannot be used as a library in Tern itself. However, if python 3.6 was installed with tern and scancode, invoking scancode is not a problem. There is an issue to incorporate the package data that scancode produces #790. I am not sure if scancode can actually map what files came with what package installation though. This is why we treat each layer as a package with files in them.

The second problem is how do we resolve data pulled in from scancode (or any other extension) and what is already available in the cache. For example, the script method finds license MIT associated with a file, but scancode finds BSD. Another easier case is the script method was not able to find the license for a file associated with a package but scancode was able to find it for the same file. The metadata needs to be consolidated, not duplicated.

nishakm:
@yannjor I will take a look at this, and your other PR this week. Again, apologies for the delay 😅

yannjor:
@nishakm Ok, no worries :blush:

tern/classes/file_data.py
```diff
@@ -262,3 +262,15 @@ def merge(self, other):
                     self.origins.add_notice_to_origins(o.origin_str, n)
             return True
         return False
+
+    def is_equal(self, other):
+        '''This method performs a deep comparison between two objects by
+        iterating over the dictionary for each object returned by to_dict
+        and comparing the vales for each key in both. This function will only
+        return true if every key value pair matches between the two
+        dictionaries.'''
```
nishakm, line 271: I don't think we need to compare all the properties in the FileData object. Just the file name, path and checksum.

nishakm (changes requested):
Right now, there are FileData object that automatically generated at the ImageLayer level. I wonder if you could also add some functionality to remove the duplicate FileData object at the ImageLayer level that is found in the package level.

yannjor:
> Right now, there are FileData object that automatically generated at the ImageLayer level. I wonder if you could also add some functionality to remove the duplicate FileData object at the ImageLayer level that is found in the package level.

I added a function for this in `common.py` that gets called by `add_base_packages`. Do you think this is a good place to do this?

nishakm:
@yannjor I tested your changes and it looks like the files that get collected don't get into the report.
Eg:
```
1. Run tern -r report -f html -i debian:buster -o report.html
2. Open report.html with a browser
```
Look at `layers -> packages -> files` you will see all files are 0.

tern/classes/package.py
```diff
@@ -180,10 +187,21 @@ def fill(self, package_dict):
...
+        raw_file_list = package_dict.get('files', False)
+        if raw_file_list:
+            file_list = []
+            for file_path in raw_file_list:
+                if file_path != '':
+                    file_name = os.path.split(file_path)[1]
+                    fd = FileData(file_name, file_path)
+                    fd.fill({'name': file_name, 'path': file_path})
+                    file_list.append(fd)
+            package_dict['files'] = file_list
```
nishakm, line 204: I think this should be `self.files = file_list`.

nishakm (changes requested):
If you don't have time to work on this, do you mind if I work on top of it? It's almost ready to merge.

yannjor:
@nishakm 
> I tested your changes and it looks like the files that get collected don't get into the report.

I think the probem is with the `base.yml` scripts, but I can't really figure out how to fix it. It used to work before my latest changes to it.
Previously I had this (what @PrajwalM2212 wrote):

```YAML
  files:
    invoke:
      1:
        container:
        - "pkgs=`dpkg --get-selections | cut -f1 -d':' | awk '{print $1}'`"
        - 'for p in $pkgs; do files=`dpkg-query -L $p`; for file in $files; do if [ -f $file ]; then echo $file; fi; done; echo LICF; done'
    delimiter: 'LICF'
```
I then changed the delimiters to `"\n"` and removed the `echo LICF`, because otherwise the output would be a string and I would have to manually split it into a list in `package.py`:
```Python
        raw_file_list = package_dict.get('files', False)
        if raw_file_list:
            file_list = []
            for file_path in raw_file_list.split('\n'): # <--- Here
            ...
            package_dict['files'] = file_list
```
I noticed that most of the other scripts use awk to print stuff. Maybe it could be used here too? I'm don't have much experience in bash scripting so I would appreciate some help :blush: I am also fine with you working on top of this if you prefer.

yannjor:
@nishakm 
> I tested your changes and it looks like the files that get collected don't get into the report.

That problem came from a change I had made in`base.yml` in an attempt to make the package file extraction part more readable. I reverted the change so it should work now. The lines that fetch the files for the packages are still really long (about 150 characters) and not very readable in my opinion. Maybe [this](https://stackoverflow.com/a/3790497) YAML multi-line string syntax could be used?

I noticed another problem with the deletion of duplicate files that are found both at the layer and package level. The file paths for layer files seem to be relative paths from the root directory (eg: `etc/file`), while the package files use absolute paths right now. Should we change the layer files to use absolute paths instead or change the package files to use relative ones? Here is the [code](https://github.com/tern-tools/tern/blob/master/tern/classes/image_layer.py#L320-L324) in `image_layer.py` that creates the layer files.

nishakm:
@yannjor Sorry again for the delay. These changes looks good so I will merge the PR.

PR #791, opened 2020-08-25
Include packages collected by scancode in reports

### LINKED ISSUE #790: Include packages collected by scancode in reports
**Describe the Feature**
Currently, the scancode extension only collects file level data even though we also ask it to look for packages
```
# run scancode against a directory
command = 'scancode -ilpcu --quiet --timeout 300 --json -' <-- p stands for packages
```

**Use Cases**
Scancode is a complete static analysis method, so no mounting and and running a process is needed. We can use this method to address #755.

**Implementation Changes**
The executor needs to collect package information and store it in the `packages` list for `ImageLayer` objects.

yannjor:
Currently only file data is included when tern is run with the scancode
extension, even though scancode can also detect packages. These changes
add the package information collected by scancode to tern's data model.

Resolves #790

Signed-off-by: Yann Jorelle <yann.jorelle@nokia.com>

### DIFF
--- a/tern/classes/package.py
+++ b/tern/classes/package.py
@@ -202,3 +202,22 @@ class Package:
             if value != other_pkg_dict[key]:
                 return False
         return True
+
+    def merge(self, other):
+        '''Compare another Package object to this instance. If the name and
+        version are the same, we use the other object to fill in missing
+        metadata in the first one excluding the files and origins. This
+        method can be used in situations where an external scanner is used to
+        collect package data that we didn't find ourselves'''
+        if not isinstance(other, Package):
+            return False
+        if self.name == other.name and self.version == other.version:
+            other_pkg_dict = other.to_dict()
+            for key, value in self.to_dict().items():
+                if value == '' and other_pkg_dict[key] != '':
+                    setattr(self, key, other_pkg_dict[key])
+                    for lic in other.pkg_licenses:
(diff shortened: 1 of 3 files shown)


nishakm:
@yannjor Thanks for the PR! The changes look good to me. I noticed that there were no changes on reconciling packages inventoried by different tools. Do you have any thoughts on that?

yannjor:
> I noticed that there were no changes on reconciling packages inventoried by different tools. Do you have any thoughts on that?

@nishakm I had not thought about that :blush: I added one commit that tries to do this. It works so that if a package detected by scancode was already found some other way, it tries to fill in missing metadata in the existing one with the metadata of the one detected by scancode. I added a merge method to the Package class similar to the one in FileData. Do you agree with this approach?

nishakm:
> > I noticed that there were no changes on reconciling packages inventoried by different tools. Do you have any thoughts on that?
> 
> @nishakm I had not thought about that 😊 I added one commit that tries to do this. It works so that if a package detected by scancode was already found some other way, it tries to fill in missing metadata in the existing one with the metadata of the one detected by scancode. I added a merge method to the Package class similar to the one in FileData. Do you agree with this approach?

I actually forgot FileData had a merge method. That will do for now. Thanks so much!

nishakm:
Tested it on an nginx container. Looks good! Merging.

PR #795, opened 2020-08-29
Refactor Dockerfiles

isacsund:
The Dockerfiles were previously quite hard to read as the commands
were done as one-liners. This change adds newlines to the commands
to improve the readability.

Installation of pip packages is also moved to its own layer, which
allows us to utilize the layer cache for apt packages in case more pip
packages need to be installed later on (e.g. scancode-toolkit) to the
image.

Signed-off-by: Isac Sund <isac@isacsund.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,34 @@
 FROM debian:buster
 
 # Install fuse-overlayfs and Tern dependencies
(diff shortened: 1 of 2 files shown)


isacsund:
I'll have a look at why the test is failing for the `golang:alpine` image.

nishakm:
@isacsund Thanks for the PR! The change looks good to me, so I will merge it. Really appreciate you taking the time <3

PR #806, opened 2020-09-11
ENH: use `python:3.7-slim` base image and install local copy of tern

kaczmarj:
This commit uses the official python Docker image to reduce the number of packages to install with `apt-get`. This also greatly decreases the final image size, by about 400 MB. This commit also installs the local copy of tern. This is useful because if a developer makes changes to tern, it is possible to test those changes with tern's docker image. Before this commit, tern was installed from PyPI, so even if changes were made to the source, the latest PyPI version would be installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,39 @@
 # Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
+#
+# Build the Docker image:
+#   docker build --tag tern .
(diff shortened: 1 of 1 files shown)


nishakm:
@kaczmarj The reason we did not choose the python image as the starting image is because inventorying it would take some significant work in getting corresponding sources for the binaries we `wget`. Yes the image is smaller, but that is because the supply chain machinery is missing.

nishakm:
Closing this PR. See previous explanation.

PR #848, opened 2021-01-04
Change vagrant box from ubuntu/bionic64 to generic/ubuntu1804

aalexanderr:
Both boxes should be the same from usage perspective, the main difference is that one is maintained by Canonical while other is maintained by HashiCorp (thus has support for more providers).

This allows for usage with different providers (I added libvirt provider that I'm using but it will be easy enough now for people to add other providers when necessary)

Moved settings such as cpus and memory to the top to highlight most needed adjustments for the user.

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -15,7 +15,7 @@ Vagrant.configure("2") do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://vagrantcloud.com/search.
-  config.vm.box = "ubuntu/bionic64"
+  config.vm.box = "generic/ubuntu1804"
   config.vm.provision :shell, path: "bootstrap.sh"
 
   config.vm.synced_folder '.', '/vagrant', disabled: true
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -8,6 +8,12 @@
 # configures the configuration version (we support older styles for
 # backwards compatibility). Please don't change it unless you know what
 # you're doing.
+
+# Customize the amount of memory on the VM:
+VM_MEMORY = 4 * 1024
+# Customize amount of cores for the VM:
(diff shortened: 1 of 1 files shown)


vagrant/Vagrantfile
```diff
@@ -59,10 +65,16 @@ Vagrant.configure("2") do |config|
...
   #   vb.gui = true
   #
-  #   # Customize the amount of memory on the VM:
-    vb.memory = "4096"
+    vb.memory = VM_MEMORY
+    vb.cpus = VM_CPUS
   end
-  #
+
+  config.vm.provider :libvirt do |lv|
```
rnjudge, line 29: Does adding this libvirt provider support change the way users would bring up the vagrant box or ssh into the VM outlined [here](https://github.com/tern-tools/tern#getting-started-with-vagrant)? If so, we'll need to update the README.

aalexanderr: As for vagrant up, vagrant ssh, vagrant destroy, etc - works the same as with virtualbox provider (or any other provider for that matter).
https://github.com/vagrant-libvirt/vagrant-libvirt#features
As for `Install` section I think its best to leave it as is- Vagrant uses Virtualbox by default, I think its best not to confuse first-time users with too much options, while people who use different providers will be able to add support for their choice and possibly PR to upstream. Let me know if you think different, I'll add a commit with README Install step addition.

rnjudge: Ok, as long as users don't need to install additional libvirt dependencies to make vagrant work we can leave the README as is.

aalexanderr: vagrant will figure out which provider to use, as far as I remember, unless specified with --provider or ENV var, Vagrant will use first usable provider it finds supported in the Vagrantfile (and it goes in order of definition so virtualbox is first, libvirt is second)

PR #854, opened 2021-01-12
Add support for SPDX JSON format 

### LINKED ISSUE #839: Support for SPDX JSON format
**Describe the Feature**
Tern currently supports SPDX tag-value format, but it would be more compatible with cloud-native distribution if it were JSON format.

**Use Cases**
This format is specifically useful to demonstrate distributing container SBoMs using OCI artifacts

**Implementation Changes**
Create a format plugin that generates SPDX JSON format.

rnjudge:
Tern currently supports SPDX tag-value format, but in order to be more compatible with cloud-native distribution, it should also support SPDX JSON format.

This commit adds `spdxjson` as a report format in Tern. It does this by first refactoring some of the existing SPDX tag-value code so it can be utilized by both the tag-value and JSON SPDX formats in a common file `tern/formats/spdx/spdx_common.py`. It also creates a new directory, `tern/formats/spdx/spdxjson` where a handful of helper files are created to gather and format the correct JSON document data.

Resolves #839 


tern/formats/spdx/spdx_common.py
```diff
@@ -0,0 +1,159 @@
...
+    '''Return a SPDX formatted checksum value. It should be of the form:
+    checksum_type: <checksum>'''
+    return '{}: {}'.format(layer_obj.checksum_type.upper(), layer_obj.checksum)
+
+
+##########################
+# Common Package Helpers #
+##########################
+
+
```
rnjudge, line 111: remove this line

tern/formats/spdx/spdx_common.py
```diff
@@ -0,0 +1,8 @@
+ -*- coding: utf-8 -*-
```
rnjudge: Should be `#  -*- coding: utf-8 -*-`

Issue #868, opened 2021-01-25
Reduce complexity of analyze_subsequent_layers()


rnjudge:
**Description**
analyze_subsequent_layers() has 5 arguments. This is an issue if we ever want to add more command line arguments in the future that need to be passed to the layer analysis, as prospector will complain when there are more than 5 arguments.

**To Do**
Create a new file, `options.py` in `tern/analyze`. Create an `Options` class in options.py. Instead of passing all the command line arguments individually to the analyze functions, create an options object to hold them all.

**Background**
This will make implementing analysis related work easier, like #840.

nishakm:
@rnjudge what do you think of just passing `args` instead of creating a different options object?

PR #871, opened 2021-02-01
Refactored analyze_subsequent_layers()

rheo-chiti:
Created Options class with redo, shell & driver attribute and changed
the function analyze_subsequent_layers() accordingly in
tern/analyze/default/container/image.py & .../multi_layer.py

Solves issue - #868

Signed-off-by: rheo-chiti <rheo-chiti>

### DIFF
--- a/tern/analyze/options.py
+++ b/tern/analyze/options.py
@@ -1,0 +1,12 @@
+class Options:
+    '''A representation of user options
+    attributes:
+        shell: the shell found in the docker container
+        redo: redo flag to repopulate the cache for found layers
+        driver: required argument when running Tern in a container. If no
+        input is provided, 'fuse' will be used as the default option.
+    '''
+    def __init__(self, shell, redo, driver=None):
+        self.shell = shell
+        self.redo = redo
+        self.driver = driver
--- a/tern/analyze/default/container/image.py
+++ b/tern/analyze/default/container/image.py
@@ -18,6 +18,7 @@ from tern.analyze import passthrough
 from tern.analyze.default.container import single_layer
 from tern.analyze.default.container import multi_layer
 from tern.report import formats
+from tern.analyze.options import Options
 
 # global logger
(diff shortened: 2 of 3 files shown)


tern/analyze/options.py
```diff
@@ -0,0 +1,12 @@
+class Options:
+    '''A representation of user options
+    attributes:
+        shell: the shell found in the docker container
+        redo: redo flag to repopulate the cache for found layers
+        driver: required argument when running Tern in a container. If no
+        input is provided, 'fuse' will be used as the default option.
+    '''
+    def __init__(self, shell, redo, driver=None):
```
aalexanderr, line 9: An issue I encoutered earlier was when I added more options I ended up with functions with too many arguments (linter complained). If there would be more args added later, someone would encounter the same problem. What do you think about creating instance of this class in __main__.py and letting argparse set members (via https://docs.python.org/3/library/argparse.html#dest) rather than via __init__()?

aalexanderr (commented):
I think those options are also used in the tern/analyze/dockerfile/run.py - you might want to use Options there too

tern/analyze/options.py
nishakm: Having the `Options` class in `__main__.py` and having argparse set members is how I would expect this to work.

rnjudge: @rheo-chiti -- can you move the `Options` class to __main__.py and use argparse to set the members?

rheo-chiti: @nishakm @rnjudge  , I moved the ```Options``` class to main.py but I am still unable to understand how to use argparse to set it's members.

nishakm: I can take a look later...

nishakm:
@rheo-chiti If you don't mind, I'd like to solve this issue. The fix is not that straightforward. Please stay tuned for more issues. Thanks!

PR #877, opened 2021-02-09
Fix tests for human readable id creation

nishakm:
The Image class now takes a repotag string of the form:
'image:tag' or 'image@hash:digest'. This commit updates the
documentation for the class and the corresponding test.

Since the human readable id is based on repo name and a repo tag,
we update the tests to check if the appropriate string is returned
For different combination of settings.


nishakm:
@rnjudge this fixes the failing CI test for `test_class_image.py`.

PR #878, opened 2021-02-10
Enabled SPDX document validation for CI

### LINKED ISSUE #713: Enable SPDX document validation for CI
**Description**
As development increases on the project, it is getting hard to maintain the generation of valid SPDX documents for the container images. So enable CI checks for SPDX if anything in the generator or formatting code changes.

**To Do**
1. Setup requires install of openjdk and the SPDX tools from their release page: https://github.com/spdx/tools/releases
2. Then run `tern report -f spdxtagvalue -i <a chosen image> -o spdx.txt` and run the validation tool on the output.

rnjudge:
As development increases for Tern, it is getting harder to maintain the
generation of valid SPDX documents for container images. This commit
enables CI checks for the SPDX documents that Tern produces to make sure
the documents are valid.

Resolves #713

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -60,11 +60,15 @@ jobs:
           python-version: '3.7'
       - name: Setup
         run: |
-          sudo apt-get install -y attr
+          sudo apt-get update && sudo apt-get install -y attr openjdk-8-jdk-headless maven
           pip install --upgrade pip
           pip install GitPython~=2.1
           pip install .
           docker pull photon:3.0 && docker save photon:3.0 > photon.tar
+          # build SPDX validation tool from source
+          git clone https://github.com/spdx/tools-java.git && cd tools-java
+          export JAVA_HOME=$(readlink -f /usr/bin/javac | sed "s:/bin/javac::")
+          mvn clean install && cd ..
       - name: Test Changes
         run: python ci/test_files_touched.py
   Test_Coverage:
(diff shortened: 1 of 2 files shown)


rnjudge:
Added changes to `spdxtagvalue/generator.py` in order to make the document not validate and the CI tests failed as expected.

Screen Shot 2021-02-09 at 8 20 22 PM

rnjudge:
@nishakm this PR downloads a specific version of the spdx tooling for the CI tests. I don't see a way to always download the latest from bintray so I wonder if we just want to update the version in the CI tooling with each release? It seems there's a new SPDX tool every few months but this might slow down as the SPDX tool gets to be more stable.

nishakm:
> @nishakm this PR downloads a specific version of the spdx tooling for the CI tests. I don't see a way to always download the latest from bintray so I wonder if we just want to update the version in the CI tooling with each release? It seems there's a new SPDX tool every few months but this might slow down as the SPDX tool gets to be more stable.

Can we build it from source?

rnjudge:
> > @nishakm this PR downloads a specific version of the spdx tooling for the CI tests. I don't see a way to always download the latest from bintray so I wonder if we just want to update the version in the CI tooling with each release? It seems there's a new SPDX tool every few months but this might slow down as the SPDX tool gets to be more stable.
> 
> Can we build it from source?

Good call.

rnjudge:
The commit now builds from source.

Screen Shot 2021-02-11 at 8 17 25 PM

PR #879, opened 2021-02-10
Fix for not retrieving cached notices

nishakm:
The first commit saves the found OS and package
format to the specific layer so they are available to
all format generators.

The second commit loads the layer notices from the
cache.


rnjudge:
LGTM (Looks good to me). LGTM (Lets get this merged).

PR #891, opened 2021-02-24
Fix for docker registries hosted on different port

aalexanderr:
This is a hot-fix, I didn't dive into how the docker images can be named and it *might* not be correct solution. If someone could help me with listing possible solutions I can adjust the PR.

Some comments should also be added, but I'd prefer to get some input first and adjust it later (as I can use it for ONAP for now)

### DIFF
--- a/tern/utils/general.py
+++ b/tern/utils/general.py
@@ -306,4 +306,9 @@ def parse_image_string(image_string):
                 'tag': '',
                 'digest_type': tokens[1],
                 'digest': tokens[2]}
+    if len(tokens) == 4:
+        return {'name': tokens[0] + ":" + tokens[1],
+                'tag': '',
+                'digest_type': tokens[2],
+                'digest': tokens[3]}
     return None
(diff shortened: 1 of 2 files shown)


nishakm:
@aalexanderr Looks like this is the right fix. I'll test it out.

nishakm:
> @aalexanderr Looks like this is the right fix. I'll test it out.

It would also be nice to add a test case for this kind of image name to `tests/test_class_docker_image.py`.

nishakm:
Tested the parsing with the repo name with the port and the digest:
```
>>> d = DockerImage("nexus3.onap.org:10001/onap/so/so-oof-adapter@sha256:d7e1f739ba732c853a638f9c90becd5e0f8d313c8d506567b0b83ac38a1d53cb")
>>> d.name
'nexus3.onap.org10001/onap/so/so-oof-adapter'
```
I think here we want to ignore the port. So `d.name` should be `nexus3.onap.org/onap/so/so-oof-adapter`.

aalexanderr:
@nishakm We actually don't want to ignore the port, it identifies where image comes from. Consider multiple registries on same  different ports (which is what nexus3.onap.org has)
I have doubts about relying on external docker registry for tests here thus added a test in test_utils_general.

aalexanderr:
This works for nexus images, but when parse_image_string would get "this.is.some.registry.gov:1337/some-image:tag" it might not work as intended

nishakm:
> This works for nexus images, but when parse_image_string would get "this.is.some.registry.gov:1337/some-image:tag" it might not work as intended

Two possible types of inputs are an image string or a tarball. I'll try both and see what happens.

nishakm:
@aalexanderr Looks good! I'll merge this.

Issue #825, opened 2020-11-10
Consolidate common code used by container/run.py and dockerfile/run.py


nishakm:
**Description**
There are some steps in common between `container/run.py` and `dockerfile/run.py`. These can be consolidated in one function and placed in `tern/analyze/default/core.py`.

**To Do**
1. Find the steps that are common
2. Place the steps in a function in `core.py`
3. Change imports in `container/run.py` and `dockerfile/run.py`
4. Run functional tests

**Super Issues**
#801

m1-key:
@nishakm I am looking forward to solving this issue. I was reading the code of both files. What I found that we can use the container/run.py functions at some places. Like cleanup and saving to the cache was there in dockerfile/run.py, for that we can use directly teardown function of container/run.py. Similarly, at one-two places, we can do the same.

nishakm:
Thanks @m1-key! Could you post a list of things you would be changing? I haven't looked at this in a while now :)

m1-key:
> Thanks @m1-key! Could you post a list of things you would be changing? I haven't looked at this in a while now :)

Till now I have found that in container/run.py we have a function named ```execute_image``` which is doing the same thing as done by two functions combined in dockerfile/run.py namely ```full_image_analysis``` and ```analyze_full_image```.
I have shared a screenshot of the same. As you can see the same sequence of instructions has been followed. So rather than having this, we can use the ```execute_image``` function itself with slight modifications. 

![WhatsApp Image 2021-02-23 at 21 27 07](https://user-images.githubusercontent.com/39923784/108870215-ed29c980-761d-11eb-9aa2-0b80aa3de5c2.jpeg)


![WhatsApp Image 2021-02-23 at 21 26 00](https://user-images.githubusercontent.com/39923784/108870278-fa46b880-761d-11eb-8f5f-7b17794dfc8b.jpeg)

nishakm:
@m1-key The trouble with using `execute_image` as part of `execute_dockerfile` is that the number of checks you will need to make sure the right arguments are passed in and out based on whether you are providing an image or a dockerfile is so much that it increases complexity. This was how `execute_image` was before (it was just called `execute` back then). The splitting of the execution paths reduced complexity and helps us target the right code blocks for the right concern. Can you identify any repeat procedures that can be moved into their own function, maybe under `default_common` that can be called in `execute_image` and `execute_dockerfile`?

[...]

Issue #801, opened 2020-09-10
The big code refactor


nishakm:
**Description**
Several features requested by our community can be enabled easily if we re-organize the code base. This would also make it easier to write tests for it and increase our code coverage, not to mention making the code readable and help us solve a good number of our technical debt issues.

This is a super issue. A number of issues will be created that will link here.

**Sub Issues**
#802
#803 
#804
#807 
#808
#811

emaanhaseem:
I and a small group of students from a UT Austin CS Virtualization class would like to claim this issue and tackle it.

nishakm:
> I and a small group of students from a UT Austin CS Virtualization class would like to claim this issue and tackle it.

Super appreciate you for offering to tackle this! It's a tricky one and time sensitive (like the end of September!). I wonder maybe if you are looking for something more long term?

emaanhaseem:
Thanks for letting us know the time sensitivity! Is there an issue you'd recommend that is more long term? For example to be completed by the end of the year. Thanks again.

nishakm:
> Thanks for letting us know the time sensitivity! Is there an issue you'd recommend that is more long term? For example to be completed by the end of the year. Thanks again.

We have a number of bugs due by next release: https://github.com/tern-tools/tern/milestone/13. Definitely try one of those. I will be filing more issues that you can take on. Let me know which ones you would like to take on.

nishakm:
All relevant issues are closed.

PR #892, opened 2021-02-25
Consolidated common code used by in run.py

m1-key:
I have made a function named default_common_procedure
which have steps common in container/run.py and dockerfile/run.py.

Solved Issue: #825
Work towards: #801

Signed-off-by: m1-key <shubhamtiwari.tiwari84@gmail.com>

### DIFF
--- a/tern/analyze/default/container/image.py
+++ b/tern/analyze/default/container/image.py
@@ -14,9 +14,13 @@ import subprocess  # nosec
 from tern.classes.notice import Notice
 from tern.classes.docker_image import DockerImage
 from tern.utils import constants
+from tern.utils import rootfs
+from tern.report import report
 from tern.analyze import passthrough
+from tern.analyze import common
 from tern.analyze.default.container import single_layer
 from tern.analyze.default.container import multi_layer
+from tern.analyze.default.dockerfile import lock
 from tern.report import formats
 
 # global logger
@@ -73,3 +77,23 @@ def analyze(image_obj, options):
         passthrough.run_extension(image_obj, options.extend, options.redo)
     else:
         default_analyze(image_obj, options)
+
+
+def common_container_procedure(image, options, image_file=None):
+    """Setup the image object and anything else for analysis"""
+    # Add a Notice object for each layer
+    for layer in image.layers:
+        origin_str = 'Layer {}'.format(layer.layer_index)
+        layer.origins.add_notice_origin(origin_str)
+    # Set up working directories and mount points
+    rootfs.set_up()
+    # analyze the image
+    analyze(image, options)
+    if image_file == "Full_Image":
+        lock.set_imported_layers(image)
+    elif image_file is None:
+        report.report_out(options, image)
+    # Add the image layers to the cache
+    common.save_to_cache(image)
+    # Clean up working directories and mount points
+    rootfs.clean_up()
--- a/tern/analyze/default/container/run.py
+++ b/tern/analyze/default/container/run.py
@@ -11,7 +11,6 @@ import logging
 
 from tern.utils import constants
 from tern.utils import rootfs
-from tern.report import report
 from tern.report import formats
 from tern import prep
 from tern.load import docker_api
@@ -83,13 +82,7 @@ def execute_image(args):
             full_image.origins.add_notice_origin(
                 formats.docker_image.format(imagetag=image_string))
             # Set up for analysis
-            setup(full_image)
-            # analyze image
-            cimage.analyze(full_image, args)
-            # report out
-            report.report_out(args, full_image)
-            # clean up
-            teardown(full_image)
+            cimage.common_container_procedure(full_image, args)
         else:
             # we cannot load the full image
             logger.error('Cannot retrieve full image metadata')
--- a/tern/analyze/default/dockerfile/run.py
+++ b/tern/analyze/default/dockerfile/run.py
@@ -13,18 +13,15 @@ import os
 import subprocess  # nosec
 
 from tern.utils import constants
-from tern.utils import rootfs
 from tern.classes.notice import Notice
 from tern.classes.image_layer import ImageLayer
 from tern.classes.image import Image
 from tern.classes.package import Package
 from tern.load import docker_api
 from tern import prep
-from tern.analyze import common
 from tern.analyze.default import filter as fltr
 from tern.analyze.default.dockerfile import parse
 from tern.analyze.default.dockerfile import lock
-from tern.analyze.default.container import run as crun
 from tern.analyze.default.container import image as cimage
 from tern.report import report
 from tern.report import formats
@@ -69,15 +66,7 @@ def analyze_full_image(full_image, options):
     """If we are able to load a full image after a build, we can run an
     analysis on it"""
     # set up for analysis
-    crun.setup(full_image)
-    # analyze image
-    cimage.analyze(full_image, options)
-    # clean up after analysis
-    rootfs.clean_up()
-    # we should now be able to set imported layers
-    lock.set_imported_layers(full_image)
-    # save to the cache
-    common.save_to_cache(full_image)
+    cimage.common_container_procedure(full_image, options, image_file="Full_Image")
     return [full_image]
 
 
@@ -106,15 +95,7 @@ def analyze_base_image(base_image, options):
     """If we are unable to load the full image, we will try to analyze
     the base image and try to extrapolate"""
     # set up for analysis
-    crun.setup(base_image)
-    # analyze image
-    cimage.analyze(base_image, options)
-    # clean up
-    rootfs.clean_up()
-    # save the base image to cache
-    common.save_to_cache(base_image)
-    # let's try to figure out what packages were going to be installed in
-    # the dockerfile anyway
+    cimage.common_container_procedure(base_image, options, image_file="Base_Image")
     stub_image = get_dockerfile_packages()
     return [base_image, stub_image]
 


tern/analyze/default/default_common.py
```diff
@@ -14,11 +14,14 @@
...
 from tern.classes.notice import Notice
 from tern.report import errors
+from tern.report import report
 from tern.utils import constants
 from tern.utils import rootfs
 from tern.utils import general
+from tern.analyze import common
 from tern.analyze.default.command_lib import command_lib
 from tern.analyze.default import filter as fltr
+from tern.analyze.default.container import image as cimage
```
nishakm, line 24: This import is causing a lot of circular import linking errors. Ideally, you'd want to follow the code structure described here: https://github.com/tern-tools/tern/blob/main/docs/navigating-the-code.md

So things in `container` should import from `default_common` and not the other way around.

nishakm (changes requested):
This may be touching another WIP: https://github.com/tern-tools/tern/issues/868  so watch out for any merge conflicts that will come up in the future.

tern/analyze/default/container/run.py
```diff
@@ -81,14 +81,9 @@ def execute_image(args):
...
-            setup(full_image)
-            # analyze image
-            cimage.analyze(full_image, args.redo, args.driver, args.extend)
-            # report out
-            report.report_out(args, full_image)
-            # clean up
-            teardown(full_image)
+            # Doing analysis
+            dcom.default_common_procedure(
+                full_image, args.redo, args.driver, args.extend, args=args, image_file=True)
```
rnjudge, line 86: I think this is going to trigger a prospector error with too many arguments.

rnjudge: Oh I see, because you call it again in `run.py` for a Docker. Hmm... prospector is still going to complain about number of args I think and I don't want to merge technical debt in here.

tern/analyze/default/dockerfile/run.py
```diff
@@ -68,16 +66,11 @@ def get_dockerfile_packages():
...
-    # clean up after analysis
-    rootfs.clean_up()
-    # we should now be able to set imported layers
+    # Doing analysis
+    print("Docker file given...")
+    dcom.default_common_procedure(
+        full_image, redo, driver, extension)
     lock.set_imported_layers(full_image)
-    # save to the cache
-    common.save_to_cache(full_image)
```
rnjudge, line 80: `save_to_cache` happens here *after* `lock.set_imported_layers(full_image)` but with your change, it now happens before we set the imported layers. I think this might change what gets saved to the cache? Have you experimented with this?

tern/analyze/default/container/run.py
m1-key: I think I can get rid of image_file here, only checking whether args are None or not will help to recognize that function is called from `dockerfile/run.py` or from `container/run.py`, hence image_file argument is not required. I think this will remove the technical debt that you are taking @rnjudge.

tern/analyze/default/default_common.py
m1-key: So @nishakm  should I remove this function `default_common_procedure` from `default_common.py` to `common.py`  ?

nishakm: No you will need to remove this line.

m1-key: If I remove this line, then how `cimage.analyze` will be used inside that `default_common.py`? Is there any way around for this?

tern/analyze/default/dockerfile/run.py
m1-key: My bad, I overlooked it. Now I have fixed this. What I did is that I removed the `image_file` argument and used `args` as my parameter to check from where this function is called. This will also help me to reduce the arguments from 6 to 5.
If `args` == `Full_Image` -> Then locking will happen first
If `args` == `Base_Image` -> Then just pass
else -> report will be called.
Here is the code: 
```
def default_common_procedure(image, redo, driver, extension, args=None):
    """Setup the image object and anything else for analysis"""
    # Add a Notice object for each layer
    for layer in image.layers:
        origin_str = 'Layer {}'.format(layer.layer_index)
        layer.origins.add_notice_origin(origin_str)
    # Set up working directories and mount points
    rootfs.set_up()
    # analyze the image
    cimage.analyze(image, redo, driver, extension)
    if args=="Full_Image":
        lock.set_imported_layers(image)
    elif args=="Base_Image":
        pass
    else:
        report.report_out(args, image)
    # Add the image layers to the cache
    common.save_to_cache(image)
    # Clean up working directories and mount points
    rootfs.clean_up()
```

tern/analyze/default/default_common.py
```diff
@@ -150,3 +153,21 @@ def abort_analysis():
...
+def default_common_procedure(image, redo, driver, extension, args=None, image_file=False):
+    """Setup the image object and anything else for analysis"""
+    # Add a Notice object for each layer
+    for layer in image.layers:
+        origin_str = 'Layer {}'.format(layer.layer_index)
+        layer.origins.add_notice_origin(origin_str)
+    # Set up working directories and mount points
+    rootfs.set_up()
+    # analyze the image
+    cimage.analyze(image, redo, driver, extension)
```
m1-key, line 167: @nishakm I need cimage here in 167th line, that's why I'm importing it in default_common. One way around is that I remove this entire function from `default_common.py` to `common.py`, this will resolve the circular imports issue that you mentioned.

nishakm:
@m1-key #895 is now merged in. You can update your PR now.

m1-key:
> @m1-key #895 is now merged in. You can update your PR now.

@nishakm I have updated the PR. I have used an argument named `image_file` which tells from where the function has been called, so according to that few specific steps will be performed in between the common set of procedures.

tern/analyze/default/container/image.py
```diff
@@ -73,3 +77,25 @@ def analyze(image_obj, options):
         passthrough.run_extension(image_obj, options.extend, options.redo)
     else:
         default_analyze(image_obj, options)
+
+
+def common_container_procedure(image, options, image_file=None):
```
rnjudge, line 82: Small nit: What do you think about renaming this function to something a little more intuitive/descriptive? Maybe `common_image_setup` or `prepare_for_analysis`? @nishakm do you have a preference?

tern/analyze/default/container/image.py
```diff
@@ -73,3 +77,25 @@ def analyze(image_obj, options):
...
+    # Set up working directories and mount points
+    rootfs.set_up()
+    # analyze the image
+    analyze(image, options)
+    if image_file == "Full_Image":
+        lock.set_imported_layers(image)
+    elif image_file == "Base_Image":
+        pass
+    else:
+        report.report_out(options, image)
```
rnjudge, line 97: small nit: Can you remove the the check for base image since we don't do anything with it? i.e.

```
    if image_file == "Full_Image":
        lock.set_imported_layers(image)
    elif image_file is None:
        report.report_out(options, image)
```

m1-key: Surely, one more thing that I think I should change is `image_file` with `image_type`? Will it be better ?

rnjudge: I think `image_file` is fine.

tern/analyze/default/container/image.py
nishakm: I actually asked @m1-key to call it "common_container_procedure". It's not an image setup thing but a list of steps in common with analyzing a given image and a given dockerfile. I am open to naming it something else that communicates this function/subroutine.

tern/analyze/default/container/image.py
```diff
@@ -73,3 +77,23 @@ def analyze(image_obj, options):
...
+    # Add a Notice object for each layer
+    for layer in image.layers:
+        origin_str = 'Layer {}'.format(layer.layer_index)
+        layer.origins.add_notice_origin(origin_str)
+    # Set up working directories and mount points
+    rootfs.set_up()
+    # analyze the image
+    analyze(image, options)
+    if image_file == "Full_Image":
+        lock.set_imported_layers(image)
```
nishakm, line 93: I am a little conflicted here. On the one hand, having a set of common steps that can be used in the container image case and the dockerfile case is great. On the other hand, the `set_imported_layers` function is only used in the dockerfile case. So I would rather not include this line here.

rnjudge: @nishakm I think the issue with removing `set_imported_layers` from here is that the call to set_imported_layers needs to happen before we call `common.save_to_cache` which is the next line in this new function.

tern/analyze/default/container/image.py
rnjudge: Ah, ok.. sorry @m1-key for the confusion!! I guess the "procedure" is just kind of throwing me off when paired with the function description that says """Setup the image object and anything else for analysis""". What about `common_analysis_procedure`?

tern/analyze/default/container/image.py
nishakm: I understand. The reason I duplicated the `cimage.analyze` function in `container/run.py` and `dockerfile/run.py` is so contributors can understand where the execution path is. So `image_file` shouldn't even exist in `container/run.py`. Perhaps `image.py` needs to live in `default` so the functions can be used both in `container/run.py` and `dockerfile/run.py`? What do you think?

m1-key: @nishakm so, should I remove the line that you asked to remove? If I do so then how I would be able to set `lock.set_imported_layers(full_image)`? Any workaround for this?

nishakm: I think the way around this is to move `image.py` out of the `container` module, but I'd like @rnjudge's input as well.

nishakm:
@m1-key After thinking about this further, I don't think we can consolidate the functionality between the container image case and the Dockerfile case anymore. So I will close the PR and the issue. Thanks so much for your work on this, and keep a lookout for more issues coming up if you would like to contribute again :)

m1-key:
> @m1-key After thinking about this further, I don't think we can consolidate the functionality between the container image case and the Dockerfile case anymore. So I will close the PR and the issue. Thanks so much for your work on this, and keep a lookout for more issues coming up if you would like to contribute again :)

Surely, I am looking forward to contribute more to Tern.

Issue #885, opened 2021-02-18
Automate Tern's release process


rnjudge:
**Describe the Feature**
When a new or patched release is cut, the Tern maintainers follow the steps outlined [here](https://github.com/tern-tools/tern/blob/main/docs/releases/release_checklist.md). This is time consuming and limits the frequency that releases can be made available. Ideally, we would automate this process.

**Use Cases**
Anytime we want to cut a release, having at least a portion of this process automated will save time and reduce errors.

**Implementation Changes**
Write a script that takes the release notes and other relevant files or information as inputs and cuts a release. It may not be possible to fully automate this, but we the script should at least prepare all the necessary material for the maintainers to cut a release faster than the fully manual process we have now.

coderpatros:
Heya, looking at the different parts of the release process.

Could "Update direct dependencies and run tests" be replaced with dependabot?

coderpatros:
Hi @rnjudge, @nishakm might of mentioned I have a software engineering student doing a uni placement at the moment.

He is starting to look at this issue. We spent a bit of time reviewing the release checklist and I would appreciate feedback on the following as an approach.

- manage direct dependency updates separately to releases, using dependabot (runs tests automatically when the PR is raised)
- prepare release PR
    - manually triggered GitHub workflow, which does the following steps
    - creates a release notes file, `next.md` with section headings and relevant `git log` output
    - copy `requirements.txt` to `docs/releases/next-requirements.txt`
    - creates `next` branch and PR
- preparation of release notes
    - maintainer 1 completes release notes in the `next` PR branch
    - maintainer 2 reviews release notes, merges PR
- create and publish release
    - manually trigger GitHub workflow, version specified when triggering release, which does the following steps
    - rename `next.md` and `next-requirements.txt` to include version, create release commit
    - build package
    - install locally built package and run smoke tests
    - build docker image
    - run locally built docker image and run smoke tests
    - create GitHub release (which also tags the commit)
    - create Tern vendor tar and upload to GitHub release
    - upload package to GitHub release
    - publish package to PyPI
    - publish docker image to Docker Hub

coderpatros:
@ReconPangolin I'll try to review what you've done in your fork tomorrow to provide some feedback before raising a draft PR here.

Can you have a look at what I've put above? I put some more thought into the steps required across the whole process. You could start working on the create and publish release steps next.

ReconPangolin:
[...]

PR #910, opened 2021-03-12
Automate Tern's Release Process

jaindhairyahere:
This is a raw version and is currently in development. 
@rnjudge @nishakm How is this for a start. I've tried to mimic the instructions in a user-friendly manner. Please suggest further improvements needed for the same.

For Issues: #885

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -119,8 +119,8 @@ echo "2. Enter the path to the release notes file"
 echo "Your choice [1/2]:- "
 l=1
 while [ $l -eq 1 ]; do
-  read -r release_notes_path
-  case $manualWork2 in
+  read -r choice
+  case $choice in
     1)
       touch "v$release_version.md"
       echo "Press any key (or y) when you are done writing the release notes and want to continue; press n otherwise."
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,181 @@
...
+echo "1. Write release notes all this stuff manually"
+echo "2. Enter the path to the release notes file"
+echo "Your choice [1/2]:- "
+l=1
+while [ $l -eq 1 ]; do
+  read -r release_notes_path
+  case $manualWork2 in
+    1)
+      touch "v$release_version.md"
+      echo "Press any key (or y) when you are done writing the release notes and want to continue; press n otherwise."
```
rnjudge, line 126: Would it be possible to pass in a path to a file here? Usually release notes are written up in advance and copy/pasting them here risks losing some of the markdown formatting.

jaindhairyahere: @rnjudge This option has been kept dynamic and is asked for, by the script, during the script's execution. So during lines 118-122, if you want to use a path instead of manually-copy-pasting, you'll have to choose option 2. I can also completely remove this choice if you want.

nishakm:
Closing this as someone else has started work on it.

Issue #810, opened 2020-09-21
Traceback in commit message linting


nishakm:
**Describe the bug**
Commit message linting doesn't gracefully exit when met with an error:
```
Run python ci/test_commit_message.py
Commit 01c5894 does not have a body.
Traceback (most recent call last):
  File "ci/test_commit_message.py", line 78, in <module>
    lint_commit(commit_id)
  File "ci/test_commit_message.py", line 42, in lint_commit
    msg_list.pop()
IndexError: pop from empty list
```

**To Reproduce**
1. Make a commit with just a title
2. Run ci/test_commit_message.py
3. See error

**Error in terminal**
```
Run python ci/test_commit_message.py
Commit 01c5894 does not have a body.
Traceback (most recent call last):
  File "ci/test_commit_message.py", line 78, in <module>
    lint_commit(commit_id)
  File "ci/test_commit_message.py", line 42, in lint_commit
    msg_list.pop()
IndexError: pop from empty list
```

**Expected behavior**
It should exit gracefully.

**Environment you are running Tern on**
On GitHub actions: https://github.com/tern-tools/tern/pull/805/checks?check_run_id=1102300442

PR #911, opened 2021-03-12
Handling the traceback in commit message linting

jaindhairyahere:
Commit message linting didn't gracefully exit when met with an error
like absence of body in commit message. Tests have been updated to
handle the raised exception and now the tests exit gracefully with an
error message.

Resolves: #810

Signed-off-by: Dhairya Jain jaindhairya2001@gmail.com

### DIFF
--- a/ci/test_commit_message.py
+++ b/ci/test_commit_message.py
@@ -9,7 +9,18 @@ import os
 import re
 import sys
 
+
 # This script is written to be used with CI Integration
+
+def has_url(string):
+    # findall() has been used
+    # with valid conditions for urls in string
+    regex = r"((?:https?:\/\/|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}\/)" \
+            r"(?:[^\s()<>]+|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))+" \
+            r"(?:\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\)|" \
+            r"[^\s`!()\[\]{};:'\".,<>?«»“”‘’]))"
+    urls = re.findall(regex, string)
+    return bool(len(urls) > 0)
 
 
 def lint_commit(commit_id):
@@ -25,23 +36,33 @@ def lint_commit(commit_id):
 
     # Check 1: Subject, body and DCO exist
     # Note that git does not allow for empty subjects
-    if len(re.split('\n\n|\r', message)) <= 2:
+    msg_list = re.split('\n\n|\r', message)
+    commit_subject = msg_list[0]
+    if len(msg_list) <= 2:
         print("Commit {} does not have a body.".format(sha_short))
         check = False
+    try:
+        # pop the subject and signature
+        msg_list.pop(0)
+        msg_list.pop()
+    except IndexError:
+        pass
 
     # Check 2: Subject length is less than about 50
-    if len(re.split('\n\n', message)[0]) > 54:
+    if len(commit_subject) > 54:
         print(
             "The subject of commit {} should be 50 characters or less.".format(
                 sha_short))
         check = False
 
     # Check 3: Each line of the body is less than 72
-    msg_list = re.split('\n\n|\r', message)
-    msg_list.pop(0)
-    msg_list.pop()
     for msg in msg_list:
         for line in msg.split('\n'):
+            if has_url(line) or line.startswith("[CM]") or line.startswith("[LINK]"):
+                print("Line contains url(s)/compiler messages. Skipping . . .\n"
+                      "Line: {0}\n"
+                      "Commit: {1}\n\n".format(line, sha_short))
+                continue
             if len(line) > 72:
                 print("Line exceeds 72 characters.\n"
                       "Line: {0}\n"


rnjudge:
Hi @jaindhairyahere -- do you have a screen shot or log of the error message you saw? We have code in the `lint_commit()` function that is supposed to check if the commit message has a body, but perhaps it is broken?

```
In file ci/test_commit_message.py :

 26     # Check 1: Subject, body and DCO exist                                      
 27     # Note that git does not allow for empty subjects                           
 28     if len(re.split('\n\n|\r', message)) <= 2:                                  
 29         print("Commit {} does not have a body.".format(sha_short))              
 30         check = False     
```

jaindhairyahere:
@rnjudge Yes, It was broken and incomplete. I didn't realize this before you pointed at this possibility, and I was just targeting to handle the exception generated at the line `msg_list.pop()`. The error message that I saw was exactly the same as in issue #810. I believe that the bug is fixed now.

rnjudge:
Thanks @jaindhairyahere for clarification! This looks good, could you remove the first commit from this PR? I think we only need the second one. On your working branch, run `git rebase -i HEAD~2` and remove the commit that you don't need.

jaindhairyahere:
> Thanks @jaindhairyahere for clarification! This looks good, could you remove the first commit from this PR? I think we only need the second one. On your working branch, run `git rebase -i HEAD~2` and remove the commit that you don't need.

It didn't work with rebase so I reset the last 2 commits and made a force push. I hope it's okay

ci/test_commit_message.py
```diff
@@ -25,21 +25,24 @@ def lint_commit(commit_id):
...
         print(
             "The subject of commit {} should be 50 characters or less.".format(
                 sha_short))
         check = False
 
     # Check 3: Each line of the body is less than 72
-    msg_list = re.split('\n\n|\r', message)
-    msg_list.pop(0)
-    msg_list.pop()
     for msg in msg_list:
```
rnjudge, line 46: This is a corner case scenario, but if we have a commit that is only a message subject and signature, since we wouldn't pop the subject and signature above, this is going to check that the signature line is less than 72 characters, which is not a requirement.

jaindhairyahere: I've already popped subject and signature in lines 35-36 of this file. So this corner case won't fail. The subject is stored in a variable (before popping) which will be checked for length <= 54 but the Signature won't ever be checked.

`if len(msg_list) <= 2:`
        `print("Commit {} does not have a body.".format(sha_short))`
        `check = False`
`else:`
       `# pop the subject and signature`
        `msg_list.pop(0)`
        `msg_list.pop()`

rnjudge: We only pop the subject and signature if `if len(msg_list) <= 2` is false. If there's no body and only a commit message and signature line, we don't enter the `else` statement (since `len(msg_list == 2`). In this case, `msg_list` would still contain the subject and signature.

ci/test_commit_message.py
```diff
@@ -9,8 +9,18 @@
...
+
 # This script is written to be used with CI Integration
 
+def hasUrl(string):
+    # findall() has been used
+    # with valid conditions for urls in string
+    regex = r"(?i)\b((?:https?://|www\d{0,3}[.]|[a-z0-9.\-]+[.]" \
+            r"[a-z]{2,4}/)(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))" \
+            r"+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:'\".,<>?«»“”‘’]))"
+    url = re.findall(regex, string)
```
rnjudge, line 21: When I tested this in the python repl, I got some empty hits with this regex but my URL was detected correctly and the overrall length of url works. Is this expected behavior?

```
>>> msgtxt
'This is a boguz commit message for testing\n\nthis is legit\n\n[4]: https://gist.github.com/JamieMagee/d9cf1ca579b230db504186cbae816f2d#file-distroless-python3-debug-log\n'
>>> regex = r"(?i)\b((?:https?://|www\d{0,3}[.]|[a-z0-9.\-]+[.]" \
...     r"[a-z]{2,4}/)(?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))" \
...     r"+(?:\(([^\s()<>]+|(\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:'\".,<>?«»“”‘’]))"
>>> 
>>> url = re.findall(regex, msgtxt)
>>> url
[('https://gist.github.com/JamieMagee/d9cf1ca579b230db504186cbae816f2d#file-distroless-python3-debug-log', '', '', '', '')]
>>> len(url)
1

```

rnjudge: Even with a simple URL like `https://www.google.com` I see the four additional blank strings in the tuple in the `url` list.

sarc-iitb: Yes, this is the expected behavior. Should I create a regex that doesn't produce the additional blank strings?

jaindhairyahere: Yes, this is the expected behavior. Should I create a regex that doesn't produce the additional blank strings?

nishakm:
@jaindhairyahere Any update on this?

sarc-iitb:
> @jaindhairyahere Any update on this?

It's ready from my side. Just waiting for reviews.

jaindhairyahere:
> @jaindhairyahere Any update on this?

It's ready from my side. Just waiting for reviews.

ci/test_commit_message.py
nishakm: I don't think the regex should be matching blank strings. It also looks like something you would have gotten from the internet :). I would suggest you try to use `re.compile` and see if you can match that URL.

ci/test_commit_message.py
```diff
@@ -9,8 +9,18 @@
 import re
 import sys
 
+
 # This script is written to be used with CI Integration
 
+def hasUrl(string):
```
nishakm, line 15: Small nit: Please change this to `has_url`.

nishakm (changes requested):
@jaindhairyahere If you could squash your two commits, I would appreciate it. I've also asked for some changes to the code. Thanks for sticking to this!

ci/test_commit_message.py
jaindhairyahere: Done

ci/test_commit_message.py
jaindhairyahere: It was because of the presence of 4 capturing groups in the regex. I've converted the last 3 to non-capturing groups

jaindhairyahere:
> @jaindhairyahere If you could squash your two commits, I would appreciate it. I've also asked for some changes to the code. Thanks for sticking to this!

I think that it's ready with the suggested changes. Please check them once.

nishakm (approved):

nishakm:
@jaindhairyahere Thanks a lot for the fix. Merging.

PR #915, opened 2021-03-17
Update community meeting time to UTC time zone

rnjudge:
Listing the community meeting time in the "Pacific" timezone was
confusing during daylight savings. Updated the meeting time to UTC time
zone so we will have a consistent meeting time and folks from around the
globe won't get thrown off by daylight savings. US timezones will still
need to account for the hour change twice a year depending on where
they are located.


CarAuto34:
Ty

CarAuto34:
Ty

CarAuto34:
Ty

CarAuto34:
Ty

Issue #888, opened 2021-02-19
Use unshare -mpf for rootless inventory


nishakm:
**Describe the Feature**
We could possibly inventory without using `sudo` by using `buildah unshare`, or use something similar (https://man7.org/linux/man-pages/man1/unshare.1.html)

**Use Cases**
- running tern without `sudo` mitigates exploits like: https://access.redhat.com/security/cve/cve-2019-14287
- we could possibly run tern in an unprivileged container?

**Implementation Changes**
Basically, you would have to run `buildah unshare` on a script that mounts the working directory, runs the inventory script, and then unmounts. There is an `unshare -pf` that is only restricted to running the inventory script, and it may work when mount, inventory and unmount are all in one command. The implementation is much easier if we use buildah.

nishakm:
#938 has an implementation for this. We can migrate to that method of inventory if it is possible to do it. The benefits are that one doesn't have to run tern as root. @rnjudge

rnjudge:
@nishakm rootless inventory with buildah would also resolve https://github.com/tern-tools/tern/issues/261. Do you want me to take a look at this? If we're going to use `unshare -mpf` for live inventorying, let's use it for post-build inventorying as well.

nishakm:
@rnjudge Yup! This is probably a dupe of #261. Generally, we want to not use `sudo`.

nishakm:
If I recall correctly: We will not be able to use `unshare` to mount overlayfs but will be able to do so once kernel version 5.12 is out.

nishakm:
Going to close this for now as we are able to inventory in an unprivileged container and the live inventory uses `unshare`.

PR #932, opened 2021-03-21
Use Debin Buster slim as Docker base image

JamieMagee:
Using Alpine Linux as the docker base image reduces the resultant image
from 681MB to 67MB

I've tested this with `docker run --privileged --rm --device /dev/fuse -v /var/run/docker.sock:/var/run/docker.sock tern report --docker-image ...`, but would appreciate input for more scenarios I haven't covered.

I would also appreciate input from @JeroenKnoops for the scancode variant of this image. Using this Alpine-based Dockerfile and adding `scancode-toolkit`, I was unable to build the image. It gets stuck resolving dependencies for scancode-toolkit. I can see that you've opened some issues in the scancode-toolkit repository, it appears that those haven't been completely resolved yet.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,38 +1,31 @@
 # Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-FROM debian:buster
+FROM python:3.9-slim-buster as base
 
-# Install fuse-overlayfs and Tern dependencies
-RUN apt-get update && \
-    apt-get -y install \
-    attr \
-    findutils \
-    git \
-    gnupg2 \
-    jq \
-    python3 \
-    python3-pip \
-    python3-setuptools \
-    tar \
-    util-linux \
-    wget && \
-    echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && \
-    wget --no-verbose https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_10/Release.key -O - | apt-key add - && \
-    apt-get update && \
-    apt-get -y install \
-    buildah \
-    fuse-overlayfs && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+FROM base as builder
 
-# Adjust storage.conf to enable Fuse storage.
-RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
+RUN mkdir /install
+WORKDIR /install
 
-# Install tern
-RUN pip3 install --upgrade pip && \
-    pip3 install --no-cache-dir \
+RUN pip install --no-warn-script-location --prefix=/install \
     tern
 
+FROM base
+
+RUN echo "deb http://deb.debian.org/debian bullseye main" > /etc/apt/sources.list.d/bullseye.list \
+    && echo "Package: *\nPin: release n=bullseye\nPin-Priority: 50" > /etc/apt/preferences.d/bullseye \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+    attr \
+    findutils \
+    fuse-overlayfs/bullseye \
+    fuse3/bullseye \
+    git \
+    jq \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /install /usr/local
+
 ENTRYPOINT ["tern", "--driver", "fuse"]
-CMD ["-h"]
+CMD ["--help"]
--- a/docker/Dockerfile.scancode
+++ b/docker/Dockerfile.scancode
@@ -1,39 +1,38 @@
 # Copyright (c) 2019-2020 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
-FROM debian:buster
+FROM python:3.9-slim-buster as base
 
-# Install fuse-overlayfs and Tern dependencies
-RUN apt-get update && \
-    apt-get -y install \
-    attr \
-    findutils \
-    git \
-    gnupg2 \
-    jq \
-    python3 \
-    python3-pip \
-    python3-setuptools \
-    tar \
-    util-linux \
-    wget && \
-    echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && \
-    wget --no-verbose https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/Debian_10/Release.key -O - | apt-key add - && \
-    apt-get update && \
-    apt-get -y install \
-    buildah \
-    fuse-overlayfs && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+FROM base as builder
 
-# Adjust storage.conf to enable Fuse storage.
-RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
+RUN mkdir /install
+WORKDIR /install
 
-# Install tern
-RUN pip3 install --upgrade pip && \
-    pip3 install --no-cache-dir \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-warn-script-location --prefix=/install \
     tern \
     scancode-toolkit[full]
 
+FROM base
+
+RUN echo "deb http://deb.debian.org/debian bullseye main" > /etc/apt/sources.list.d/bullseye.list \
+    && echo "Package: *\nPin: release n=bullseye\nPin-Priority: 50" > /etc/apt/preferences.d/bullseye \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+    attr \
+    findutils \
+    fuse-overlayfs/bullseye \
+    fuse3/bullseye \
+    git \
+    jq \
+    libgomp1 \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /install /usr/local
+
 ENTRYPOINT ["tern", "--driver", "fuse"]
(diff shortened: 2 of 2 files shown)


JeroenKnoops:
I'm looking into this. Trying to install it from the scancode-toolkit releases.

nishakm:
@JamieMagee I'm hesitant to accept Alpine as a base image because:

1. `musl` and `busybox` do not have a 1:1 parity with `binutils` so the shell implementations we have in the command library may not behave the same way in the container as they would on an ubuntu or fedora OS
2. Eventually, we want to be able to deploy container images along with each release and we would like to be able to provide the corresponding sources for everything in the container images (Tern is primarily an OSS compliance tool). Alpine makes it difficult to trace back the binaries to the sources. Debian has a built-in mechanism to do that.

JamieMagee:
I understand the hesitance, but hopefully I can answer your concerns:
1. `musl` should be [feature compatible with `glibc`](https://www.musl-libc.org/faq.html). We've been running an alpine-based image in production for a couple of weeks now, and haven't seen any crashes that have been caused by `musl`. Yes, there are differences between `busybox` and `coreutils` (I couldn't find any uses of `binutils` in Tern, but correct me if I'm wrong) but individual GNU versions of utilities can be installed, as well as [`coreutils` itself](https://pkgs.alpinelinux.org/package/v3.13/main/x86_64/coreutils). I've only had to add a dependency on GNU `grep` due to the need for PCRE regexes as of Tern 2.5.0
2. Can you clarify what you mean? I know with `apt-get` you can run `apt-get download --print-uris coreutils` which gives:
	```sh
	$ apt-get download --print-uris coreutils
	'http://deb.debian.org/debian/pool/main/c/coreutils/coreutils_8.30-3_amd64.deb' coreutils_8.30-3_amd64.deb 2707932 SHA256:ae6e5cd6e9aaf74d66edded3931a7a6c916625b8b890379189c75574f6856bf4
	```

	Alpine does have `apk info --quiet --webpage`, but the output provides the upstream URL

	```sh
	$ apk info --quiet --webpage zlib
	apt-get download --print-uris coreutils
	zlib-1.2.11-r3 webpage:
	https://zlib.net/
	
	```

nishakm:
>     1. `musl` should be [feature compatible with `glibc`](https://www.musl-libc.org/faq.html).

That link seems to be broken? I'll take a look when it goes back online.

> 
>     2. Can you clarify what you mean? I know with `apt-get` you can run `apt-get download --print-uris coreutils` which gives:
>        ```shell
>        $ apt-get download --print-uris coreutils
>        'http://deb.debian.org/debian/pool/main/c/coreutils/coreutils_8.30-3_amd64.deb' coreutils_8.30-3_amd64.deb 2707932 SHA256:ae6e5cd6e9aaf74d66edded3931a7a6c916625b8b890379189c75574f6856bf4
>        ```
>        
>        
>        Alpine does have `apk info --quiet --webpage`, but the output provides the upstream URL
>        ```shell
>        $ apk info --quiet --webpage zlib
>        apt-get download --print-uris coreutils
>        zlib-1.2.11-r3 webpage:
>        https://zlib.net/
>        ```

You can use [Debian source mirrors](https://github.com/tern-tools/tern/blob/main/tern/scripts/debian/apt_get_sources.sh) to download the corresponding source code for the packages that are included in a Debian image. Furthermore, the Debian image is probably based on a [snapshot](http://snapshot.debian.org/) and could be reproduced using debian tools. Alpine doesn't have the same or similar machinery.

I might as well bring up the fact that Alpine doesn't maintain a security advisory database like Debian, and a number of their container images seem to not have the security fixes. There is also a compliance risk for anyone consuming Tern container images (many companies have internal policies forbidding the use of busybox and are vicariously suspicious of the OS itself).

cc: @jku @joshuagl

nishakm:
@JamieMagee I would have been OK with using `Distroless` instead, but it's a PITA to find the sources for that too. There may be a way to automate it though. That would probably be a reasonable compromise of small size and all the other nice things.

JamieMagee:
Okay, I'm with you now.

I think the same thing is possible with Alpine, thought admittedly it's not as polished. To get the sources of `musl` for `alpine:3.13.2`

```sh
$ apk add --update alpine-sdk
$ git clone --depth 1 --branch v3.13.2 git://git.alpinelinux.org/aports
$ cd aports/main/musl
$ abuild -F fetch verify
$ ls src/
__stack_chk_fail_local.c  getent.c                  iconv.c                   musl-v1.2.2.tar.gz
getconf.c                 handle-aux-at_base.patch  ldconfig                  revert-faccessat2.patch
```

Using the aports git repository I can also build a rootfs, or any package version. However, Alpine doesn't keep around compiled binaries like debian does.

Alpine does have a security advisory database at https://secdb.alpinelinux.org/. The main difference between Alpine's database and others is that Alpine only releases advisories for vulnerabilities that they've fixed. Looking at their GitLab [issues page for security vulnerabilities](https://gitlab.alpinelinux.org/alpine/aports/-/issues?scope=all&utf8=%E2%9C%93&state=opened&label_name[]=tag%3Asecurity) they do stay relatively on top of things.

I had also investigated Distroless as a possibility, but unfortunately [their Python images](https://github.com/GoogleContainerTools/distroless#docker) are still in preview ☹️

jku:
With regards to fetching sources:

> I think the same thing is possible with Alpine, thought admittedly it's not as polished. 

It ends up being more complicated than it at first seems:
* your method ends up fetching sources that match the _current_ contents of the v3.13.2 branch, not your container image -- now typically you should be able to find a single commit that _should_ match the contents of your container base image ... but I'll point out that that's not quite how the alpine (or most linux distro) build processes work: e.g. the musl package was actually built a month before v3.13.2 from aports commit c6c2cd54. Contrast this relative uncertainty with Debian storing the sources they used and making them easily available
* nothing guarantees or checks that the upstream sources are still available when you need them. In fact some upstreams like openssl inexplicably like to move their sources around so in practice it's not uncommon that `abuild fetch` fails when run after an Alpine release

Working around these issues is certainly possible but it likely ends up being a partly manual process. Also, for some people having sources is just not that important... Others, however have policies that state that sources are required and in this case adding hacks and manual processes definitely feels like a chore (when other solutions don't really require them).

> Alpine does have a security advisory database at https://secdb.alpinelinux.org/

This is actually nice, thanks for the link (I had done the git grepping manually before). I don't want to make this into a bigger difference than it is in practice but these aren't _advisories_: they don't give any advice and as you said they only list CVEs that are marked as fixed in aports git. In practice this may be fine but ... not quite the same thing

JamieMagee:
Listening to the feedback I've received so far I've rewritten my changes using the `python:3.9-slim-buster` image instead. I've still managed to simplify the image, and reduce the size from 681MB to 208MB.

I succeeded in getting the scancode variant of the image building this time, but I would still appreciate input from @JeroenKnoops here.

Additionally, I removed the `buildah` dependency. I couldn't find it used anywhere in Tern, but maybe it's preparatory work for #888? Not sure.

rnjudge:
> Additionally, I removed the `buildah` dependency. I couldn't find it used anywhere in Tern, but maybe it's preparatory work for #888? Not sure.

The buildah requirement was introduced when we added support for non-linux hosts with fuse-overlayfs. I think we need it for that but you could try without it
? (or I can try later) https://github.com/tern-tools/tern/pull/771/commits/cf1c707ad0ba178b99a7e00f8955ae79223c41f9

JamieMagee:
I am able to run tern run this Dockerfile on Windows via WSL2, but I don't have access to a Mac to test. Is that what you mean?

JeroenKnoops:
I will test the following setups:

- [ ] new smaller image with various docker images default output
- [ ] new smaller image with various docker images spdxtagvalue output
- [ ] new smaller image with scancode with various docker images default output
- [ ] new smaller image with scancode with various docker images spdxtagvalue output
- [ ] new smaller image in a github-action which I think uses fuse-overlayfs

docker/Dockerfile.scancode
```diff
@@ -1,39 +1,36 @@
...
+
+RUN echo "deb http://deb.debian.org/debian bullseye main" > /etc/apt/sources.list.d/bullseye.list \
+    && echo "Package: *\nPin: release n=bullseye\nPin-Priority: 50" > /etc/apt/preferences.d/bullseye \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+    attr \
+    findutils \
+    fuse-overlayfs/bullseye \
+    git \
+    jq \
```
JeroenKnoops, line 30:
```suggestion
    jq \
    libgomp1 \   
```
Necessary for scancode..  Without it gives an error:

```
...
OSError: (\'libgomp.so.1: cannot open shared object file: No such file or directory\'
....
```

JeroenKnoops:
I've scanned `debian:buster` with and without scancode and they all are fine.

When scanning f.e. `rust:1.46.0` I get an error: `ERROR - rootfs - Command failed. fuse: warning: library too old, some operations may not work`

I'm going to dive into this..

JamieMagee:
I suspect that error message is due to the `fuse-overlayfs` being brought in from Debian Bullseye. Comparing the dependencies for `fuse-overlayfs` for [Buster](https://packages.debian.org/buster/fuse-overlayfs) vs [Bullseye](https://packages.debian.org/bullseye/fuse-overlayfs), neither specifies a version of the `fuse3` dependency. However, if you look at the version of `fuse3` that ships in [Buster](https://packages.debian.org/buster/fuse3) vs [Bullseye](https://packages.debian.org/bullseye/fuse3) it's `3.4.1-1+deb10u1` and `3.10.2-2` respectively.

This might be fixed by explicitly installing `fuse3/bullseye`

docker/Dockerfile
```diff
@@ -1,38 +1,31 @@
...
+    && apt-get install -y --no-install-recommends \
+    attr \
+    findutils \
+    fuse-overlayfs/bullseye \
+    fuse3/bullseye \
+    git \
+    jq \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /install /usr/local
```
nishakm, line 28: Interesting - is using multiple stages like this meant to reduce build time?

JamieMagee: Usually, yes that would be the main benefit. But my goal here was to isolate the compilation stage from the image that runs.

For this specific Dockerfile there's not much benefit. There is a larger benefit for the scancode variant, and I'm trying to keep the two images consistent.

JeroenKnoops:
I've tested the latest version of Jamie's code and it works great and the image size is a lot smaller!

nishakm (approved):

nishakm:
Thanks for your work @JamieMagee and your testing help @JeroenKnoops! I will merge this.

PR #953, opened 2021-04-20
Fixed an issue with the export command on some images

quepop:
Running:
`tern report -i nexus3.onap.org:10001/onap/vid -f json -o vid.json` 
resulted in an error: 
`export: Bad variable name` 
because tern tried to run the export command with spaces inside the variable value.
Adding quotation marks fixed the bug.


nishakm:
@quepop Thanks for the bug fix! Could you amend your commit message to include your explanation in the body? Otherwise, this change looks good to me.

Issue #961, opened 2021-05-07
Dependency "debut" is now "debian-inspector"


cole-h:
**Description**
Replace usage of `debut` with` debian-inspector`

**Implementation**
* Replace `debut` with `debian-inspector` in `requirements.{in,txt}`
* Replace `debut` at https://github.com/tern-tools/tern/blob/3710b08a5e4f36670cc02061419dce5e5cf56b09/tern/analyze/common.py#L22-L23

**References**
https://github.com/nexB/debian-inspector/commit/0978bd5a214b4eb8e8f0c38ccb3893a08f4e40eb

**Super Issues**
N/A

PR #962, opened 2021-05-07
Replace `debut` with `debian-inspector`

cole-h:
The debut library was renamed to debian-inspector.

https://github.com/nexB/debian-inspector/commit/0978bd5a214b4eb8e8f0c38ccb3893a08f4e40eb

---

Fixes https://github.com/tern-tools/tern/issues/961.


cole-h:
Ah, whoops, I must've missed that part, sorry!

PR #981, opened 2021-06-01
Introduce JSON consumer for package diffs

nishakm:
This PR does the following:
1. Introduces the Consumer class under the formats plugin.
2. Implements the JSON consumer which demonstrates consuming
    an SBOM of the same format from a previous execution at
    container build time.
3. Adds a command line option for entering SBOMs from previous
    runs using the same format.

The option must be used in conjunction with the --live option as
follows:
1. Generate an SBOM for a starting filesystem:
     `tern report --live <path to filesystem> -f json -o sbom1`
2. Generate an SBOM for a reused filesystem:
     `tern report --live <path to filesystem> -f json -ctx sbom1 -o sbom2`

Generating SBOMs in this way allows downstream container
builders to reuse existing SBOMs rather than generating them from
scratch every time, thus creating sharable SBOMs.


tern/__main__.py
```diff
@@ -208,6 +208,13 @@ def main():
                                help="Generate a report for the current state "
                                "of the filesystem. This is useful when "
                                "running Tern in a new namespace or VM.")
+    parser_report.add_argument('-ctx', '--with-context', metavar='FILES',
+                               nargs='+',
+                               help="When generating a report, take into "
+                               "account data collected from a previous run "
+                               "without including that data in the report. "
+                               "This is useful when running Tern multiple "
```
rnjudge, line 216: Do you want to add anything here about how folks can string together multiple SBoMs or the fact that they must be ordered for Tern to read them or is this something that you are going to write out a longer explanation for in the README?

rnjudge: Another question: Do you want to clarify that the reports provided should only be for one layer or can Tern handle reports that represent multiple layers?

nishakm: Good points. Let me update this message to be more clear.

tern/__main__.py
```diff
@@ -208,6 +208,17 @@ def main():
...
+                               nargs='+',
+                               help="""Generate a report taking into account
+                               reports generated by previous runs against
+                               the same filesystem. This option is meant to
+                               be used with the --live option like this:
+                               tern report --live <path/to/filesystem> -ctx
+                               report1 report2 report3... -f json -o report4.
+                               The order of the reports must correspond to the
+                               order in which they were generated. Supported
+                               input report formats are: json""")
```
nishakm, line 221: @rnjudge Does this look better?

tern/__main__.py
rnjudge: @nishakm looks perfect! LGTM

nishakm:
@rnjudge OK to merge?

PR #1004, opened 2021-07-15
added pkg values for pip, pip3, gem, npm, and go

nsapse:
Thanks for being a great guide, Nisha. First contributions to open source can feel really scary. Let me know if there are any issues.


nsapse:
Will resubmit with proper commit message and go_lang values.

Issue #948, opened 2021-04-13
Use skopeo to pull container images


rnjudge:
**Background**
Reference: https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/ 

TL;DR - Kubernetes currently uses Dockershim to access containerd container runtime. Dockershim is being removed from Kubelet which removes support for Docker as a container runtime in favor of other runtimes that use [Container Runtime Interface](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) created for K8s.

We would like Tern to run in Kubernetes. This means that we cannot use Docker commands to pull images from the registry with the`-i` command line option if we want Tern to run in Kubernetes after v1.2.0. 

**Proposal**
Let's try to use `subprocess` + [ORAS](https://github.com/deislabs/oras) instead of the Docker python library to pull container images from a registry when running with the -i option. This will mean that we need to include the go module as a dependency of Tern. 
* Create an OCI image class and implement the `load_image(`) method for it in accordance with how ORAS pulls images.
* We may need an additional tool on top of ORAS. ORAS will send all the bytes of the downloaded image to stdout and we need to figure out how we store these results to analyze.

**Use Cases**
* Deploying Tern as a k8s cluster on a CRD

rnjudge:
It does not appear that ORAS supports anonymous pulling of container images from Dockerhub at this time. Skopeo does support this behavior and therefore, we should explore using skopeo instead. A previous PR (https://github.com/tern-tools/tern/pull/712) started to work on this and we may want to revive parts of this.

falcorocks:
+1 on skopeo, I have used it to successfully achieve something similar to what you are trying to do here.
Abandoning docker in favour of something like skopeo would also make it much better for CI, where docker in docker is still a nogo for security reasons!

rnjudge:
@mukultaneja is working on this. @falcorocks if you have any tips or pointers, feel free to drop them here :)

mukultaneja:
@rnjudge @nishakm, how are we going to make `skopeo` as a requirement to tern moving onwards?

PR #1006, opened 2021-07-15
Using OCI images instead of Docker Images

mukultaneja:
Start analyzing OCI images using tern

Work towards: #948

Signed-off-by: Mukul Taneja <mtaneja@vmware.com>


mukultaneja:
@nishakm @rnjudge, Initial commit towards Issue #948. I would keep updating the same PR with small changes and request you to review them step by step.

tern/analyze/default/container/image.py
```diff
@@ -23,9 +23,16 @@
 logger = logging.getLogger(constants.logger_name)
 
 
+def download_container_image(image_tag_string):
+    '''Download the docker image and convert it into
+    oci format'''
+    pass
```
nishakm, line 29: Prospector says this is an unnecessary pass statement. So you can remove it.

tern/analyze/oci/parse.py
```diff
@@ -0,0 +1,50 @@
...
+
+def get_oci_image_layers(image_string, manifest):
+    ''' Returns OCI image layers '''
+    blob_path = os.path.join(image_string, "blobs/sha256")
+    layers = list()
+    layer_data = json.load(open(manifest)).get("layers")
+    for layer in layer_data:
+        layer_path = os.path.join(blob_path, layer.get("digest").split(":")[1])
+        layers.append(layer_path)
+    return layers
```
nishakm, line 50: Nice!

nishakm (changes requested):
Sorry for reviewing this so late! Thanks for continuing to work on this! It looks good to me. I wonder if you can add some tests for the OCIImage class?
Also, there are some prospector errors that come up, but seem to be easy to resolve. Let me know if you're stuck with anything.

mukultaneja:
@nishakm @rnjudge, Please review this PR. I have made the required changes and tested with the below command
`tern -c report -i docker://debian:latest`.

tern/analyze/default/container/image.py
```diff
@@ -23,10 +23,29 @@
...
+    try:
+        # extract the docker image
+        docker_image = image_tag_string
+        if 'docker://' in image_tag_string:
+            docker_image = image_tag_string.split('docker://')[1]
+        image_attr = general.parse_image_string(docker_image)
+        oci_image = 'oci://{0}/{1}'.format(
+            rootfs.working_dir, image_attr.get('name'))
+        docker_image = 'docker://{0}'.format(docker_image)
+        process = subprocess.Popen(['skopeo', 'copy', docker_image, oci_image])
```
nishakm, line 37: Any reason why picking `subprocess.Popen` rather than `subprocess.check_output`?

nishakm (changes requested):
@mukultaneja Sorry for taking so long to get back to you on this. I have one comment on the implementation as it is causing a problem with the security linter. Also, can you rebase your change so you can pick up the latest prospector changes?

tern/analyze/default/container/image.py
mukultaneja: `Popen` runs commands in a separate process ( child process) instead of running as part of the same process so It is kind of like a less burden and a little secure I feel to the current process.

nishakm: This is actually an open execution path. `process.wait()` without detecting the end of execution may cause some non-determinism. My suggestion for now is to use `subprocess.check_output` so we are sure the process has ended before we go poking at the downloaded files. It should also resolve the security linting errors.

In the future, why not try `asyncio` or `multiprocessing` modules to reduce tern's execution time? There are plenty of places where we can apply async and parallel processing :)

mukultaneja: I can surely use `subprocess.check_output` as you suggested but using `process.wait()` let the parent process wait for the child process to finish first. It is a purely deterministic flow here. Maybe I can add a timeout to `process.wait()` so that it makes sure until this timeout the execution should get stopped else raise a Timeout exception.
Anyways, let me use `subprocess.check_output` so that I can resolve security linting issues.

mukultaneja:
@nishakm, I tried to get away with `security issue` by having `shell==True / False` but seems the linter does not allow to use `subprocess` at all. Any suggestions?

tern/analyze/default/container/image.py
```diff
@@ -10,9 +10,9 @@
 import docker
 import logging
 import subprocess  # nosec
-
+from tern.utils import general, rootfs
```
nishakm, line 13: Could you use a separate line for each import?

tern/analyze/default/container/image.py
```diff
@@ -23,10 +23,29 @@
...
+    try:
+        # extract the docker image
+        docker_image = image_tag_string
+        if 'docker://' in image_tag_string:
+            docker_image = image_tag_string.split('docker://')[1]
+        image_attr = general.parse_image_string(docker_image)
+        oci_image = 'oci://{0}/{1}'.format(
+            rootfs.working_dir, image_attr.get('name'))
+        docker_image = 'docker://{0}'.format(docker_image)
+        cmd = 'skopeo copy {0} {1}'.format(docker_image, oci_image)
```
nishakm, line 37: `tern.utils.rootfs` has a `shell_command` function that will take a command like `["skopeo", "copy"]` and some extra arguments like `docker_image` and `oci_image` and invoke it using `subprocess`. In `rootfs.py` the security linting error is suppressed, so the linter will not complain. Why not try this method?

tern/classes/oci_image.py
```diff
@@ -0,0 +1,149 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2017-2020 VMware, Inc. All Rights Reserved.
```
nishakm, line 3: I believe the copyright here is just 2021

nishakm (changes requested):
@mukultaneja I've added a suggestion. I will also try to test the changes.

mukultaneja:
@nishakm I made all the requested changes and this time all tests got passed too :)

tern/analyze/default/container/image.py
```diff
@@ -23,10 +24,30 @@
...
+    '''Download the docker image and convert it into oci format'''
+    try:
+        # extract the docker image
+        docker_image = image_tag_string
+        if 'docker://' in image_tag_string:
+            docker_image = image_tag_string.split('docker://')[1]
+        image_attr = general.parse_image_string(docker_image)
+        oci_image = 'oci://{0}/{1}'.format(
+            rootfs.working_dir, image_attr.get('name'))
+        docker_image = 'docker://{0}'.format(docker_image)
```
nishakm, line 37: I just found out that if we use `docker://` rather than `dockerdaemon` we encounter a TLS error on the registry side. I also found that [crane](https://github.com/google/go-containerregistry/) is a much more lightweight tool to pull container images, and conforms with OCI's on-disk layout spec. @rnjudge Do you want to merge this PR and then let @mukultaneja update it to use crane?

nishakm (commented):
@rnjudge Need your input on this PR :)

nishakm:
Hi @mukultaneja, sorry again for the delay in reviewing! I tried running tern with these changes. I get the following error:
```
2021-11-30 07:13:17,531 - DEBUG - rootfs - Running command: skopeo copy docker://debian:buster oci:///home/nisha/.tern/debian
Traceback (most recent call last):
  File "/home/nisha/terndev/bin/tern", line 10, in <module>
    sys.exit(main())
  File "/home/nisha/terndev/tern/tern/__main__.py", line 287, in main
    do_main(args)
  File "/home/nisha/terndev/tern/tern/__main__.py", line 116, in do_main
    crun.execute_image(args)
  File "/home/nisha/terndev/tern/tern/analyze/default/container/run.py", line 65, in execute_image
    full_image = cimage.load_full_image(
  File "/home/nisha/terndev/tern/tern/analyze/default/container/image.py", line 54, in load_full_image
    test_image.load_image(load_until_layer)
  File "/home/nisha/terndev/tern/tern/classes/oci_image.py", line 130, in load_image
    self._image_id = self.get_image_id(self._manifest)
AttributeError: 'OCIImage' object has no attribute 'get_image_id'
```
Also, could you please rebase your changes on top of main?

nishakm:
@mukultaneja few other requests:
- Could you give more descriptive titles for each of your commits? I see each one having: "Using OCI images instead of Docker Images"
- Could you add a commit with documentation changes to https://github.com/tern-tools/tern#getting-started-on-linux and changes to the various dockerfiles and the Vagrantfile to install skopeo?

nishakm:
Hi @mukultaneja. I have included your changes in #1090. Please take a look. I will close this PR once the more updated one is merged. Thanks again for your work on this!

Issue #989, opened 2021-06-28
debian package licenses not reported in default report table


rnjudge:
**Describe the bug**
When you generate a default report for debian images, not license information is reported in the table. This is because the table collects Package object `license` information but not `pkg_licenses` information where the debian package information is stored. 

**To Reproduce**
1. `tern report -i debian:bullseye`

**Error in terminal**
```
	Packages found in Layer: 
	+------------------------+------------------------------+---------+
	| Package                | Version                      | License |
	+------------------------+------------------------------+---------+
	| adduser                | 3.118                        |         |
	| apt                    | 2.2.2                        |         |
	| base-files             | 11                           |         |
	| base-passwd            | 3.5.49                       |         |
	| bash                   | 5.1-2+b1                     |         |
	| bsdutils               | 1:2.36.1-7                   |         |
	| coreutils              | 8.32-4+b1                    |         |
	| dash                   | 0.5.11+git20200708+dd9ef66-5 |         |
	| debconf                | 1.5.75                       |         |
	| debian-archive-keyring | 2021.1.1                     |         |
	| debianutils            | 4.11.2                       |         |
	| diffutils              | 1:3.7-5                      |         |

###########################################
# Summary of licenses found in Container: #
###########################################
[...]

PR #1029, opened 2021-08-16
Report Debian package licenses in default report

mukultaneja:
Adds a new column 'Pkg Licenses' in the table
to report package licenses as part of default
report

Resolved: #989

Signed-off-by: Mukul Taneja <mtaneja@vmware.com>

### DIFF
--- a/tern/formats/default/generator.py
+++ b/tern/formats/default/generator.py
@@ -9,6 +9,7 @@ default report generator
 
 import logging
 
+from textwrap import wrap
 from tern.report import formats
 from tern.formats import generator
 from tern.report import content
@@ -101,7 +102,8 @@ def get_layer_info_list(layer):
     layer_file_licenses_list = []
     file_level_licenses = None
     package_list = PrettyTable()
-    package_list.field_names = ["Package", "Version", "License", "Pkg Format"]
+    package_list.field_names = ["Package", "Version", "License(s)",
+                                "Pkg Format"]
     package_list.align = "l"
     package_list.print_empty = False
 
@@ -113,8 +115,17 @@ def get_layer_info_list(layer):
         file_level_licenses = ", ".join(layer_file_licenses_list)
 
     for package in layer.packages:
+        pkg_licenses = [package.pkg_license]
+        if package.pkg_format == 'deb':
+            # If many pkg_licenses, wrap the line for the table at 45 chars
+            pkg_licenses = wrap(", ".join(package.pkg_licenses),
+                                width=45) or ['']
         package_list.add_row([package.name, package.version,
-                              package.pkg_license, package.pkg_format])
+                              pkg_licenses[0], package.pkg_format])
+
+        # Each wrapped line must be added as a new row under the same package
+        for pkg_license in pkg_licenses[1:]:
+            package_list.add_row(['', '', pkg_license, ''])
 
     return file_level_licenses, package_list.get_string()
 


tern/formats/default/generator.py
```diff
@@ -101,7 +101,8 @@ def get_layer_info_list(layer):
     layer_file_licenses_list = []
     file_level_licenses = None
     package_list = PrettyTable()
-    package_list.field_names = ["Package", "Version", "License", "Pkg Format"]
+    package_list.field_names = ["Package", "Version", "License", "Pkg Format",
+                                "Pkg Licenses"]
```
rnjudge, line 105: Hmm, I'm not sure we should add a separate `pkg_licenses` column. I suggest changing the existing `License` to `License(s)`.

mukultaneja: @rnjudge makes sense. Just updated the PR.

rnjudge:
This works, but we may need to re-think this as it completely messes up the formatting to be unreadable with the default format -- something I was not expecting!! Let me (or if you want to, feel free @mukultaneja ) take a look at the PrettyTable python library and see if there's any way to wrap cells.

mukultaneja:
@rnjudge, Please review the recent changes where I am using `textwrap` python module to wrap the cell value. After this change I am getting the report like below [ a brief glance ]

```
	| ncurses-base           | 6.2+20201114-2               | X11, BSD-3-clause, MIT/X11                                             | deb        |
	| ncurses-bin            | 6.2+20201114-2               | X11, BSD-3-clause, MIT/X11                                             | deb        |
	| passwd                 | 1:4.8.1-1                    |                                                                        | deb        |
	| perl-base              | 5.32.1-4+deb11u1             | GPL-1+ or Artistic, and BSD-4-clause-POWERDOG, GPL-1+ or Artistic or   | deb        |
	|                        |                              | Artistic-dist, RRA-KEEP-THIS-NOTICE, Expat or GPL-1+ or Artistic,      |            |
	|                        |                              | DONT-CHANGE-THE-GPL, Expat, TEXT-TABS, BZIP, ZLIB, Artistic-dist,      |            |
	|                        |                              | GPL-2+, BSD-4-clause-POWERDOG, Artistic-2, REGCOMP, Artistic, GPL-1+   |            |
	|                        |                              | or Artistic, and Unicode, LGPL-2.1, GPL-2+ or Artistic, GPL-1+,        |            |
	|                        |                              | BSD-3-clause, SDBM-PUBLIC-DOMAIN, CC0-1.0, Unicode, GPL-3+-WITH-BISON- |            |
	|                        |                              | EXCEPTION, Artistic or GPL-1+ or Artistic-dist, BSD-3-clause-with-     |            |
	|                        |                              | weird-numbering, REGCOMP, and GPL-1+ or Artistic, GPL-1+ or Artistic,  |            |
	|                        |                              | GPL-1+ or Artistic, and Expat, HSIEH-DERIVATIVE, HSIEH-BSD,            |            |
	|                        |                              | BSD-3-clause-GENERIC, GPL-1+ or Artistic, and BSD-3-clause-GENERIC     |            |
	| sed                    | 4.7-1                        |                                                                        | deb        |```

**Comment by nishakm** [comment]
@rnjudge Is this good to go?

**Comment by rnjudge** [comment]
@mukultaneja did you have a chance to check if the PrettyTable library has a way to wrap text in cells?

**Comment by mukultaneja** [comment]
@rnjudge, I tried that but did not get many options except `set_style` but this does not support data wrapping as we are looking for here.

File: tern/formats/default/generator.py [hunk 2]
```diff
@@ -113,8 +114,14 @@ def get_layer_info_list(layer):
...
     for package in layer.packages:
+        pkg_licenses = package.pkg_license
+        if package.pkg_format == 'deb':
+            pkg_licenses = wrap(", ".join(package.pkg_licenses)) or ['']
         package_list.add_row([package.name, package.version,
-                              package.pkg_license, package.pkg_format])
+                              pkg_licenses[0], package.pkg_format])
+
+        for pkg_license in pkg_licenses[1:]:
+            package_list.add_row(['', '', pkg_license, ''])
```
rnjudge, line 124: I think we can get rid of this additional loop by making changes to how we define pkg_licenses. See comment above.

rnjudge:
Hi @mukultaneja, sorry for the delay. I tested this with a non-debian image and I get an issue with the output. Each letter of the license is printed on a separate line. For photon:3.0: 

Screen Shot 2021-11-02 at 8 00 40 PM

I think I know the fix, will comment in-line. Would also like you to add the following change in line 105 of default/generator.py file:
```
-  package_list.field_names =  ["Package", "Version", "License", "Pkg Format"]
+  package_list.field_names = ["Package", "Version", "License(s)", "Pkg Format"]
```

tern/formats/default/generator.py
```diff
@@ -113,8 +114,14 @@ def get_layer_info_list(layer):
         file_level_licenses = ", ".join(layer_file_licenses_list)
 
     for package in layer.packages:
+        pkg_licenses = package.pkg_license
```
rnjudge, line 117: To fix the error I mentioned in my [comment](https://github.com/tern-tools/tern/pull/1029/files#issuecomment-958615542), this should be:
```
        pkg_licenses = [package.pkg_license]
```

tern/formats/default/generator.py
```diff
@@ -113,8 +114,14 @@ def get_layer_info_list(layer):
...
+        if package.pkg_format == 'deb':
+            pkg_licenses = wrap(", ".join(package.pkg_licenses)) or ['']
         package_list.add_row([package.name, package.version,
-                              package.pkg_license, package.pkg_format])
+                              pkg_licenses[0], package.pkg_format])
+
```
rnjudge, line 122: Can you add a comment here describing this extra loop? Something like:
```
# Each wrapped line must be added as a new row under the same package
```

tern/formats/default/generator.py
```diff
@@ -113,8 +114,14 @@ def get_layer_info_list(layer):
...
+        for pkg_license in pkg_licenses[1:]:
```
rnjudge, line 123: Can you add a comment just above on line 22 describing this extra loop? Something like:
```
# Each wrapped line must be added as a new row under the same package
```

tern/formats/default/generator.py
```diff
@@ -113,8 +114,14 @@ def get_layer_info_list(layer):
         file_level_licenses = ", ".join(layer_file_licenses_list)
 
     for package in layer.packages:
+        pkg_licenses = package.pkg_license
+        if package.pkg_format == 'deb':
+            pkg_licenses = wrap(", ".join(package.pkg_licenses)) or ['']
```
rnjudge, line 119: Let's add a `width` value to this wrap as well as a comment. I think 45 should be sufficient so the table line is not too ridiculously long:

```
            # If many pkg_licenses, wrap the line for the table at 45 chars
            pkg_licenses = wrap(", ".join(package.pkg_licenses),
                                                width=45) or ['']

```

mukultaneja:
@rnjudge I have taken care of all your suggestions, please review the PR once again.

tern/formats/default/generator.py
```diff
@@ -113,8 +114,16 @@ def get_layer_info_list(layer):
         file_level_licenses = ", ".join(layer_file_licenses_list)
 
     for package in layer.packages:
+        pkg_licenses = [package.pkg_license]
+        if package.pkg_format == 'deb':
+            # If many pkg_licenses, wrap the line for the table at 45 chars
+            pkg_licenses = wrap(", ".join(package.pkg_licenses), width=45) or ['']
```
rnjudge, line 120: small nit: this line is longer than 72 characters. I suggest breaking it up just before `width=45`

tern/formats/default/generator.py
```diff
@@ -101,7 +102,7 @@ def get_layer_info_list(layer):
     layer_file_licenses_list = []
     file_level_licenses = None
     package_list = PrettyTable()
-    package_list.field_names = ["Package", "Version", "License", "Pkg Format"]
+    package_list.field_names = ["Package", "Version", "License(s)", "Pkg Format"]
```
rnjudge, line 105: small nit: This line is also now longer than 72 chars :( suggest breaking it up after `License(s)`

rnjudge:
@mukultaneja A few small nits in the formatting and this is ready to go. Also, can you please update your commit message? It currently says `Adds a new column 'Pkg Licenses' in the table` which is not what the commit does anymore. Something like this would be great:

```
Report Debian package licenses in default report

Include Debian package licenses in the default report table and wrap
the table column width at 45 characters so the table renders as a
reasonable size.

Resolves: #989

Signed-off-by: Mukul Taneja <mtaneja@vmware.com>
```

mukultaneja:
@rnjudge, Please review the changes!

rnjudge:
@mukultaneja LGTM!

mukultaneja:
@rnjudge @nishakm is it good to go then?

PR #1030, opened 2021-08-23
Automates the Tern release process

### LINKED ISSUE #885: Automate Tern's release process

ReconPangolin:
The release process is now partially automated by two workflow actions triggered by a workflow dispatch:
prepare_release.yml - Updates requirements, creates a release summary and runs tests
upload_release.yml - Moves the release requirements and summary to docs/releases and creates a new github release
Currently the release is not uploaded to PyPi so any steps after 'twine check dist/*' are not yet implemented.

Resolves #885


ReconPangolin:
The test 'tern report -w photon.tar' was failing with 'CRITICAL - __main__ - Expected path to tar archive.' so changes may need to be made to how the environment is setup in prepare_release to fix testing.

ReconPangolin:
I fixed the majority of issues identified by the linters, a couple were false positives as the python file is a script meant to run on its own not a module.

nishakm:
> I fixed the majority of issues identified by the linters, a couple were false positives as the python file is a script meant to run on its own not a module.

@ReconPangolin can you try adding the `pydriller` module to `setup.cfg` under the `dev` section? That might help with the linting error about importing the module.

nishakm:
We will probably revisit this issue after the Beta release. Till then we will hold off on any changes.

PR #1045, opened 2021-09-28
Fix: duplicate scancode files being reported

### LINKED ISSUE #1000: existence of cache file(.tern/) effects report output
**Describe the bug**
Output of tern differs whether there exist cache files(.tern/) or not.
image

debian_buster_img#.json : created by command 
`tern report -f json -i debian:buster -o debian_buster_img#.json`
debian_buster_img_scancode#.json : created by command
`tern report -f json -x scancode -i debian:buster -o debian_buster_img_scancode#.json`

**To Reproduce**
1. Pull debian:buster
2. Run command `tern report -f json -i debian:buster -o debian_buster_img1.json`
3. Run command `tern report -f json -i debian:buster -o debian_buster_img2.json`
4. Remove cache `rm -rf ~/.tern/`
5. Run command `tern report -f json -x scancode -i debian:buster -o debian_buster_img_scancode1.json`
6. Run command `tern report -f json -x scancode -i debian:buster -o debian_buster_img_scancode2.json`
7. Run command `tern report -f json -i debian:buster -o debian_buster_img3.json`
8. Run command `tern report -f json -x scancode -i debian:buster -o debian_buster_img_scancode3.json`
9. Remove cache `rm -rf ~/.tern/`
10. Run command `tern report -f json -i debian:buster -o debian_buster_img4.json`
11. Remove cache `rm -rf ~/.tern/`
11. Run command `tern report -f json -x scancode -i debian:buster -o debian_buster_img_scancode4.json`

**Error in terminal**
no error occurred

**Expected behavior**
Same command need to provide same output

Enter all that apply
Tern at commit 273e3c8cd8969df3dacc56c5d878d65378d8e4bf
Python 3.6.13
Docker version 19.03.13, build 4484c46d9d

Distributor ID: Ubuntu
Description:    Ubuntu 16.04.7 LTS
Release:        16.04
Codename:       xenial

rnjudge:
When running scancode with an empty cache, Tern was reporting the same
file twice due to a logic bug in the add_file_data() function. This fix
adds a flag to ensure that a file is only added to the layer object
file list if it does not already exist there.

Resolves #1000

### DIFF
--- a/tern/extensions/scancode/executor.py
+++ b/tern/extensions/scancode/executor.py
@@ -166,13 +166,16 @@ def add_file_data(layer_obj, collected_files):
     # the file level data already in the layer object
     logger.debug("Collecting file data...")
     while collected_files:
+        merged = False
         checkfile = collected_files.pop()
         for f in layer_obj.files:
-            if f.merge(checkfile):
+            merged = f.merge(checkfile)
+            if merged:
                 # file already exists and has now been updated
                 break
-        # file didn't previously exist in layer so add it now
-        layer_obj.files.append(checkfile)
+        if not merged:
+            # file didn't previously exist in layer so add it now
+            layer_obj.files.append(checkfile)
 
 
 def add_package_data(layer_obj, collected_packages):


tern/extensions/scancode/executor.py
```diff
@@ -165,14 +165,17 @@ def add_file_data(layer_obj, collected_files):
...
         for f in layer_obj.files:
             if f.merge(checkfile):
                 # file already exists and has now been updated
+                updated = True
                 break
-        # file didn't previously exist in layer so add it now
-        layer_obj.files.append(checkfile)
+        if not updated:
+            # file didn't previously exist in layer so add it now
+            layer_obj.files.append(checkfile)
```
nishakm, line 178: Can you check my logic here? I think this can be achieved this way:
```
for f in layer_obj.files <-- cycle through existing files
    merged = f.merge(checkfile) <-- try to update the file
    if not merged: <-- the file didn't get updated, means this is a unique file
        layer_obj.files.append(checkfile)
```

nishakm (changes requested):

tern/extensions/scancode/executor.py
rnjudge: I think that should work. I'll test it.

rnjudge: I think we still need the `break` and to update the file only when the for loop has been broken or completed. Otherwise, with your proposal, say we have three files in the layer [A, B, C] and two files from scancode that needs updating: [A, C]. We go through this for loop and compare A to A, merged = True so file gets updated but then we move to the next f in layer files and compare B from the layer files to the same A from scancode... merged = False and so we add the file, even though it already exists and has been updated... and we are facing the same problem we have now. I think we could do:

```
merged = False <-- variable needs to exist outside of for loop so we can check it once the loop completes.
for f in layer_obj.files:
   merged = f.merge(checkfile)
   if merged:
      break
if not merged: <-- if we are here it's because we broke out of the loop (in which case the file has been updated and merged=True or because we never found a match and merged=False and file needs to be updated.
   layer_obj.files.append(checkfile)
```

Thoughts?

nishakm: Yup! This makes sense.

nishakm (approved):

PR #1076, opened 2021-10-30
Dockerfile for recent changes

sameer1046:
This Dockerfile file builds the recent changes in the tern source code.


rnjudge:
Hi @sameer1046 -- can you help me understand the purpose/motivation of this PR? We already have a [Dockerfile](https://github.com/tern-tools/tern/blob/main/ci/Dockerfile) that installs recent changes that we use for our CI tests and not sure the reason behind removing GitHub CI workflow.

sameer1046:
The ci docker file copy the tar file for installation. For that i need to build a tar file.
COPY dist/tern-*.tar.gz
But here it copy the current directory to the container and builds the image. 
COPY . ./tern
So here one step is reduced.

rnjudge:
Thanks @sameer1046 -- Sure, but what is the reason to setup the CI using Azure pipelines? As you can see from the tests not run in this PR -- it bypasses the default CI tests that we have in place.

sameer1046:
Thanks @rnjudge 
The test didn't run because it's waiting for approval.

rnjudge:
That was strange... it didn't show the "approve and run" button for the CI changes. I see it now.

I still am not convinced this change is right as it is. Especially adding a new Dockerfile at the main project level 99% similar to the file located at `ci/Dockerfile` -- it doesn't seem like the right place and it feels like we should be able to make a small modification to the ci Dockerfile and get the same result.

Can you open a feature request issue and try to add a little background as to what you are trying to accomplish and how it might benefit other users of the project? It looks like you are trying to  setup the repo to run azure pipelines, is that right? Please include as much detail as possible :) Thanks @sameer1046!

sameer1046:
yes correct. can we have the ci docker file in root folder?

PR #1077, opened 2021-11-01
DO NOT MERGE: check if labeled tag works

rnjudge:
This commit is a test to see if the "labeled" type works to prevent CI
workflow runs when a PR is not labeled by a maintainer with write
access.


rnjudge:
It works :) Closing.

Issue #1042, opened 2021-09-24
 Multiple checks for keys in check_sourcable function


nishakm:
**WARNING**
If this is your first tern contribution, please update your local git settings using [these instructions](https://github.com/nishakm/puns#how-to-configure-your-git).

DO NOT USE THE GITHUB UI!

When creating your commit message for your PR, make sure to use `git commit -s` rather than `git commit -m`

Remember to add a `Fixes: #1042` line in your commit message.

**Description**
There are multiple checks for certain keys in the check_sourcable function
which can be made by iterating through various dictionaries rather than calling the
keys() function on the dictionaries. Here we update the checks for keys directly
against the dictionaries.

**Implementation**
Make the following changes in `tern/analyze/default/command_lib/command_lib.py`:
```
176 def check_sourcable(command, package_name):                                     
177     '''Given a command and package name find out if the sources can be traced   
178     back. We find this out by checking the package against the command library  
179     If the package has a url or source retrieval steps associated with it       
180     then we return True. If not then we return false'''                         
181     sourcable = False                                                           
182     if command in command_lib['snippets'].keys(): <-- change to command_lib['snippets']                              
183         for package in command_lib['snippets'][command]['packages']:            
184             if package['name'] == package_name or \                             
185                     package['name'] == 'default':                               
186                 if 'url' in package.keys() or \   <-- change to package                              
187                         'src' in package.keys():  <-- change to package                              
188                     sourcable = True                                            
189     return sourcable 
```

JinduOkoli:
HI, can I be assigned to this?

sayantani11:
@rnjudge Can I be assigned this?

rnjudge:
@sayantani11 Sure!

PR #1080, opened 2021-11-08
Multiple checks for keys in check_sourcable function

sayantani11:
There are multiple checks for certain keys in the check_sourcable
function
which can be made by iterating through various dictionaries rather than
calling the
keys() function on the dictionaries. Here we update the checks for keys
directly
against the dictionaries.

Fixes: #1042

Signed-off-by: sayantani11 <ii.sayantani.ii@gmail.com>

### DIFF
--- a/tern/analyze/default/command_lib/command_lib.py
+++ b/tern/analyze/default/command_lib/command_lib.py
@@ -179,12 +179,12 @@ def check_sourcable(command, package_name):
     If the package has a url or source retrieval steps associated with it
     then we return True. If not then we return false'''
     sourcable = False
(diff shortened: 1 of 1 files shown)


rnjudge:
Thanks for the PR @sayantani11. One small nit - can you fix the line spacing on your commit? There's some odd new lines in there and it should all be one paragraph. For an example see how it's formatted in my comment [here](https://github.com/tern-tools/tern/pull/1059#issuecomment-938763350).


Screen Shot 2021-11-08 at 7 14 56 PM
Thanks!

PR #1086, opened 2021-11-22
Enable Tern to run in an unprivileged container

nishakm:
Fixes #1086

This PR brings in changes required to enable Tern to run in an
unprivileged container. This allows us to run Tern on platforms
that do not allow containers with CAP_SYS_ADMIN privileges.

The major change is to include a method of applying the diff
filesystem layers using bulk copying of the files and directories
into the merge directory, rather than using a storage driver like
overlay2 or fuse. However, in order to do this, we also need to
manually handle whiteout files. Commits 1 and 2 handle
detection and use of whiteout files. Commit 1 reverts a change
that excludes whiteout files during the tarball extraction and Commit
2 handles checking and tracking of whiteout files in each image
layer.

Commit 3 removes any unnecessary mounting of inodes, thus
removing the need to run any kind of bind mounts. Commit 4
introduces the manual application of layers to replace overlay
mounting. Commit 5 handles applying layers when there is a
cache hit. Finally Commit 6 modifies the Dockerfiles to no
longer run tern with an external driver. The option is still available
via the `--driver fuse` option, hence not breaking operation for
downstream users. At the same time, the docker_run.sh script
will use the default driver rather than the fuse driver.

### DIFF
--- a/tern/__main__.py
+++ b/tern/__main__.py
@@ -145,11 +145,12 @@ def main():
                         help="Change default working directory to specified"
                         " absolute path.")
     parser.add_argument('-dr', '--driver', metavar="DRIVER_OPTION",
-                        help="Required when running Tern in a container."
-                        "Using 'fuse' will enable the fuse-overlayfs driver "
-                        "to mount the diff layers of the container. If no "
-                        "input is provided, 'fuse' will be used as the "
-                        "default option.")
+                        default='default',
+                        help="Choose from the following storage drivers: \n"
+                        "overlay2: Use the kernel's overlay2 storage driver\n"
+                        "fuse: Use the fuse-overlayfs system tool\n"
+                        "If no option is given, the default method of "
+                        "applying container layers in userspace will be used.")
 
     # sys.version gives more information than we care to print
     py_ver = sys.version.replace('\n', '').split('[', maxsplit=1)[0]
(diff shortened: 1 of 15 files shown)


tern/__main__.py
```diff
@@ -145,11 +145,11 @@ def main():
...
-                        help="Required when running Tern in a container."
-                        "Using 'fuse' will enable the fuse-overlayfs driver "
-                        "to mount the diff layers of the container. If no "
-                        "input is provided, 'fuse' will be used as the "
-                        "default option.")
+                        default='default',
+                        help="Choose from the following storage drivers: \n"
+                        "overlay2: Use the kernel's overlay2 storage driver\n"
+                        "fuse: Use the fuse-overlayfs system tool\n"
+                        "If no option is given, 'default' will be used.")
```
rnjudge, line 152: what is "default" in this case?

nishakm: The bulk copy while taking care of whiteouts method.

rnjudge: Yeah, I meant that we might want to clarify this. A new user likely won't know what the "default" method is.

nishakm: Do you have a recommendation on what to call this method?

rnjudge: I think default is fine but maybe phrasing like "If no option is given, the default bulk copy method will be used" ? just something that gives a hint as to what the default method is. If they want to know more, we have that info in the README.

Issue #1131, opened 2022-03-11
Purl Generation Alpine


ThiefaineM:
**Describe the bug**
The generated purl for alpine composants does not match the one use in OSSINDEX.

**To Reproduce**
1- Create a Docker with an alpine image such as postgres:12.0-alpine
2- Use Tern to generate a BOM in cycloneDx json format
3- Look at the purl generate: for example the one for libxml2

**Expected behavior**
It seems that for alpine the purl looks like **"pkg:alpine/libxml2@2.9.9-r2"** instead of **"pkg:apk/alpine/libxml2@2.9.9-r2"**. Here the two result depending on the purl : 
https://ossindex.sonatype.org/component/pkg:alpine/libxml2 
https://ossindex.sonatype.org/component/pkg:apk/alpine/libxml2
It looks like that for alpine the package type should not be written

ThiefaineM:
I think that it can be fix by removing the apk field in the definition of purl_types_with_namespaces in the file tern/formats/cyclonedx/cyclonedx_common.py

ivanayov:
Can I work on this?

rnjudge:
@ivanayov I think @ThiefaineM already opened a PR with a fix for this.

rnjudge:
@ThiefaineM Mentioned this in my PR, but it does not look like apk/alpine packages are actually defined in the purl spec so curious why ossindex has a preference, do you know?

ThiefaineM:
I do not have an idea on why they did it like this sorry.
Nevertheless by looking at the spec, I saw that Python packages (with others) should have their name is lowercase. But it is not the case as generated purl for Django is : "pkg:pip/Django@3.2.3"
Should I create an other issue specific to it ?

rnjudge:
> 

Yes, please. By the way, I opened an [issue](https://github.com/package-url/purl-spec/issues/159) in the purl repository asking about the alpine/apk purl type entry. It would make me feel better to have a specified Alpine type in the purl spec before merging your PR but I'll see what they respond with. I'm not opposed to merging your PR if the purl maintainers seem agreeable with the current ossindex expectation.

PR #1135, opened 2022-03-21
Correction of the purl generation for apk package

ThiefaineM:
The purl generation for apk packages was faulty.
For example the purl for alpine busybox was pkg:apk/alpine/busybox@1.31.1-r9
instead of pkg:alpine/busybox@1.31.1-r9.
Here are the result in the OSSINDEX base with the two purls:
- https://ossindex.sonatype.org/component/pkg:apk/alpine/busybox@1.31.1-r9
- https://ossindex.sonatype.org/component/pkg:alpine/busybox@1.31.1-r9

Resolves: #1131

### DIFF
--- a/tern/formats/cyclonedx/cyclonedxjson/package_helpers.py
+++ b/tern/formats/cyclonedx/cyclonedxjson/package_helpers.py
@@ -23,7 +23,10 @@ def get_package_dict(os_guess, package):
     purl_type = package.pkg_format
     purl_namespace = cyclonedx_common.get_purl_namespace(os_guess, package.pkg_format)
     if purl_type:
-        purl = PackageURL(purl_type, purl_namespace, package.name, package.version)
+        if purl_type=="apk":
+            purl = PackageURL(purl_namespace, package.name, package.version)
+        else:
+            purl = PackageURL(purl_type,purl_namespace, package.name, package.version)
         package_dict['purl'] = str(purl)
 
     if package.pkg_license:


rnjudge:
Hi @ThiefaineM thanks for the PR.. In you description you say:
```
For example the purl for alpine busybox was pkg:apk/alpine/busybox@1.31.1-r9
instead of pkg:apk/alpine/busybox@1.31.1-r9.
```

but these seem like the same string to me? According to the links you provide it seems like the package url should not have the `apk` string, correct? This proposed change like a fine workaround but I'm also wondering if the package url spec needs to be updated to include "Alpine". According to the listed types, it seems like the purl spec is [still looking](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#other-candidate-types-to-define) for `alpine` package definition so it's odd to me that ossindex can find one alpine package but not the other despite purl not having a defined apk/alpine package type.

ThiefaineM:
Hello @rnjudge !
Yes I definitely made a error, I edited the PR and I will edit the commit message :)
So yes, according to OSSINDEX, it should not have the apk mentionned. In the spec of Purl type, it's indeed still not defined. 
I think that OSSINDEX decided on their own this format for Alpine as none was proposed. 
If not mistaken, OSSINDEX is the biggest base where we can search with purl. So yes I think purl spec should be updated to fit the choice of OSSINDEX. But I am no expert on this.

rnjudge:
Hi @ThiefaineM I am going to cut a release tonight so I am taking over your changes to ensure they get merged in time (The commit message for this PR still needs to be updated as of this writing). Thank you so much for your contributions and making Tern+cyclonedx more purl friendly :)

Issue #1130, opened 2022-03-10
Purl generation


ThiefaineM:
**Please read the Code of Conduct before you proceed**
[Code of Conduct](/CODE_OF_CONDUCT.md)

**Ask your question here**
Hello, I was using Tern on my Dockerfile where I use Python and the framework Django to generate CyclondeDX BOM. The returned purl for django is **"pkg:pip/Django@3.2.3"**. Nevertheless, the "correct" purl to identify django seems to be **"pkg:pypi/django@3.2.3"** according to OSSINDEX and Package URL examples. I got the same issue on most components where the expected pypi is pip. Is this a normal behaviour or am I missing something ?
Best regards.

rnjudge:
cc @coderpatros :)
Thanks for the issue @ThiefaineM!

Looking at what generates the purl:
```
 purl_type = package.pkg_format
    purl_namespace = cyclonedx_common.get_purl_namespace(os_guess, package.pkg_format)
    if purl_type:
        purl = PackageURL(purl_type, purl_namespace, package.name, package.version)
        package_dict['purl'] = str(purl)
```

the purl_type is using the package format, which for python package is `pip` https://github.com/tern-tools/tern/blob/1d9f547cbb5a47fe6f91ab8e0adc9aa966c82242/tern/analyze/default/command_lib/base.yml#L278

I think some simple logic could fix this. I wonder if there's other package format types that might also need tweaking?

ThiefaineM:
Thank you for your quick reaction, I will not hesitate to create an issue if I encounter others in the future.

PR #1136, opened 2022-03-21
Modification of package format for pip

ThiefaineM:
Packages download with pip or pip3 are taken from the index Pypi.
To be more accurate for the purl / in general, it could be better to use Pypi
to describe the packages format.
Here for the purl for a python component:
- https://ossindex.sonatype.org/component/pkg:pip/pip-tools
- https://ossindex.sonatype.org/component/pkg:pypi/pip-tools

The scond one is recognized but not the first.

Resolves: #1130

### DIFF
--- a/tern/analyze/default/command_lib/base.yml
+++ b/tern/analyze/default/command_lib/base.yml
@@ -275,7 +275,7 @@ rpm:
     delimiter: "\n"
 # pip ----------------------------------------------------------------------
 pip:
-  pkg_format: 'pip'
+  pkg_format: 'pypi'
   os_guess:
     - 'None'
   path:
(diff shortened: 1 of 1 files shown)


rnjudge:
@nishakm I'm okay with this change, wdyt? The sample reports will need to be updated as well but I can do that in another PR.

ThiefaineM:
It is updated, sorry for the inconvenience !

nishakm:
LGTM!

rnjudge:
Shoot, @ThiefaineM - I think there is an extra commit attached to this PR now. Can you remove the "[Merge branch 'tern-tools:main' into purl-gen-pip](https://github.com/tern-tools/tern/pull/1136/commits/bb29864d04d5658c311bf372d4c903c827763770)" commit?

ThiefaineM:
Oh no, I was not sure if I should have done it I guess I have the answer now.
It will be done tomorrow, it's not easy to work in a different timezone, I am sorry.

Issue #1140, opened 2022-03-29
Purl Generation name


ThiefaineM:
**Describe the bug**
According to the purl spec, some package names should always be in lowercase : https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst
One of them is for pypi languages, but at the moment the generated purl for Django is : "pkg:pypi/Django@3.2.3"

**To Reproduce**
1. Launch Tern on a docker with django with a cyclonedxjson report

I will work on the PR (if it's okay of course).

ThiefaineM:
Hello @rnjudge !
I have a question: so there is the list of package recognized in the purl spec and the one in Tern : tern/tern/analyse/defaukt/command_lib/base.yml
Should I only focus on the twelve in base.yml and ignored the others ?

rnjudge:
Hi @ThiefaineM, base.yml lists the package managers that Tern is able to collect metadata for (I imagine the purl spec has many more than 12). If you have other package mangers that you would like to see added to base.yml please let me know and we can look in to them. The scripts in base.yml are what Tern runs to collect package metadata for each package manager so any package manager we add will have to have a corresponding command or set of commands that can be run in a container from a command line to collect metdata.

ThiefaineM:
OK, then for the pull request I only worked with the one present in base.yml.
I think base.yml has most of the interesting packages manager but **maven** could be a big plus :)

PR #1141, opened 2022-03-29
Purl name generation

ThiefaineM:
A function has been added at the
creation of the purl to lowercase
the name according to the package
format

Resolve: #1140 

### DIFF
--- a/tern/formats/cyclonedx/cyclonedxjson/package_helpers.py
+++ b/tern/formats/cyclonedx/cyclonedxjson/package_helpers.py
@@ -15,7 +15,7 @@ def get_package_dict(os_guess, package):
     ''' Given a package format, namespace and package object return a
     CycloneDX JSON dictionary representation of the package. '''
     package_dict = {
-        'name': package.name,
+        'name': cyclonedx_common.get_purl_name(package.name, package.pkg_format),
         'version': package.version,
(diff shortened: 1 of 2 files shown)


tern/formats/cyclonedx/cyclonedxjson/package_helpers.py
```diff
@@ -15,7 +15,7 @@ def get_package_dict(os_guess, package):
     ''' Given a package format, namespace and package object return a
     CycloneDX JSON dictionary representation of the package. '''
     package_dict = {
-        'name': package.name,
```
rnjudge, line 18: can we just return `package.name.lower()` here? Are there any purls that require uppercase?

ThiefaineM: Among the packages manager supported by Tern, gem and pacman do not have specifications on their name in the purl-spec.
Also if you want to extend to other packages managers, some of them such as composer or maven do not specification on the name. And some of them should have uppercase such as nugget.

rnjudge:
Hi @ThiefaineM I am going to cut a release tonight so I am taking over your changes to ensure they get merged in time. Thank you so much for your contributions and making Tern+cyclonedx more purl friendly :)

PR #1145, opened 2022-03-31
Readme

vargenau:
README says:

Note that Scancode currently works on Python 3.5 and 3.6 but not 3.7 onwards

This is no longer true.

https://scancode-toolkit.readthedocs.io/en/latest/getting-started/install.html#prerequisites

ScanCode needs a Python 3.6+ interpreter; We support all Python versions from 3.6 to 3.9.


rnjudge:
Hi @vargenau, thanks again for the PR! It looks like there is an extra commit attached to the commit you are intending and also missing a commit message. I can merge once those are fixed!

PR #1159, opened 2022-04-20
Update license data to show actual license type instead of license ref

### LINKED ISSUE #1147: In SPDX output, have PackageLicenseDeclared give the license directly instead of creating an unnecessary LicenseRef.
**Describe the Feature**
In SPDX output, have PackageLicenseDeclared give the license directly instead of creating an unnecessary LicenseRef.

**Use Cases**
It will be easier for tools to process the SPDX files.

**Implementation Changes**
One would need to be able to identify a license already in SPDX format.

Tern is run with the following command:
```
tern report -f spdxtagvalue -i danmcni/danm-cni-plugins:latest -o danm-cni-plugins.spdx
```

[danm-cni-plugins.spdx.txt](https://github.com/tern-tools/tern/files/8429011/danm-cni-plugins.spdx.txt)

In the SPDX output file, you have the following:
```
PackageLicenseDeclared: LicenseRef-c66410f
```
and
```
LicenseID: LicenseRef-c66410f
ExtractedText: <text>Original license: GPL-2.0-only</text>
```
The license here is correctly detected and the name is the one found in the SPDX license list.
So it would be much better to output:
```
PackageLicenseDeclared: GPL-2.0-only
```

ivanayov:
Previously the `PackageLicenseDeclared` and `licenseDeclared` data 
for `spdxtagvalue` and  `spdxjson` respectively were set to license reference
of type `LicenseRef-df8cb33` which is not informative. This change updates
that data to the actual license info, f.e. `MIT`, in case a license is declared,
or the `LicenseRef-df8cb33` value if it's not

Resolves #1147

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ regex>=2022.3
 GitPython~=3.1
 prettytable~=3.2
 packageurl-python>=0.9.9
+license_expression>=21.6.14
--- a/tern/formats/spdx/spdx_common.py
+++ b/tern/formats/spdx/spdx_common.py
@@ -13,6 +13,7 @@ import logging
 import re
 import uuid
 
+from license_expression import get_spdx_licensing
 from tern.utils import constants
 from tern.formats.spdx.spdxtagvalue import formats as spdx_formats
 
@@ -42,6 +43,14 @@ def get_string_id(string):
 def get_license_ref(license_string):
     """ For SPDX tag-value format, return a LicenseRef string """
     return 'LicenseRef-' + get_string_id(license_string)
+
+def is_spdx_license_expression(license_data):
+    '''Return True if the license is a valid SPDX license expression, else
+    return False'''
+    licensing = get_spdx_licensing()
+    if ',' in license_data:
+        license_data = license_data.replace(',', ' ')
+    return licensing.validate(license_data).errors == []
 
 
 ########################
--- a/tern/formats/spdx/spdx_common.py
+++ b/tern/formats/spdx/spdx_common.py
@@ -49,8 +49,16 @@ def is_spdx_license_expression(license_data):
     return False'''
     licensing = get_spdx_licensing()
     if ',' in license_data:
-        license_data = license_data.replace(',', ' ')
+        license_data = license_data.replace(',', ' and ')
     return licensing.validate(license_data).errors == []
+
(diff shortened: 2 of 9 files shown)


rnjudge:
Thanks again for the PR, @ivanayov :) One other change that I think we will need to make is in the part of spdx document creation where the list of license refs gets generated. If we use the SPDX license expression instead of using a licenseref, we don't need to include the license ref extracted text at the end of the document. 

This set of license ref/extracted text gets created in `get_image_packages_license_block()` in `tern/formats/spdx/spdxtagvalue/image_helpers.py` and `get_image_extracted_licenses()` in `tern/formats/spdx/spdxjson/image_helpers.py`. Something like this would due for the change (assuming `is_spdx_license_expression` gets moved to `spdx_common.py` per suggestion above):

```
--- a/tern/formats/spdx/spdxtagvalue/image_helpers.py
+++ b/tern/formats/spdx/spdxtagvalue/image_helpers.py
@@ -57,9 +57,10 @@ def get_image_packages_license_block(image_obj):
             if package.pkg_license:
                 licenses.add(package.pkg_license)
     for lic in licenses:
-        block += spdx_formats.license_id.format(
-            license_ref=spdx_common.get_license_ref(lic)) + '\n'
-        block += spdx_formats.extracted_text.format(orig_license=lic) + '\n'
+        if not spdx_common.is_spdx_license_expression(lic):
+            block += spdx_formats.license_id.format(
+                license_ref=spdx_common.get_license_ref(lic)) + '\n'
+            block += spdx_formats.extracted_text.format(orig_license=lic) + '\n'
     return block


--- a/tern/formats/spdx/spdxjson/image_helpers.py
+++ b/tern/formats/spdx/spdxjson/image_helpers.py
@@ -31,9 +31,10 @@ def get_image_extracted_licenses(image_obj):
                 unique_licenses.add(package.pkg_license)
     extracted_texts = []
     for lic in list(unique_licenses):
-        extracted_texts.append(json_formats.get_extracted_text_dict(
-            extracted_text=lic, license_ref=spdx_common.get_license_ref(
-                lic)))
+        if not spdx_common.is_spdx_license_expression(lic):
+            extracted_texts.append(json_formats.get_extracted_text_dict(
+                extracted_text=lic, license_ref=spdx_common.get_license_ref(
+                    lic)))
     return extracted_texts

```

PR #1162, opened 2022-05-11
Add subdir for tar files in the working dir

HontoNoRoger:
The safeguard of adding a subdir based on the first part of the
file name is necessary for docker-like image tar archives,
which don't have a folder for each layer but rather another tar.gz
file or similar. This is the case for kaniko built docker image
archives.

Without it, tar archives will overwrite the untar_dir on every new
layer analyzed. This will result in the os-release file not being
found as the os analyzer expects the files of the first layer
in the dir but it's actually the files of the last dir.

Signed-off-by: Roger Lehmann <roger.lehmann@newtron.de>

### DIFF
--- a/tern/classes/image_layer.py
+++ b/tern/classes/image_layer.py
@@ -318,8 +318,18 @@ class ImageLayer:
         """Get the directory where contents of the image layer are untarred"""
         # the untar directory is based on the image layout
         if self.image_layout == 'docker':
+            # Images built with Kaniko and its tarPath parameter identify
+            # as Docker image based on the metadata format, but keep their
+            # layer tar files in the root of the main tar file.
+            # So we will make sure if there's no subdir for the layer tar
+            # files, we use the first part of the file name as dir name.
+            # This is to prevent overwriting the extracted layers on every
+            # subsequent layer inspection.
+            subdir_name = os.path.dirname(self.tar_file)
+            if not subdir_name:
+                subdir_name = self.tar_file.split('.', 1)[0]
             return os.path.join(rootfs.get_working_dir(),
-                                os.path.dirname(self.tar_file),
+                                subdir_name,
                                 constants.untar_dir)
         # For OCI layouts, the tar file may be at the root of the directory
         # So we will return a path one level deeper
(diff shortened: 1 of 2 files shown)


rnjudge:
Thank you so much for the PR @HontoNoRoger and I apologize for the delay in reviewing.

Can you tell me the steps for how I could test this with a Kaniko raw image or provide one to me to use for testing?

HontoNoRoger:
Hi @rnjudge thanks for getting back to me.
Sure, sorry I didn't initially post a description for reproducing this.

How make an example Kaniko raw image and test it
----------------------------------------------------------------------

1. Open up a terminal and go to the Tern project dir
1. create a folder, e.g. `kaniko`
2. create a `Dockerfile` in the created `kaniko` directory like this one:
```docker
FROM ubuntu:20.04

RUN apt-get update && apt-get install -y curl && apt-get clean
```
3. run Kaniko directly in Docker via
```bash
docker run \
    -v $PWD/kaniko:/workspace \
    gcr.io/kaniko-project/executor:latest \
    --dockerfile="/workspace/Dockerfile" \
    --tarPath="/workspace/image.tar" \
    --destination="image" \
    --no-push \
```
4. after it ran successfully, you should now have a `kaniko/image.tar` you can use for the check
5. check licenses with Tern now via `tern --clear-cache report --raw-image kaniko/image.tar`

This should give you the licenses when the patch in this PR is applied.
Without the patch Tern returns with an error saying it cannot find a base os, therefore no package manager and therefore no licenses.

rnjudge:
@HontoNoRoger ok, this fix worked for me so I'm happy to merge. Would you be willing to add a comment or two to your changes to clarify the different execution paths? This will make it easier for future contributors :) 

Thanks so much for your PR!

HontoNoRoger:
Hi @rnjudge I've added a code comment explaining the purpose of the changes, extended the test case for `testGetUntarDir` and referenced the issue you mentioned. Please let me know if I can futher improve the quality of this PR. Thank you for taking the time to review it. :slightly_smiling_face:

Issue #1170, opened 2022-06-06
FilesAnalyzed: False has a wrong case for "False"


vargenau:
**Describe the bug**
Tern sometimes generates `FilesAnalyzed: False` instead of `FilesAnalyzed: false`

**To Reproduce**

```
tern report -f spdxtagvalue -i mariadb:latest -o mariadb.spdx
```

**Expected behavior**
In SPDX 2.2, section 1.7.7, it is stated: "Tags and format properties are case sensitive".

So boolean `false` should be in lower case like in the examples.

**Environment you are running Tern on**

```
Tern version 2.10.0
   python version = 3.10.4 (main, Apr  2 2022, 09:04:19) 
```

**Please attach files if they exist**
- tern.log
- report.*

rnjudge:
Hi @vargenau -- Good catch! You may also want to file an issue with the [SPDX tools](https://github.com/spdx/tools) repo as the document seems to validate for me even with the `False` value.

```
(ternenv) rjudge:tern$ tern --version
Tern at commit 43fd06fd0bae58b02d632be7f833613a7de06b16
   python version = 3.8.10 (default, Nov 26 2021, 20:14:08)

(ternenv) rjudge:tern$ tern report -f spdxtagvalue -i mariadb:latest -o mariadb.spdx

(ternenv) rjudge:tern$ cat mariadb.spdx | grep False
FilesAnalyzed: False

(ternenv) rjudge:tern$ java -jar /home/rjudge/spdx-tools-2.2.7-jar-with-dependencies.jar Verify mariadb.spdx
This SPDX Document is valid.
```

vargenau:
Hi @rnjudge 

Here it is: <https://github.com/spdx/spdx-online-tools/issues/374>

vargenau:
In fact, it is <https://github.com/spdx/spdx-online-tools/issues/372>

Since it was moved to <https://github.com/spdx/spdx-java-tagvalue-store/issues/22>, I did not find it the list and thought I had not created it.

PR #1171, opened 2022-06-06
Use correct case for 'false'. Resolves: #1170

vargenau:
Use correct case for 'false'.

Resolves: #1170

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>


rnjudge:
This PR looks good. The failing prospector linting test is unrelated to the change so I'll look into that then get this merged. Thanks for the PR @vargenau!

Issue #1211, opened 2023-02-06
Export specific format versions (SPDX)


vargenau:
We now have SPDX 2.2 (the ISO version) and SPDX 2.3, and soon we will have SPDX 3.0.

It would be good to be able to specify in which version of SPDX we want the result.

Some users are required to use the ISO version, and some have tools that require a specific version of SPDX.

Syft has planned to implement it:
[Export specific format versions (SPDX) ](https://github.com/anchore/syft/issues/1519)

We could use the same syntax to specify the version.

rnjudge:
Thanks @vargenau. This is on our radar but due to limited resources, not high priority.. Ideally, we would only support the latest SPDX release since.

PR #1233, opened 2023-06-23
Add SPDX generation using spdx-tools

armintaenzertng:
This is set up to produce the same SPDX output as the current spdx generation module while utilising the spdx-tools library. The goal is to replace the current module with this new one, which will allow easy migration to more SPDX formats as well as SPDXv3.
I tried to stay close to the structure of the original implementation.

I tested this using the following commands:
`tern report -i golang:1.12-alpine -f spdxjson -o spdx_test.json`
and
`tern report -i golang:1.12-alpine -f spdxjson_new -o new_spdx_test.json`
I compared the resulting json files [using jd](http://play.jd-tool.io/), treating arrays as unordered sets. The only differences were in timestamps, UUIDs, and two differences in how json output is generated:
* The enum value in spdx-tools is `PACKAGE_MANAGER`, not `PACKAGE-MANAGER`
* The optional key `documentDescribes` has been deprecated and is therefore not used by the spdx-tools. Only the corresponding `DESCRIBES` relationship is serialized.


armintaenzertng:
I added support for YAML, XML and RDF-XML formats.

tern/formats/spdx_new/constants.py
```diff
@@ -0,0 +1,15 @@
...
+DOCUMENT_NAME = 'Tern report for {image_name}'
+SPDX_VERSION = 'SPDX-2.2'
+DATA_LICENSE = 'CC0-1.0'
+DOCUMENT_COMMENT = 'This document was generated by ' \
+    'the Tern Project: https://github.com/tern-tools/tern'
+DOCUMENT_NAMESPACE = 'https://spdx.org/spdxdocs/tern-' \
+    'report-{version}-{image}-{uuid}'
+LICENSE_LIST_VERSION = Version(3, 20)
+CREATOR_NAME = 'tern-{version}'
+DOCUMENT_NAME_SNAPSHOT = 'Tern SPDX JSON SBoM'  # TODO: different name here that is not specific to JSON
```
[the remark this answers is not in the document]
armintaenzertng: done

tern/formats/spdx_new/file_helpers.py
```diff
@@ -0,0 +1,125 @@
...
+
+def get_file_dict(filedata: FileData, template: Template, layer_id: str) -> SpdxFile:
+    """Given a FileData object and its SPDX template mapping, return an
+    SPDX representation of the file. A layer_id is used to
+    distinguish copies of the same file occurring in different places in the
+    image"""
+    mapping = filedata.to_dict(template)
+
+    if filedata.licenses:
+        license_info_in_file = []
```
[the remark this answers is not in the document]
armintaenzertng: Thanks! Done

rnjudge:
@armintaenzertng how does one denote which version of SPDX documents they want using this PR? I am assuming this PR, by default, generates SPDX 2.3 documents. However, we can't drop support for SPDX 2.2 since we have users who want it because it is the ISO standard version. There needs to be a way to denote SPDX version to generate on the command line before we can merge this.

tern/formats/spdx_new/constants.py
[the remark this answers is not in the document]
rnjudge: small nit: SBoM should be SBOM

armintaenzertng:
This PR currently replicates the behavior of the current state. That is, SPDX 2.2 is hardcoded into the output.
I will gladly add support for multiple SPDX versions, but we should clarify how this would work in the current workflow.
- Should there be an optional parameter that denotes the SPDX version (that gets ignored if no SPDX output is required)? If yes, how do we call that?
- Or will we add two (and more in the future) parameters for the `-f` flag, like `spdxjson2.2`, `spdxjson2.3`, `spdxjson3.0` etc.? This is probably easier to implement, but will lead to much code duplication, probably not the best idea.

armintaenzertng:
I added the versioning I described above.
@rnjudge, please have a look if that is OK with you. :)

rnjudge:
@armintaenzertng do you want to schedule a zoom call about this? I would like to avoid mass code duplication as that was the whole point of using the SPDX tools library.

armintaenzertng:
I added a CLI version parameter for the output format. Formats that don't support this (i.e. everything except for SPDX) will raise an error if this is set.

tern/formats/cyclonedx/cyclonedxjson/generator.py
```diff
@@ -50,12 +50,14 @@ def get_document_dict(image_obj_list):
...
 class CycloneDXJSON(generator.Generate):
-    def generate(self, image_obj_list, print_inclusive=False):
+    def generate(self, image_obj_list, version, print_inclusive=False):
         ''' Generate a CycloneDX document
         The whole document should be stored in a dictionary which can be
         converted to JSON and dumped to a file using the write_report function
         in report.py. '''
         logger.debug('Generating CycloneDX JSON document...')
+        if version is not None:
+            raise ValueError("The version parameter is not supported for CycloneDX JSON.")
```
rnjudge, line 60: Not sure we need to raise an error here. Maybe just print a warning that the version parameter is not supported and will be ignored?

tern/formats/spdx_new/package_helpers.py
```diff
@@ -0,0 +1,168 @@
...
+
+purl_types_with_namespaces = [
+    'deb',
+    'rpm',
+    'apk',
+    'alpm'
+]
+
+
+def get_purl(package_obj: Package) -> str:
```
rnjudge, line 152: This function needs updating per: https://github.com/tern-tools/tern/pull/1237/files#diff-8853c35f8baed4fa963a218102cecf6fc9153733acaaf79b6210af00bc44e068

tern/formats/spdx_new/make_spdx_model.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2021 VMware, Inc. All Rights Reserved.
```
rnjudge, line 3: small nit: copyright year needs updating

armintaenzertng: updated to 2023

tern/formats/spdx_new/package_helpers.py
armintaenzertng: updated

armintaenzertng:
I added SPDX-2.3 functionality.

In particular, this means that if the SPDX version is 2.3, we set the `primary_package_purpose` of the container package to CONTAINER and omit concluded license, declared license and copyright text in SpdxPackages if possible.

tern/__main__.py
```diff
@@ -216,6 +216,9 @@ def main():
                                "available formats: "
                                "spdxtagvalue, spdxjson, cyclonedxjson, json, "
                                "yaml, html")
+    parser_report.add_argument('-fv', '--format-version',
```
rnjudge, line 219: since this is only for SPDX, suggest making that more clear in the wording. Maybe -sv for spdx-version? or just v? Also clarify in the `help` that it is only for the SPDX version and will otherwise be ignored.

rnjudge: Also probably want to indicate what the valid input is. When I test with -fv 2.3 it doesn't work.

armintaenzertng: Done, thanks for the input! :)

vargenau:
This would fix https://github.com/tern-tools/tern/issues/1211

vargenau:
@rnjudge @armintaenzertng 
For specifying the SPDX version of the output, I would propose the following:
```
tern report -f spdxtagvalue@2.3 -i golang:1.12-alpine -o alpine.spdx
```
and
```
tern report -f spdxjson@2.3 -i golang:1.12-alpine -o alpine.spdx.json
```

Advantages would be that we have one less argument and it is similar to the Syft syntax:
https://github.com/anchore/syft/
```
spdx-tag-value: A tag-value formatted report conforming to the SPDX 2.3 specification
spdx-tag-value@2.2: A tag-value formatted report conforming to the SPDX 2.2 specification
spdx-json: A JSON report conforming to the SPDX 2.3 JSON Schema
spdx-json@2.2: A JSON report conforming to the SPDX 2.2 JSON Schema
```

That is what I had prototyped in https://github.com/tern-tools/tern/pull/1228

What do you think?

vargenau:
Hi @armintaenzertng 
My proposal aim was:
- to have one less option
- to have a syntax similar to Syft
If it is more complex to implement, no problem.

vargenau:
> @rnjudge: `tern report -i photon:3.0 -f spdxjson -sv 2.3 -o output.json` followed by `pyspdxtools -i output.json` yields no errors or invalidations, so I'll mark this "Ready for review" now. There are still some open issues regarding Spdx documents with file information, which I collected in #1240. As these existed in the old SPDX implementation already, I'd propose to fix them in a separate PR after this one to keep things clearer (this one is already quite large).

Is is really `-sv 2.3` ?
In Rose message above, it was `parser_report.add_argument('-fv', '--format-version',`

armintaenzertng:
I meant `-sv`, this was changed according to [this comment](https://github.com/tern-tools/tern/pull/1233#discussion_r1266006186).

vargenau:
> I meant `-sv`, this was changed according to [this comment](https://github.com/tern-tools/tern/pull/1233#discussion_r1266006186).

Thank you, I had missed that comment.

ci/test_files_touched.py
```diff
@@ -98,17 +98,29 @@
...
     re.compile('tern/formats/spdx'): [
-        'tern report -f spdxtagvalue -i photon:3.0 -o spdx.spdx && ' \
-        'java -jar tools-java/target/tools-java-*-jar-with-dependencies.jar '\
+        'tern report -f spdxtagvalue -i photon:3.0 -o spdx.spdx && '
+        'java -jar tools-java/target/tools-java-*-jar-with-dependencies.jar '
         'Verify spdx.spdx',
-        'tern report -f spdxjson -i photon:3.0 -o spdx.json && ' \
-        'java -jar tools-java/target/tools-java-*-jar-with-dependencies.jar '\
-        'Verify spdx.json'],
+        'tern report -f spdxjson -i photon:3.0 -o spdx.json && '
```
rnjudge, line 112: @armintaenzertng Do these lines need the continuation like the lines that were removed?

setup.cfg
```diff
@@ -44,14 +44,19 @@ include_package_data = True
...
     html = tern.formats.html.generator:HTML
     cyclonedxjson = tern.formats.cyclonedx.cyclonedxjson.generator:CycloneDXJSON
+    spdxjson = tern.formats.spdx.spdxjson.generator:SpdxJSON
+    spdxyaml = tern.formats.spdx.spdxyaml.generator:SpdxYAML
+    spdxxml = tern.formats.spdx.spdxxml.generator:SpdxXML
+    spdxrdf = tern.formats.spdx.spdxrdf.generator:SpdxRDF
+    spdxtagvalue = tern.formats.spdx.spdxtagvalue.generator:SpdxTagValue
+    spdxtagvalue_legacy = tern.formats.spdx_legacy.spdxtagvalue.generator:SpdxTagValue
+    spdxjson_legacy = tern.formats.spdx_legacy.spdxjson.generator:SpdxJSON
+    spdxjsonc = tern.formats.spdx_legacy.spdxjson.consumer:SpdxJSON
```
rnjudge, line 59: @armintaenzertng What's the idfference between `spdxjsonc` and `spdxjson`?

armintaenzertng: `spdxjsonc` seems to provide the ability to parse SPDX files generated with tern and read from them the layer information. I have not touched that code. Seeing as you don't know about it either, is that feature still supported?
I'm trying to find out how to actually use it via the CLI.

armintaenzertng: It seems like this is actually not meant to be called directly from the CLI, but instead is part of the `--live` option, where you can specify reports generated in previous runs via the `--with-context` flag.

I believe that both `jsonc` and `spdxjsonc` can therefore be removed from the entrypoints list, but seeing that I did not touch any of that code, I propose to do this in a separate PR.

rnjudge: Oh ok, this was functionality Nisha worked on. I didn't realize it had its own entry point but let's leave it for now.

tern/formats/html/generator.py
```diff
@@ -321,16 +321,22 @@ def get_report_dict(image_obj_list):
...
         for the images'''
+        if spdx_version is not None:
+            logger.warning("The SPDX version parameter is not supported for HTML.")
+
         report_dict = get_report_dict(image_obj_list)
         report = create_html_report(report_dict, image_obj_list)
         return report
 
-    def generate_layer(self, layer):
+    def generate_layer(self, layer, version: str):
```
rnjudge, line 334: Same comment as above. Update to `spdx_version`.

setup.cfg
```diff
@@ -44,14 +44,19 @@ include_package_data = True
...
     yaml = tern.formats.yaml.generator:YAML
     html = tern.formats.html.generator:HTML
     cyclonedxjson = tern.formats.cyclonedx.cyclonedxjson.generator:CycloneDXJSON
+    spdxjson = tern.formats.spdx.spdxjson.generator:SpdxJSON
+    spdxyaml = tern.formats.spdx.spdxyaml.generator:SpdxYAML
+    spdxxml = tern.formats.spdx.spdxxml.generator:SpdxXML
+    spdxrdf = tern.formats.spdx.spdxrdf.generator:SpdxRDF
+    spdxtagvalue = tern.formats.spdx.spdxtagvalue.generator:SpdxTagValue
+    spdxtagvalue_legacy = tern.formats.spdx_legacy.spdxtagvalue.generator:SpdxTagValue
+    spdxjson_legacy = tern.formats.spdx_legacy.spdxjson.generator:SpdxJSON
```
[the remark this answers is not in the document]
armintaenzertng: Why do you say that?
I just tried running `tern report -f spdxjson_legacy -i photon:3.0` and it worked without errors for me.

We can of course rename them to make it clearer that this code will be deprecated.

tern/formats/default/generator.py
```diff
@@ -154,8 +157,11 @@ def generate(self, image_obj_list, print_inclusive=False):
             return report
         return report + print_licenses_only(image_obj_list)
 
-    def generate_layer(self, layer):
+    def generate_layer(self, layer, version: str):
```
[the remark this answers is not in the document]
armintaenzertng: Good catch, that was an oversight!

tern/formats/html/generator.py
armintaenzertng: fixed

armintaenzertng:
Small update: I changed the spdx-tools dependency to the new 0.8.1 version.

armintaenzertng:
small fix: I changed the check for the SPDX version to use the official string "SPDX-2.2" consistently.

PR #1260, opened 2026-07-27
fix: align ImageLayer docstring fields

### LINKED ISSUE #1069: There is a mixup of lines in the documentation of `image_layer.py`
I think the lines in the following documentation are somehow not in correct order. Maybe chunks were added between multi line statements. Especially `import_str:` is broken.

https://github.com/tern-tools/tern/blob/a3e76cdb0e08d92e896822634cbcf670bdd05e30/tern/classes/image_layer.py#L18-L43

Eljees:
## Summary

- move the `import_str` continuation back next to its field
- keep the `layer_index` documentation as a separate entry

## Why

The continuation of the `import_str` description was placed after the `layer_index` entry, which made both fields misleading.

Closes #1069.

## Validation

- `git diff --check`
- `python -m pytest tests/test_class_image_layer.py` — 12 passed; one Windows-only path separator assertion failed (`\` versus `/`) in `testGetUntarDir`, unrelated to this docstring-only change


Eljees:
Ping - this has been sitting since 27 July. It is docs-only: the ImageLayer docstring lists fields that no longer match the class. Fine to close it if it is not worth the churn.

